PR #4, opened 2024-10-03
feat: add memory, update examples, update base embedder

olbychos:
- include memory and backends to support memory functionality.
- provide examples of memory usage.

### DIFF
--- a/dynamiq/memory/backend/in_memory.py
+++ b/dynamiq/memory/backend/in_memory.py
@@ -11,7 +11,7 @@ class InMemoryError(Exception):
     pass
 
 
-class BM25:
+class BM25DocumentRanker:
     """BM25 implementation for scoring documents."""
 
     def __init__(self, documents: list[str], k1: float = 1.5, b: float = 0.75):
@@ -93,7 +93,7 @@ class InMemory(MemoryBackend):
             return filtered_messages[:limit]
         query_terms = query.lower().split()
         document_texts = [msg.content for msg in filtered_messages]
-        bm25 = BM25(document_texts)
+        bm25 = BM25DocumentRanker(document_texts)
         scored_messages = [(msg, bm25.score(query_terms, msg.content)) for msg in filtered_messages]
         scored_messages = [(msg, score) for msg, score in scored_messages if score > 0]
         scored_messages.sort(key=lambda x: x[1], reverse=True)


olbychos (commented):
@acoola, please check the 'test' section. It appears that the usage tokens and related components have failed.

dynamiq/memory/backend/qdrant.py
```diff
@@ -0,0 +1,81 @@
...
+
+        try:
+            self.client = self.connection.connect()
+        except Exception as e:
+            raise ConnectionError(f"Failed to connect to Qdrant: {e}")
+
+        # Check if the collection exists, create it if not
+        try:
+            self.client.get_collection(collection_name=self.collection_name)
+        except Exception as e:
```
vitalii-dynamiq, line 26: is it possible to use isinstance() to check the type of execption?

olbychos: Yes, I have added a few checks and they are sufficient.

dynamiq/memory/backend/pinecone.py
```diff
@@ -0,0 +1,126 @@
+import uuid
+
+from pinecone import Pinecone as PineconeClient
+from pinecone import ServerlessSpec
+
+from dynamiq.components.embedders.base import BaseEmbedder
+from dynamiq.connections import Pinecone as PineconeConnection
+from dynamiq.memory.backend.base import Backend
```
vitalii-dynamiq, line 8: should we name it MemoryBackend? so it's not confused with anything else in the future?

dynamiq/memory/backend/base.py
```diff
@@ -0,0 +1,32 @@
...
+        """Adds a message to the memory storage."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_all(self) -> list[Message]:
+        """Retrieves all messages from the memory storage."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def search(self, query: str, search_limit: int) -> list[Message]:
```
vitalii-dynamiq, line 20: I think we're missing a critical parameter here - filters. I should be able to filter by certain conditions before search. and I guess search query and filters can be allowed empty

dynamiq/memory/backend/pinecone.py
olbychos: This is similar to our caching module. Initially, we used the term 'storage,' but then we switched to 'backend.'

dynamiq/memory/backend/base.py
olbychos: I will check the filters. 

The search query is empty — that makes sense.

dynamiq/memory/backend/pinecone.py
vitalii-dynamiq: @acoola I suggest MemoryBackend

olbychos: @MaksymAI @delamainer @acoola @maksDev123 Let's vote on the final name, please.

dynamiq/memory/backend/qdrant.py
```diff
@@ -0,0 +1,146 @@
...
+
+    pass
+
+
+class Qdrant(MemoryBackend):
+    """Qdrant implementation of the memory storage backend."""
+
+    name = "Qdrant"
+
+    def __init__(self, connection: QdrantConnection, embedder: BaseEmbedder, collection_name: str = "conversations"):
```
delamainer, line 24: Should we always use `index_name` instead of `collection_name`. For writers and readers, we always call it `index_name`. I would prefer to keep it consistent. 

Generally, I would prefer to keep the same names as those used by the original database terminology (e.g. `index` for weaviate, `collection` for chroma...)

dynamiq/nodes/agents/base.py
```diff
@@ -71,10 +78,12 @@ def __init__(self, **kwargs):
         self._intermediate_steps: dict[int, dict] = {}
         self._run_depends: list[dict] = []
         self._init_prompt_blocks()
+        if self.memory is None:
+            self.memory = Memory()
```
delamainer, line 82: I think memory should be optional. Now we use it always. What if the user don't want to use memory for some reason?

delamainer (dismissed):

dynamiq/components/embedders/base.py
```diff
@@ -176,3 +176,16 @@ def embed_documents(self, documents: list[Document]) -> dict:
             doc.embedding = emb
 
         return {"documents": documents, "meta": meta}
+
+    @property
+    def embedding_size(self) -> int:
```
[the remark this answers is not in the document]
olbychos: It's a complicated question that depends on the model. We can either rely on the user or double-check it using this method.

dynamiq/nodes/agents/base.py
olbychos: Reasonable, remove it.

dynamiq/memory/backend/pinecone.py
```diff
@@ -0,0 +1,151 @@
...
+class Pinecone(MemoryBackend):
+    """Pinecone implementation of the memory storage backend."""
+
+    name = "Pinecone"
+
+    def __init__(
+        self,
+        connection: PineconeConnection,
+        embedder: BaseEmbedder,
+        index_name: str = "conversations",
```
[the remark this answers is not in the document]
olbychos: It is connected in a way that allows us to obtain this number from the user or to calculate it ourselves for verification.

dynamiq/nodes/agents/base.py
```diff
@@ -130,12 +142,19 @@ def execute(
         self.reset_run_state()
         config = ensure_config(config)
         self.run_on_node_execute_run(config.callbacks, **kwargs)
+        user_id = input_data.get("user_id", None)
+        self.memory.add_message(role=MessageRole.USER, content=input_data.get("input"), metadata={"user_id": user_id})
```
delamainer: There should be a different way to pass more metadata. Now it looks like we are only limited to `user_id`. What about `session_id` and other metadata that users might come up with?

dynamiq/memory/backend/pinecone.py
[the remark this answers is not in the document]
delamainer: I also noticed that we have a lot of terms for the same things.
- dimensions
- dimension
- embedding_size

dynamiq/memory/backend/qdrant.py
delamainer: Let's keep it consistent with the our package. 

Even though I don't like this :/

dynamiq/memory/backend/pinecone.py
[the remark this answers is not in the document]
delamainer: We already have `dimensions` field in `BaseEmbedder`. Can we use it somehow? 

Is there a better way achieve the same without calling the API each time?

dynamiq/components/embedders/base.py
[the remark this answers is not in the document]
vitalii-dynamiq: @olbychos it's probably better for a user to provide a model along with the embeddibgs size. this operation is going to be ver expensive.

dynamiq/memory/backend/in_memory.py
```diff
@@ -0,0 +1,56 @@
...
+from dynamiq.prompts import Message
+
+
+class InMemoryError(Exception):
+    """Base exception class for InMemory backend errors."""
+
+    pass
+
+
+class InMemory(MemoryBackend):
```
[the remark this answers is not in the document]
vitalii-dynamiq: We agreed to look at this later. for now one memory object is going to be used by one node. @olbychos just add TODO to fix race conditions later

dynamiq/memory/backend/pinecone.py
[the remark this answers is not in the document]
vitalii-dynamiq: I agree, we should be using our BaseEmbedder and child classes derived from it

dynamiq/memory/backend/qdrant.py
```diff
@@ -0,0 +1,146 @@
...
+from dynamiq.prompts import Message
+
+
+class QdrantError(Exception):
+    """Base exception for Qdrant-related errors."""
+
+    pass
+
+
+class Qdrant(MemoryBackend):
```
vitalii-dynamiq, line 19: @olbychos actually we already have Qdrant and other implementations developed by @MaksymAI. can you check if there is an overlap? I see that there's a major overlap with `dynamiq/storages/vector`

dynamiq/nodes/agents/base.py
olbychos: I would like to manage 'user_id' and 'session_id' independently while also retaining other metadata. I will provide the changes for your review.

dynamiq/memory/backend/qdrant.py
olbychos: Yes, there are some overlaps as well as key differences in the end use cases.
I think we can create a shared Qdrant client later and use "Mixins" to differentiate the backends for memory and RAG.
We can apply the same approach to other vector storage systems, but that can be handled in a separate PR.

vitalii-dynamiq: Okay, but we better do this straight away, so we don't forget it

olbychos: I’ve added a TODO about this:
- We may need to rethink the proper interface for combining the vector store and memory backend( to keep non vector also)
- There will be major code changes, which will mainly affect RAG.  
- We will discuss this with @MaksymAI  to consider how it can be implemented, what is crucial, and to work on the contract.

acoola (commented):
Great job. Left few comments regarding minor improvements.

For further memory YAML support potentially we would need to convert Memory and all Backends/Configs to pydantic models (but this need to double check)

olbychos:
> Great job. Left few comments regarding minor improvements.
> 
> For further memory YAML support potentially we would need to convert Memory and all Backends/Configs to pydantic models (but this need to double check)

Thank you for your feedback. I have made the necessary corrections; please review it again.

dynamiq/memory/backend/sqlite.py
```diff
@@ -0,0 +1,176 @@
...
+        except Exception as e:
+            raise SQLiteError(f"Error initializing SQLite backend: {e}") from e
+
+    def _validate_table_name(self, create_if_not_exists: bool = False):
+        """Validates the table name to prevent SQL injection and optionally creates it."""
+        if not re.match(r"^[A-Za-z0-9_]+$", self.index_name):
+            raise SQLiteError(f"Invalid table name: '{self.index_name}'")
+
+        try:
+            with sqlite3.connect(self.db_path) as conn:
```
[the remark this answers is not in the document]
vitalii-dynamiq: makes sense

dynamiq/memory/memory.py
```diff
@@ -0,0 +1,89 @@
...
+        self.backend = backend
+
+    def add_message(self, role: MessageRole, content: str, metadata: dict = None):
+        """Adds a message to the memory."""
+        try:
+            metadata = metadata or {}
+            metadata["timestamp"] = datetime.utcnow().timestamp()
+            message = Message(role=role, content=content, metadata=metadata)
+            self.backend.add(message)
+            logger.debug(f"Memory {self.backend.name}: Added message: {message.role}: {message.content[:20]}...")
```
vitalii-dynamiq, line 29: this might break if there is less than 20 elements

dynamiq/memory/memory.py
```diff
@@ -0,0 +1,89 @@
...
+        messages = self.backend.get_all()
+        logger.debug(f"Memory {self.backend.name}: Retrieved {len(messages)} messages")
+        return messages
+
+    def get_all_messages_as_string(self, format: str = "plain") -> str:
+        """Retrieves all messages as a formatted string."""
+        messages = self.get_all_messages()
+        return self._format_messages_as_string(messages, format)
+
+    def search_messages(self, query: str = None, filters: dict = None) -> list[Message]:
```
vitalii-dynamiq, line 45: should we just use .search()

dynamiq/memory/memory.py
```diff
@@ -0,0 +1,89 @@
...
+        """Initializes the Memory with the given configuration and backend.
+
+        If no backend is provided, an InMemory backend is used by default.
+        """
+        if not isinstance(backend, MemoryBackend):
+            raise TypeError("backend must be an instance of Backend")
+        self.config = config
+        self.backend = backend
+
+    def add_message(self, role: MessageRole, content: str, metadata: dict = None):
```
vitalii-dynamiq, line 22: also here it can just be .add()

dynamiq/memory/backend/in_memory.py
```diff
@@ -0,0 +1,56 @@
...
+        """Retrieves all messages from the in-memory list."""
+        return sorted(self.messages, key=lambda msg: msg.metadata.get("timestamp", 0))  # Sort by timestamp
+
+    def search(self, query: str = None, limit: int = None, filters: dict = None) -> list[Message]:
+        """Searches for messages, applying optional query and/or filters."""
+        limit = limit or self.config.search_limit
+        matching_messages = self.messages
+
+        if query:
+            matching_messages = [msg for msg in matching_messages if query.lower() in msg.content.lower()]
```
MaksymAI, line 37: Could you please explain this use case a little bit? We are searching for the exact inclusion of a query in the previous messages.

My question is: Do we want to extract the exact duplicate of the previous user query input? If yes, how will this information help us?

Or do we want to extract answers where this query was mentioned? Are we 100% sure that the answers include the needed input query in the exact match formulation?

I think it would be beneficial to have the ability for semantic retrieval with high similarity confidence, and later with the option to assess with an LLM or cross-encoder if the memory contains the same information or similar but different information that we want to reuse.

olbychos: @MaksymAI there is inmemory implementation, without embeddings and not vector store, so just simple search

olbychos: @MaksymAI we have semantic search only upon vector stores backend

dynamiq/memory/backend/sqlite.py
```diff
@@ -0,0 +1,196 @@
...
+            raise SQLiteError(f"Error validating or creating table: {e}") from e
+
+    def _create_table(self):
+        """Creates the messages table."""
+        query = self.CREATE_TABLE_QUERY.format(index_name=self.index_name)
+        try:
+            with sqlite3.connect(self.db_path) as conn:
+                cursor = conn.cursor()
+                cursor.execute(query)  # nosec B608
+                conn.commit()
```
MaksymAI, line 85: I recommend creating a separate method for the execute function and reusing it everywhere. I have found my old SQLite helper; maybe we can reuse it
```
import sqlite3
from sqlite3 import Cursor
from typing import Dict, List, Tuple


class SqliteClient:
    _connections = {}

    def __init__(self, db_path: str):
        """
        Initializes a new instance of the SqliteClient class and creates
        a connection to the SQLite database, if it not exists yet

        Parameters:
        db_path (str): The path of the SQLite database.
        """
        self.db_path = db_path
        self._init_connection(db_path)

    def __del__(self):
        """Closes the SQLite connection when the object is destroyed"""
        if self._connections.get(self.db_path) is not None:
            self._connections[self.db_path].close()
            self._connections.pop(self.db_path)

    def _init_connection(self, db_path: str):
        if self._connections.get(db_path) is None:
            self._connections[db_path] = sqlite3.connect(
                db_path, check_same_thread=False, timeout=10
            )
        self.cursor = self._connections[db_path].cursor()

    def execute(self, query: str, params: Tuple = (), commit: bool = False) -> Cursor:
        self._init_connection(self.db_path)
        executed_cursor = self.cursor.execute(query, params)

        if commit:
            self._connections[self.db_path].commit()

        return executed_cursor

    def query_all(self, query: str, params: Tuple = ()) -> List[Dict]:
        """
        Executes a query on the SQLite database and returns all the results.

        Parameters:
        query: The SQL query to be executed.
        params: query parameters

        Returns:
        list: A list of dicts containing the results of the query
        """
        executed_cursor = self.execute(query, params)
        columns = [column[0] for column in executed_cursor.description]
        query_res = [dict(zip(columns, row)) for row in executed_cursor.fetchall() or []]
        return query_res

    def query_one(self, query: str, params: Tuple = ()) -> Dict:
        """
        Executes a query on the SQLite database and returns the first result.

        Parameters:
        query: The SQL query to be executed.
        params: query parameters

        Returns:
        tuple: A dict containing the first result of the query
        """
        executed_cursor = self.execute(query, params)
        columns = [column[0] for column in executed_cursor.description]
        query_res = dict(zip(columns, executed_cursor.fetchone() or []))
        return query_res

    def table_exists(self, table_name):
        query = "SELECT name FROM sqlite_master WHERE type='table' AND name=?"
        return self.query_one(query, (table_name,))
```
And example of use
```
self.sqlite = SqliteClient(db_path=db_path)
self.sqlite.query_all(SQL_QUERY, (self.index_name,))
```

scripts/generate_mkdocs.py
```diff
@@ -8,9 +8,9 @@ def create_directory(path):
 
 
 def get_python_files(directory):
-    """Get all Python files in the specified directory, excluding __init__.py files."""
+    """Get all Python files in the specified directory, excluding __init__.py.py files."""
```
MaksymAI, line 11:
```suggestion
    """Get all Python files in the specified directory, excluding __init__.py files."""
```

tests/integration/nodes/test_memory.py
```diff
@@ -0,0 +1,81 @@
...
+AGENT_ROLE = "helpful assistant"
+AGENT_GOAL = "is to provide useful information and answer questions"
+
+
+@pytest.fixture
+def openai_connection():
+    return connections.OpenAI(
+        id=str(uuid.uuid4()),
+        api_key=os.getenv("OPENAI_API_KEY"),
+    )
```
MaksymAI, line 24: I think if we use tests with real credentials, we should move this test into a separate folder [integration_with_creds](https://github.com/dynamiq-ai/dynamiq/tree/main/tests/integration_with_creds)

dynamiq/memory/backend/sqlite.py
```diff
@@ -0,0 +1,196 @@
...
+        except sqlite3.Error as e:
+            raise SQLiteError(f"Error clearing database: {e}") from e
+
+    def search(self, query: str = None, limit: int = None, filters: dict = None) -> list[Message]:
+        """Searches for messages in SQLite based on the query and/or filters."""
+        limit = limit or self.config.search_limit  # Use default if not provided
+        try:
+            self._validate_table_name()
```
acoola, line 155: It seems this one validation is also not necessary

dynamiq/nodes/agents/base.py
```diff
@@ -150,6 +168,27 @@ def execute(
...
 
+    def _retrieve_memory(self, input_data):
+        """
+        Retrieves memory based on the selected strategy: 'relevant', 'all', or 'both'.
+        """
+        user_id = input_data.get("user_id", None)
+        filters = {"user_id": user_id} if user_id else None
+
+        if self.memory_retrieval_strategy == "relevant":
+            relevant_memory = self.memory.get_search_results_as_string(query=input_data.get("input"), filters=filters)
```
MaksymAI, line 179: Minor: It might be worth extracting the query to a higher level and reusing it here and later for easier code readability
`query = input_data.get("input")`

dynamiq/memory/memory.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def add(self, role: MessageRole, content: str, metadata: dict = None):
+        """Adds a message to the memory."""
+        try:
+            metadata = metadata or {}
+            metadata["timestamp"] = datetime.utcnow().timestamp()
+            message = Message(role=role, content=content, metadata=metadata)
+            self.backend.add(message)
+            logger.debug(
+                f"Memory {self.backend.name}: Added message: {message.role}: {message.content[:min(20, len(message.content))]}..."  # noqa: E501
```
MaksymAI, line 30: What do you think about adding an additional check for non-null content for this debug logging just in case? 
I remember I had similar case and issue later where it was okay for the message to be null (from a tool call or something similar), but my logger for debugging encountered an error because of that)

acoola (approved):

dynamiq/memory/config.py
```diff
@@ -0,0 +1,6 @@
+from pydantic import BaseModel, Field
+
+
+class Config(BaseModel):
```
MaksymAI, line 4: What do you think about renaming `Config` to `MemoryConfig`? Or maybe even removing this config and making `search_limit` and `search_filters` attributes of the Memory class?

It took me some time to understand what `Config` refers to here. Initially, I thought it was related to Node parameters for the workflow.

dynamiq/memory/memory.py
```diff
@@ -0,0 +1,91 @@
...
+
+        If no backend is provided, an InMemory backend is used by default.
+        """
+        if not isinstance(backend, MemoryBackend):
+            raise TypeError("backend must be an instance of Backend")
+        self.config = config
+        self.backend = backend
+
+    def add(self, role: MessageRole, content: str, metadata: dict = None):
+        """Adds a message to the memory."""
```
MaksymAI, line 23: I prefer more extensive docstrings with detailed information about input parameters, which can later be rendered with mkdocs. For instance, what parameters are expected in metadata, etc.
However, this is a minor preference on my part, and it can remain as it is now.

dynamiq/memory/memory.py
```diff
@@ -0,0 +1,91 @@
...
+        """Initializes the Memory with the given configuration and backend.
+
+        If no backend is provided, an InMemory backend is used by default.
+        """
+        if not isinstance(backend, MemoryBackend):
+            raise TypeError("backend must be an instance of Backend")
+        self.config = config
+        self.backend = backend
+
+    def add(self, role: MessageRole, content: str, metadata: dict = None):
```
MaksymAI, line 22:
```suggestion
    def add(self, role: MessageRole, content: str, metadata: Optional[dict] = None):
```

dynamiq/nodes/agents/base.py
```diff
@@ -104,12 +107,15 @@ def _init_prompt_blocks(self):
             "tools": "{tool_description}",
             "instructions": "",
             "output_format": "Provide your answer in a clear and concise manner.",
+            "relevant_information": "{relevant_memory}",
+            "conversation_history": "{context}",
```
MaksymAI, line 111: I am curious if the order of blocks matters. Maybe it's better to first have blocks with relevant information and conversation history, and then add the block with the output format closer to the end so that the LLM remembers it better.

Also, do you think "relevant information" and "relevant_memory" are the same? I would prefer to rename the block to "relevant information from previous conversation."

Additionally, it might be helpful to mention somewhere in the prompt that the chat conversation is for context, so it should only be used if it's helpful for answering, etc.

MaksymAI (approved):
Overall, the initial version looks good to me for merging. Great work!

I left comments in places where I think the code should be unified and reused with existing code. Let's resolve them in the next PR and just keep them here as a reference for what should be improved.

The main question to discuss is whether it's okay for us to simply create a function to convert a Message to a Document and reuse existing methods for writing and retrieving documents, or if we want to refactor and generalize the methods to something like writing and retrieving information while keeping the logic almost the same.

dynamiq/memory/backend/pinecone.py
```diff
@@ -0,0 +1,147 @@
...
+            embedding_result = self.embedder.embed_text(message.content)
+            embedding = embedding_result["embedding"]
+            metadata = {
+                "content": message.content,
+                "role": message.role.value,
+            }
+            if message.metadata:
+                metadata.update(message.metadata)
+            message_id = str(uuid.uuid4())
+            self._index.upsert(vectors=[{"id": message_id, "values": embedding, "metadata": metadata}])
```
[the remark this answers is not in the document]
olbychos: I like the second option

dynamiq/memory/config.py
olbychos: It was adjusted by @delamainer previously we have MemoryConfig

dynamiq/memory/backend/in_memory.py
```diff
@@ -8,6 +11,46 @@ class InMemoryError(Exception):
     pass
 
 
+class BM25:
+    """BM25 implementation for scoring documents."""
+
+    def __init__(self, messages, k1=1.5, b=0.75):
+        """Initialize with a list of messages and parameters for BM25."""
```
MaksymAI, line 18: Let's keep the class implementation in a more general format. Instead of depending on the message format, just use a list of texts.

dynamiq/memory/config.py
MaksymAI: What do you think about moving the parameters of Config into the Memory class as attributes? 
Or do you think it's more beneficial to keep them in a separate config?

olbychos: It would be more beneficial for me to keep them in memory. Let's also move forward with that.

dynamiq/memory/backend/in_memory.py
```diff
@@ -8,6 +11,46 @@ class InMemoryError(Exception):
     pass
 
 
+class BM25:
+    """BM25 implementation for scoring documents."""
+
+    def __init__(self, messages, k1=1.5, b=0.75):
```
acoola: Please add annotations

dynamiq/memory/backend/in_memory.py
```diff
@@ -8,6 +11,46 @@ class InMemoryError(Exception):
     pass
 
 
+class BM25:
```
acoola, line 14: Please improve the naming as `BM25` is not clear.

olbychos: It's just the name of algorithm - standart in field, from here
```
https://en.wikipedia.org/wiki/Okapi_BM25
``

File: dynamiq/memory/backend/in_memory.py [hunk 30]
```diff
@@ -0,0 +1,108 @@
...
+class InMemoryError(Exception):
+    """Base exception class for InMemory backend errors."""
+
+    pass
+
+
+class BM25:
+    """BM25 implementation for scoring documents."""
+
+    def __init__(self, documents: list[str], k1=1.5, b=0.75):
```
acoola, line 17: Please add annotations

dynamiq/memory/backend/in_memory.py
acoola: I mean `BM25DocumentRanker` or smth like this

PR #5, opened 2024-10-03
fix: caching and entity parent_run_id

acoola:
- handle different input types, kwargs, dict order, common workflow kwargs in caching
- fix parent_run_id in case of missed parent

### DIFF
--- a/dynamiq/cache/managers/workflow.py
+++ b/dynamiq/cache/managers/workflow.py
@@ -3,6 +3,7 @@ from typing import Any
 
 from dynamiq.cache.config import CacheConfig
 from dynamiq.cache.managers import CacheManager
+from dynamiq.utils import format_value
 
 
 class WorkflowCacheManager(CacheManager):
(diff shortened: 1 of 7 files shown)


dynamiq/cache/managers/workflow.py
```diff
@@ -3,6 +3,7 @@
 
 from dynamiq.cache.config import CacheConfig
 from dynamiq.cache.managers import CacheManager
+from dynamiq.utils import format_value
 
 
 class WorkflowCacheManager(CacheManager):
```
vitalii-dynamiq, line 9: As part of the cache config, is there a way to set TTL?

acoola: General functionality exists. Should we enable this?

acoola: Here the runnable config with TTL option
https://github.com/dynamiq-ai/dynamiq/blob/eb7f27bbf197afabb44bca223e9019d2ae3717fe/dynamiq/cache/config.py#L23

vitalii-dynamiq (approved):

PR #7, opened 2024-10-11
feat: add react agent, max_loops handler prompt

olbychos:


dynamiq/nodes/agents/react.py
```diff
@@ -245,8 +270,9 @@ class ReActAgent(Agent):
     """Agent that uses the ReAct strategy for processing tasks by interacting with tools in a loop."""
 
     name: str = "React"
-    max_loops: int = Field(default=15, ge=1)
+    max_loops: int = Field(default=15, ge=2)
     inference_mode: InferenceMode = InferenceMode.DEFAULT
+    raise_on_max_loops: bool = Field(default=True, description="Raise exception when max loops are exceeded.")
```
delamainer, line 275: Is there a better name? I cannot say that I like it, but at the same time, I don't have a better suggestion.

@vitalii-dynamiq @acoola, any ideas?

dynamiq/nodes/agents/exceptions.py
```diff
@@ -48,9 +48,14 @@ class InvalidActionException(RecoverableAgentException):
...
     """
-    Exception raised when the agent reaches the maximum number of loops without a final answer.
+    Exception raised when the agent exceeds the maximum number of allowed loops.
+
+    This exception is recoverable, meaning the agent can continue after catching this exception.
     """
 
-    pass
+    def __init__(
+        self, message="Maximum number of loops reached without finding a final answer.", recoverable: bool = True
```
acoola, line 59: Please add annotation for `messages`

dynamiq/nodes/agents/react.py
delamainer: Great idea!

dynamiq/nodes/agents/react.py
```diff
@@ -1,17 +1,24 @@
...
-from dynamiq.nodes.agents.exceptions import ActionParsingException, AgentMaxLoopsReached, RecoverableAgentException
+from dynamiq.nodes.agents.exceptions import ActionParsingException, MaxLoopsExceededException, RecoverableAgentException
 from dynamiq.nodes.node import NodeDependency
 from dynamiq.nodes.types import InferenceMode
 from dynamiq.prompts import Message, Prompt
 from dynamiq.runnables import RunnableConfig, RunnableStatus
 from dynamiq.utils.logger import logger
 
+
+class BehaviorOnMaxLoops(str, Enum):
```
delamainer, line 17: @acoola Do we need to replicate this enum in multiple places? Or should we reuse the same that we already have?

Such things might be common; at least we already have it in multiple places. So, I would probably put it somewhere in the `node` package and use it from there.

acoola: Of course, we should limit duplication, so let's use some common name and move to `nodes/types.py`

acoola (approved):

PR #10, opened 2024-10-16
feat: add search example 

olbychos:
- Include a search example designed with UI and dynamic code integration.
- It can be an advanced version after UI updates involving input/output transformers.

### DIFF
--- a/examples/use_case_search/server_via_dynamiq.py
+++ b/examples/use_case_search/server_via_dynamiq.py
@@ -1,6 +1,6 @@
 """
 This script demonstrates how to build an approach using dynamically deployed workflows.
-It utilizes three workflows: rephrase, search, and answer.
+It utilizes workflow from three steps: rephrase, search, and answer.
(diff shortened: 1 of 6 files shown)


vitalii-dynamiq (approved):

examples/use_case_search/server_via_dynamiq.py
```diff
@@ -0,0 +1,77 @@
+import os
+import re
+
+import requests
+
+from dynamiq.utils.logger import logger
+
+ENDPOINT_REPHRASE = "https://cce4b55a-bd05-46c5-a512-a601b092ff82.apps.sandbox.getdynamiq.ai"  # for example
```
vitalii-dynamiq, line 8: if someone decides to run it these URL will not be available

olbychos: I will move those URLs to the comments section and add a note that we need to deploy them on the UI first.

PR #11, opened 2024-10-16
feat: add ability to define inline prompts in yaml

acoola:


examples/dag/dag.yaml
```diff
@@ -34,7 +30,11 @@ nodes:
     name: OpenAI-1
     model: gpt-3.5-turbo
     connection: openai-conn
-    prompt: openai-ai-prompt
+    prompt:
+      id: openai-ai-prompt
```
delamainer, line 34: Is `id` required?

delamainer (approved):

examples/dag/dag.yaml
delamainer: I don't think we need an `id` in inline prompt.

examples/dag/dag.yaml
```diff
@@ -7,10 +7,6 @@ connections:
     api_key: ${oc.env:ANTHROPIC_API_KEY}
 
 prompts:
-  openai-ai-prompt:
```
vitalii-dynamiq, line 10: keep both examples as we support both

olbychos (approved):

examples/dag/dag.yaml
acoola: It's not required, we autogenerate in case of a miss

examples/dag/dag.yaml
acoola: I left other 3 prompts in `prompts` section, but this one removed as it used as inline prompt

PR #12, opened 2024-10-17
test: Refine Bank Customer Support Example

maksDev123:


examples/customer_support_workflow/bank_rag_tool.py
```diff
@@ -29,7 +29,7 @@ class BankRAGTool(Node):
     text embedding and answer generation.
     """
 
-    name: str = "Bank LLM"
+    name: str = "Bank RAG Tool"
```
vitalii-dynamiq, line 32: can we have better names here? probably Bank FAQ Search?

examples/customer_support_workflow/main.py
```diff
@@ -1,120 +1,69 @@
...
+    # Create connection to OpenAI
+    connection = OpenAIConnection()
+    llm = OpenAI(
+        connection=connection,
+        model="gpt-4o-mini",
+        temperature=0.01,
     )
 
     # Create a ReActAgent for handling bank documentation queries
     agent_bank_documentation = ReActAgent(
```
vitalii-dynamiq, line 24: do we need it as a separate agent? why this RAG tool can't be added to the previous agent? or if we always need to search info in the internal DB, why to wrap it in an agent?

examples/customer_support_workflow/bank_api.py
```diff
@@ -0,0 +1,49 @@
+from flask import Flask, jsonify, request
```
acoola, line 1: We don't use `Flask` in our examples. Please use `Fastapi` if you need API server. We already have fex examples with it and it exists in our examples pkgs group

examples/customer_support_workflow/main.py
```diff
@@ -1,120 +1,69 @@
...
 
-    # Create a ReActAgent for handling internal bank API queries
-    agent_bank_support = ReActAgent(
-        name="Bank Support: Internal API",
-        role="customer support assistant for Internal Bank",
-        goal="help with provided customer requests",
-        llm=llm_react_agent,
-        tools=[BankApiSim(), HumanFeedbackTool()],
+if __name__ == "__main__":
+    # Create connection to OpenAI
```
acoola, line 15: Lest move logic to some function/functions and after just call single func in `if __name__ == "__main__"`

olbychos:
@vitalii-dynamiq, do we want to replicate the behavior here without using orchestrators?
@maksDev123

examples/customer_support_workflow/main.py
maksDev123: Initial Idea is always search in Database. Adding it to agent is not obligatory indeed. But article describes connections between agents and rag, where can be described that agent will take care of proper request (refrase if needed) and format structured output.

maksDev123:
> @vitalii-dynamiq, do we want to replicate the behavior here without using orchestrators?
> @maksDev123

Orchestrator sometimes do not use documentation and goes straight to call API.

examples/customer_support_workflow/bank_rag_tool.py
```diff
@@ -138,16 +138,12 @@ def _use(self, input_text: str) -> str:
...
         answer = flow_result.output.get(self.answer_generation_node.id).get("output")
-
-        logger.info(f"Input: {input_text}")
-        logger.info(f"Output: {answer}")
         return answer
 
-    def execute(
-        self, input_data: dict[str, Any], config: RunnableConfig = None, **kwargs
-    ) -> dict[str, Any]:
+    def execute(self, input_data: dict[str, Any], _: RunnableConfig = None, **kwargs) -> dict[str, Any]:
```
acoola, line 146: It seems here typo `_:`

maksDev123: I removed config to simplify example.

examples/customer_support_workflow/main.py
```diff
@@ -1,120 +1,74 @@
...
+    human_feedback_tool = HumanFeedbackTool()
+
+    # Create a ReActAgent for handling internal bank API queries
+    agent_bank_support = ReActAgent(
+        name="API Agent",
+        role="Customer support assistant with access to Internal Bank API",
+        llm=llm,
+        tools=[api_call, human_feedback_tool],
+        depends=[NodeDependency(node=agent_bank_documentation)],
+        input_transformer=InputTransformer(selector={"input": f"${[agent_bank_documentation.id]}.output.content"}),
```
vitalii-dynamiq, line 65: let's use new way of passing inputs

examples/customer_support_workflow/bank_api.py
```diff
@@ -0,0 +1,42 @@
...
+    if database[card_number]["blocked"]:
+        return {"response": "Card is already blocked."}
+    database[card_number]["blocked"] = True
+    return {"response": "Card was blocked successfully."}
+
+# More endpoints
+
+
+if __name__ == "__main__":
+    uvicorn.run(app, host="127.0.0.1", port=8000)
```
acoola, line 42: Please define host and port as constants and after import them and use in https://github.com/dynamiq-ai/dynamiq/pull/12/files#diff-40b349843c33a55c95c271847d00971e21f6b154cc14b8da24bad513383e742dR34.
This would improve transparency

examples/customer_support_workflow/main.py
```diff
@@ -1,120 +1,76 @@
...
+    # Create api call tool
+    api_call = HttpApiCall(
+        connection=connection,
+        name="Bank API",
+        description="""
+        An internal bank API.
+
+        Available endpoints:
+        * 'block_card' (int card_number, int pin_code)
+        * 'make_transaction' (int card_number_sender, int card_number_reciever, int amoumt)
```
acoola, line 46: Are `make_transaction` and `request_report` still make sense here as we don't have those endpoints?

examples/customer_support_workflow/bank_api.py
vitalii-dynamiq: for examples I think it's okay, let's not overcomplicate

examples/customer_support_workflow/main.py
```diff
@@ -1,120 +1,76 @@
...
+        role="Customer support assistant for Internal Bank Documentation",
+        llm=llm,
+        tools=[BankRAGTool()],
     )
 
-    # Create a ManagerAgent to oversee the workflow
-    agent_manager = LinearAgentManager(llm=llm_agent)
+    # Create connection to Bank API
+    connection = HttpConnection(
+        method="POST",
```
acoola, line 33: Not sure if all APIs would be POST, but as a simple example it make sense.
For more realistic example Tool/Agent should be able to operate with different endpoints (maybe make sense to pass swagger and so on). But this it another level of complexity.

maksDev123: I was thinking about another version of this tool.


    @function_tool
    def bank_api(parameters: dict[str, str]):
        """
        An internal bank API.

        Available endpoints:
        * 'block_card' (int card_number, int pin_code)
        * 'make_transaction' (int card_number_sender, int card_number_reciever, int amoumt)
        * 'request_report' (int userId, int pin_code)

        As name of tool use - bank_api. Name of required endpoint has to be provided as parameter. 
        """
        if endpoint := parameters.get('endpoint', ""):
            return requests.post( f"http://127.0.0.1:5004/{endpoint}",
                                    json=parameters).json()['response']
        return "No endpoint chosen, no parameter with name 'endpoint' was passed."

examples/customer_support_workflow/main.py
maksDev123: Wanted to leave example as simple as possible. I will add very simplified version of 2 rest endpoints.

examples/customer_support_workflow/bank_rag_tool.py
olbychos, line 1: Let's use our current rag tool.
For example: https://github.com/dynamiq-ai/dynamiq/blob/main/examples/agents/use_react_wf.py#L132
And tool is retriever:
https://github.com/dynamiq-ai/dynamiq/blob/main/dynamiq/nodes/tools/retriever.py

olbychos (changes requested):

examples/customer_support_workflow/bank_rag_tool.py
maksDev123: Good point, used it.

acoola (approved):

olbychos (approved):

PR #13, opened 2024-10-17
fix: vector storages connections and creation flag

acoola:
- pass connection to vector store component as we don’t do this by default
- change `create_if_not_exist` flag default value to `False` to skip creation for retrievers

### DIFF
--- a/dynamiq/storages/vector/chroma/chroma.py
+++ b/dynamiq/storages/vector/chroma/chroma.py
@@ -41,7 +41,7 @@ class ChromaVectorStore:
         connection: Chroma | None = None,
         client: Optional["ClientAPI"] = None,
         index_name: str = "default",
-        create_if_not_exist: bool = True,
+        create_if_not_exist: bool = False,
     ):
         """
         Initialize the ChromaVectorStore.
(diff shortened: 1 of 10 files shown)


dynamiq/storages/vector/chroma/chroma.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         connection: Chroma | None = None,
         client: Optional["ClientAPI"] = None,
         index_name: str = "default",
-        create_if_not_exist: bool = True,
+        create_if_not_exist: bool = False,
```
vitalii-dynamiq, line 44: I don't know if this the right name? first it's not clear what does it create.
maybe:
- auto_create_index
- create_index_if_absent
- create_index_if_missing
- initialize_index

@acoola, @MaksymAI, @delamainer

acoola: Generally I see this is the common name and each vector store handles the creation of needed resources:
- Chroma creates collection with proper index
- Pinecone creates index
- Qdrant creates collection + index
- Weaviate creates collection

delamainer: Personally, I don't have strong opinions on naming here. Maybe you right and it's not clear what it creates.

`create_index_if_not_exist`?

MaksymAI (approved):

PR #16, opened 2024-10-18
feat: add example llm with structured output

olbychos:


examples/llm/structured_output.py
```diff
@@ -0,0 +1,81 @@
...
+        return parsed_event
+    except ValidationError as e:
+        print(f"Validation Error: {e}")
+        print("Raw Response:", response)
+        return None
+
+
+def run_openai_node(prompt: Prompt, schema: dict, inference_mode: InferenceMode):
+    openai_node = OpenAI(model="gpt-4o-mini")
+    response = openai_node.run(input_data={}, schema=schema, prompt=prompt, inference_mode=inference_mode)
```
vitalii-dynamiq, line 29: is schema only supported in run method? not in the init?

olbychos: Yes, only in run
https://github.com/dynamiq-ai/dynamiq/blob/e8f4f77b81ce0f8a92051acca9eed9c67b221aa3/dynamiq/nodes/llms/base.py#L35C1-L54C8

Add this to init? Like inference mode also

acoola: as far as I remember we planned to add `schema` to Input and Output nodes, but it seems this not related to the LLM schema

vitalii-dynamiq: this is related to a different topic.
@olbychos we also need to support this in the init

olbychos: Okay, I’ll add it and let’s merge afterwards

dynamiq/nodes/llms/base.py
```diff
@@ -51,6 +51,9 @@ class BaseLLM(ConnectionNode):
         presence_penalty (float | None): Penalize new tokens based on their existence in the text.
         frequency_penalty (float | None): Penalize new tokens based on their frequency in the text.
         tool_choice (str | None): Value to control which function is called by the model.
+        inference_mode (InferenceMode): Mode of inference. Defaults to InferenceMode.DEFAULT.
+        schema_ (Dict[str, Any]): schema_ for structured output or function calling. Defaults to empty dict.
```
vitalii-dynamiq, line 55: why schema_ with underscore?

olbychos: We encountered an error with Pydantic validation related to the 'schema' attribute due to Pydantic's internal methods. Therefore, I have decided to use aliases instead.

acoola: `schema` is a deprecated function, but it still exists. So we defined `schema_` but with `schema` alias. Fo for initialization/json manipulation and so on, it looks like `schema`.

dynamiq/nodes/llms/base.py
```diff
@@ -51,6 +51,11 @@ class BaseLLM(ConnectionNode):
         presence_penalty (float | None): Penalize new tokens based on their existence in the text.
         frequency_penalty (float | None): Penalize new tokens based on their frequency in the text.
         tool_choice (str | None): Value to control which function is called by the model.
+        inference_mode (InferenceMode): Determines how the model handles inference tasks and formats outputs.
+        - InferenceMode.DEFAULT: Generates unstructured, free-form natural language text.
+        - InferenceMode.STRUCTURED_OUTPUT: Produces structured JSON output.
+        - InferenceMode.FUNCTION_CALLING: Structured output for tools (functions) to be called.
+        schema_ (Dict[str, Any]): schema_ for structured output or function calling. Defaults to empty dict.
```
acoola, line 58: Please update the docstring with correct types - `dict[str, Any] | type[BaseModel] | None`

dynamiq/nodes/llms/base.py
```diff
@@ -199,13 +208,36 @@ def _handle_streaming_completion_response(
...
+            inference_mode (InferenceMode): The inference mode to use.
+            schema (Dict[str, Any]): The schema to use.
+
+        Returns:
+            tuple[Optional[Dict[str, Any]], Optional[Dict[str, Any]]]: Response format and tools.
+        """
+        response_format = None
+        tools = None
+
+        match inference_mode:
```
acoola, line 226: It would be great to validate that `schema` is not None in case of InferenceMode.STRUCTURED_OUTPUT and InferenceMode.FUNCTION_CALLING

acoola (approved):

PR #17, opened 2024-10-21
feat: update pinecone params

MaksymAI:


olbychos (approved):
I've just updated the process to create a client if it doesn't exist and to return the correct client.

delamainer (approved):

acoola (approved):

PR #19, opened 2024-10-23
fix: python and firecrawl tools

acoola:


olbychos (approved):

dynamiq/nodes/tools/firecrawl.py
```diff
@@ -147,6 +147,6 @@ def execute(
                 "warning": scrape_result.get("data", {}).get("warning"),
             }
 
-        logger.debug(f"Tool {self.name} - {self.id}: finished with result {str(result[:200])}...")
+        logger.debug(f"Tool {self.name} - {self.id}: finished with result {str(result)[:200]}...")
```
vitalii-dynamiq, line 150: it will break if there's less than 200 elements

acoola: Nope:
```
>>> a = "12345"
>>> a[:10]
'12345'
>>> a[:1000]
'12345'
```

PR #21, opened 2024-10-24
feat: memory refactor

olbychos:


examples/memory/demo_memory_pinecone.py
```diff
@@ -42,5 +41,5 @@
 print("Results with filter only:", [r.content for r in results])
 
 # Clear memory
-memory.clear()
-print("Is memory empty?", memory.is_empty())
+# memory.clear()
```
vitalii-dynamiq, line 44: let's remove if not needed

olbychos: It was unintentional; we need it for demo purposes.

olbychos:
@acoola @delamainer @maksDev123 @MaksymAI kindly reminder to review

vitalii-dynamiq:
@olbychos let's add memory example to README also

dynamiq/memory/backend/qdrant.py
```diff
@@ -1,13 +1,14 @@
...
-from qdrant_client.http.exceptions import ApiException, UnexpectedResponse
-from qdrant_client.models import FieldCondition, Filter, MatchAny, MatchValue, Range
+from qdrant_client.http.exceptions import UnexpectedResponse
 
 from dynamiq.components.embedders.base import BaseEmbedder
 from dynamiq.connections import Qdrant as QdrantConnection
 from dynamiq.memory.backend.base import MemoryBackend
 from dynamiq.prompts import Message
+from dynamiq.storages.vector.policies import DuplicatePolicy
+from dynamiq.storages.vector.qdrant import QdrantVectorStore
```
vitalii-dynamiq, line 10: @olbychos should we also add Weaviate?

olbychos:
> @olbychos let's add memory example to README also

Did you mean gitbook or readme under this folder?Because we have example of use memory here: https://github.com/dynamiq-ai/dynamiq/tree/main/examples/memory

delamainer (approved):

dynamiq/memory/backend/qdrant.py
olbychos: @vitalii-dynamiq we can add this, but this optional
Let’s proceed with overall implementation memory and stabilize it, and afterwards we can easily in future add new types and so on

README.md
```diff
@@ -372,6 +372,48 @@ answer = result.output.get(answer_generator.id).get("output", {}).get("content")
...
+```python
+from dynamiq.memory import Memory
+from dynamiq.memory.backend.in_memory import InMemory
+from dynamiq.nodes.agents.simple import SimpleAgent
+from dynamiq.connections import OpenAI as OpenAIConnection
+from dynamiq.nodes.llms import OpenAI
+
+AGENT_ROLE = ("helpful assistant,"
+              "goal is to provide useful information and answer questions")
+llm = OpenAI(
```
acoola, line 387: Please reformat in python file that code snippets and paste here

dynamiq/memory/backend/pinecone.py
```diff
@@ -1,153 +1,150 @@
...
 
     def __init__(
         self,
         connection: PineconeConnection,
         embedder: BaseEmbedder,
-        cloud: str,
-        region: str,
         index_name: str = "conversations",
+        namespace: str = "default",
+        index_type: PineconeIndexType | None = None,
```
acoola, line 26: with `create_if_not_exist=True`, `index_type` can't be Null. Still we handle this case and user would see error during init if index not exists and `index_type` is Null. So probably could ignore this

examples/memory/demo_memory_pinecone.py
```diff
@@ -1,19 +1,18 @@
...
 from dynamiq.components.embedders.openai import OpenAIEmbedder
 from dynamiq.connections import Pinecone as PineconeConnection
 from dynamiq.memory import Memory
 from dynamiq.memory.backend import Pinecone
 from dynamiq.prompts import MessageRole
 
+
 pinecone_connection = PineconeConnection()
 embedder = OpenAIEmbedder(dimensions=1536)
+
```
acoola, line 10: This works just if the default index already exists.

tests/integration/nodes/writers/test_qdrant_writer_flow.py
```diff
@@ -90,6 +90,5 @@ def test_write_workflow(
     )
     assert response.output == expected_output
 
-    mock_set_up_collection.assert_called_once()
```
acoola, line 93: why did you remove this check?

olbychos: There was an error with setting up the collection while rewriting the documents.
https://github.com/dynamiq-ai/dynamiq/pull/21/files/27826cf68cb6b5485e37b5df1affe52235f36a68#diff-f0eb78186b536379a2b20893abe3fe3f2460976ece7e1f525afeff0c466809e9L337

examples/memory/demo_memory_pinecone.py
olbychos: Based on your previous comment, if there is an error, we will set up 'index_type.' What do you suggest?

dynamiq/memory/backend/pinecone.py
olbychos: Alternatively, set the default value of index_type to 'serverless'.
WDYT?

tests/integration/nodes/writers/test_qdrant_writer_flow.py
acoola: So if we disabled rewriting the documents and this mock is not called let's change it to 
`mock_set_up_collection.assert_not_called()`

dynamiq/memory/backend/pinecone.py
```diff
@@ -1,153 +1,150 @@
...
-                response = self._index.query(
-                    vector=dummy_vector,
-                    top_k=limit,
-                    include_metadata=True,
-                    filter=normalized_filters,
+                documents = self.vector_store._embedding_retrieval(
+                    query_embedding=dummy_vector,
+                    namespace=self.namespace,  # Add namespace
+                    filters=normalized_filters,
+                    top_k=limit or 10,
```
MaksymAI, line 128: Let's move this constant to the function's default arguments.
```suggestion
                    top_k=limit,
```

dynamiq/memory/backend/qdrant.py
```diff
@@ -21,122 +22,167 @@ class Qdrant(MemoryBackend):
...
         self.index_name = index_name
         self.embedder = embedder
 
         try:
-            self.client = self.connection.connect()
+            self.vector_store = QdrantVectorStore(
+                connection=connection,
+                index_name=index_name,
+                dimension=embedder.dimensions,
+                create_if_not_exist=True,
```
MaksymAI, line 49: I would include these parameters at the initialization level to allow the user to decide whether to create the index or not, instead of hard-coding it to always allow index creation
```suggestion
                create_if_not_exist=create_if_not_exist,
```

dynamiq/memory/backend/qdrant.py
```diff
@@ -21,122 +22,167 @@ class Qdrant(MemoryBackend):
...
+            self.vector_store._set_up_collection(
+                collection_name=index_name,
+                embedding_dim=embedder.dimensions,
+                create_if_not_exist=True,
+                recreate_collection=False,
+                similarity=metric,
+                use_sparse_embeddings=False,
+                sparse_idf=False,
+                on_disk=on_disk,
+            )
```
MaksymAI, line 67: Do we really need to duplicate this code? 
I was thinking that the collection setup would be executed during init or actual client usage with parameters from the init

MaksymAI (approved):
Good job!

dynamiq/memory/backend/qdrant.py
olbychos: Updated

dynamiq/memory/backend/pinecone.py
MaksymAI: Please update it in the next PR!)

dynamiq/memory/backend/qdrant.py
MaksymAI: Please update it in the next PR!)

PR #26, opened 2024-10-28
feat: Tool abstraction

maksDev123:


dynamiq/nodes/agents/react.py
```diff
@@ -498,6 +458,100 @@ def _extract_final_answer_xml(self, llm_output: str) -> str:
         logger.info(f"Agent {self.name} - {self.id}: Final answer found: {final_answer['answer']}")
         return final_answer["answer"]
 
+    def generate_input_formats(self, tools: list[ToolMixin]) -> str:
+        input_formats = []
+        for tool in tools:
+            params = [
+                " ".join(param[:-1]) for param in tool.input_params if param[-1].get("is_accessible_to_agent", True)
+            ]
+            input_formats.append(f" - {tool.name}\n" f"\t* {"\n\t* ".join(params)}")
```
olbychos, line 467: It seems there are errors with quotes

dynamiq/nodes/agents/react.py
```diff
@@ -261,6 +223,7 @@ class ReActAgent(Agent):
         default=Behavior.RAISE,
         description="Define behavior when max loops are exceeded. Options are 'raise' or 'return'.",
     )
+    format_shema: list = []
```
delamainer, line 226: `format_schema`

dynamiq/nodes/tools/basetool.py
```diff
@@ -0,0 +1,45 @@
...
+
+from dynamiq.nodes import NodeGroup
+from dynamiq.runnables import RunnableConfig
+
+
+class DefaultInputSchema(BaseModel, ABC):
+    input: str
+
+
+class ToolMixin:
```
delamainer, line 14: Should this be an abstract class?

delamainer: What about naming? `BaseTool`?

dynamiq/nodes/tools/basetool.py
```diff
@@ -0,0 +1,45 @@
...
+    group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
+    name: str = "base_tool"
+    description: str = "base tool"
+    input_schema: type[BaseModel] = Field(default=DefaultInputSchema)
+
+    @field_validator("input_schema")
+    def passwords_match(cls, v: BaseModel, info: ValidationInfo) -> BaseModel:
```
delamainer, line 21: `passwords`?

dynamiq/nodes/tools/basetool.py
```diff
@@ -0,0 +1,45 @@
...
+                tags = field.json_schema_extra
+            description = field.description or "No description"
+            field_type: str = field.annotation.__name__
+
+            params.append((name, field_type, description, tags))
+        return params
+
+    def execute(self, input_data: dict[str, Any], config: RunnableConfig | None = None, **kwargs) -> dict[str, Any]:
+        """Executes the requested action based on the input data."""
+        return self.run_tool(self.input_schema(**input_data), config=config, **kwargs)
```
delamainer, line 45: Should we perform the schema validation before running the tool?

dynamiq/nodes/tools/e2b_sandbox.py
```diff
@@ -68,14 +64,14 @@ class E2BInterpreterTool(ConnectionNode):
...
-    group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
-    name: str = "code-interpreter_e2b"
+    name: str = "code-interpreter-e2b"
     description: str = DESCRIPTION
     connection: E2BConnection
     installed_packages: list = []
     files: list[tuple[str | bytes, str]] | None = None
     persistent_sandbox: bool = True
     _sandbox: Sandbox | None = None
+    input_schema: type[E2BInterpreterInputSchema] = E2BInterpreterInputSchema
```
delamainer, line 74: I think such fields should be private. `_input_schema`?

dynamiq/nodes/tools/basetool.py
delamainer: [Here](https://github.com/dynamiq-ai/dynamiq/blob/c8f66ac6b8c27e3f5d91912fb8bf6051bd7f13e7/dynamiq/nodes/agents/base.py#L61), it should be: `tools: list[BaseTool] = []`

delamainer (dismissed):
I think we need to have such schemas for all nodes. 

Now we have this in all node, which make it hard to use:
```python3
def execute(
    self, input_data: dict[str, Any], config: RunnableConfig | None = None, **kwargs
) -> dict[str, list[Any]]:
```

It would be great to have something like this:
```python3
def execute(
    self, input_data: LLMDocumentRankerInputSchema, config: RunnableConfig | None = None, **kwargs
) -> LLMDocumentRankerOutputSchema:
```

dynamiq/nodes/agents/react.py
maksDev123: It is just implicit string concatenation but flattened to one line. Or you mean something else?

olbychos: @maksDev123 
```
    line 467
    input_formats.append(f" - {tool.name}\n" f"\t* {"\n\t* ".join(params)}")
                                                      ^
SyntaxError: unexpected character after line continuation character
```

It might be better not to flatten it to a single line.

olbychos: Update:
```
input_formats.append(
                f" - {tool.name}\n \t* {{\n\t* " + "\n\t* ".join(params) + "}}"
            )
 ```

**Comment by acoola** [comment]
Small notes and steps that we discussed with @maksDev123 some time ago:
1. Add input schema for tools (this PR)
2. Add input schema for the rest nodes (could be done group by group)
3. Move input params validation based on input schema into base `Node.run`. As a result we are going to validate before execution/retries
4. Add output schema in the same way

File: dynamiq/nodes/tools/e2b_sandbox.py (hunk 6 above)
Comment (maksDev123) [reply dynamiq/nodes/tools/e2b_sandbox.py:74]: I guess user might want to get this schema, to do some validation or whatever.
@acoola ?

File: dynamiq/nodes/tools/basetool.py (hunk 5 above)
Comment (maksDev123) [reply dynamiq/nodes/tools/basetool.py:45]: Make sense, schema validation can be done in Node.run. This is something that should be implemented for all Nodes at once. So I think we can leave it on next PRs.

File: dynamiq/nodes/tools/e2b_sandbox.py (hunk 6 above)
Comment (olbychos) [reply dynamiq/nodes/tools/e2b_sandbox.py:74]: Maybe we need to keep it private, but use public getter and setter methods
```
 _input_schema: Type[E2BInterpreterInputSchema] = E2BInterpreterInputSchema
 
    @property
    def input_schema(self) -> Type[E2BInterpreterInputSchema]:
        """
        Provides access to the input schema.
        """
        return self._input_schema

    @input_schema.setter
    def input_schema(self, schema: Type[E2BInterpreterInputSchema]):
        """
        Sets a new input schema, allowing flexibility in validation needs.
        """
        if not isinstance(schema, type):
            raise TypeError("input_schema must be a type.")
        #validation can be also here
        self._input_schema = schema
```

Comment (acoola) [reply dynamiq/nodes/tools/e2b_sandbox.py:109]: We don't need getter and setter. This schema should be as Constant, so try to use ClassVar to define this.
So user should be able to get node schema without initialization too:
```
schema = E2BInterpreterTool.input_schema
tool = E2BInterpreterTool(...)
assert schema == tool.input_schema
```

File: dynamiq/nodes/tools/basetool.py [hunk 7]
```diff
@@ -0,0 +1,52 @@
+from abc import ABC, abstractmethod
```
acoola, line 1: Let's use `base.py` name similar to others packages

dynamiq/nodes/tools/basetool.py
acoola: Why do you use tool mixing and don't apply changes to `Node`?
It seems make sense to update `Node` and add `input_schema` to base class now (Optional for now).
Also makes sense to define `validate_input_schema` with validation Node if input_schema exists.

dynamiq/nodes/tools/basetool.py
```diff
@@ -0,0 +1,45 @@
...
+
+class DefaultInputSchema(BaseModel, ABC):
+    input: str
+
+
+class ToolMixin:
+    group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
+    name: str = "base_tool"
+    description: str = "base tool"
+    input_schema: type[BaseModel] = Field(default=DefaultInputSchema)
```
acoola: Please use class variable, not intance variable. So schema available on class level too
```suggestion
    input_schema: ClassVar[type[BaseModel]] = DefaultInputSchema
```

dynamiq/nodes/tools/basetool.py
```diff
@@ -0,0 +1,45 @@
...
+    def passwords_match(cls, v: BaseModel, info: ValidationInfo) -> BaseModel:
+        # Check if input_schema is correct
+        return v
+
+    @abstractmethod
+    def run_tool(self):
+        pass
+
+    @property
+    def input_params(self) -> str:
```
acoola: As schema on class level, this helper also should be classmethod

dynamiq/nodes/tools/basetool.py
acoola: We shouldn't allow to define own schema for nodes that have static/constant schema. So I don't see sene in this helper here.

dynamiq/nodes/tools/basetool.py
```diff
@@ -0,0 +1,45 @@
...
+            tags = {}
+            if field.json_schema_extra:
+                tags = field.json_schema_extra
+            description = field.description or "No description"
+            field_type: str = field.annotation.__name__
+
+            params.append((name, field_type, description, tags))
+        return params
+
+    def execute(self, input_data: dict[str, Any], config: RunnableConfig | None = None, **kwargs) -> dict[str, Any]:
```
acoola, line 47: We don't need to make any renaming functions on Node level and so on. 
You could use decorators to handle those validations before `execute`.
But I recommended to move this logic to `Node` class

dynamiq/nodes/tools/basetool.py
maksDev123: Class properties are deprecated, so I added separate `get_input_schema` method

dynamiq/nodes/node.py
```diff
@@ -256,9 +257,15 @@ def _validate_dependency_status(depend: NodeDependency, depends_result: dict[str
...
-                failed_depend=depend, message=f"Dependency {depend.node.id}: skipped"
-            )
+            raise NodeSkippedException(failed_depend=depend, message=f"Dependency {depend.node.id}: skipped")
+
+    @property
+    def input_schema(self) -> BaseModel | None:
+        return self._input_schema
+
+    @classmethod
+    def get_input_schema(cls) -> BaseModel | None:
```
acoola, line 267: It's redundant. `Node.input_schema` and `node_instance.input_schema` should be enough

dynamiq/nodes/node.py
```diff
@@ -329,6 +336,40 @@ def validate_depends(self, depends_result):
                     depend=dep, depends_result=depends_result
                 )
 
+    @property
+    def input_schema_params(self) -> list[tuple]:
+        """
+        Return list of input parameters along with their type and description.
+        """
+        params = []
+        for name, field in self.input_schema.model_fields.items():
```
acoola, line 345: Why just `self.input_schema.model_fields` dict not fit as the return of this func. It handles all other Field attrs (e.g title, frozen, required, ....)?

dynamiq/nodes/node.py
```diff
@@ -369,6 +410,8 @@ def transform_input(self, input_data: dict, depends_result: dict[Any, RunnableRe
             else:
                 inputs[key] = value
 
+        inputs = self.validate_input_schema(inputs)
```
acoola, line 413: Here you have 2 ways:
1. Just use `self.validate_input_schema(inputs)` as still return dict. So you just going to validate and raise error if inputs are invalid. 
2. Transform inputs from `dict` to `input_schema` model by `inputs = self.validate_input_schema(inputs). But in this case all Nodes.execure with input_schema should handle this case. Also callbacks methods like `run_on_node_start` should handle `schema` too (not just `dict`)

dynamiq/nodes/tools/e2b_sandbox.py
```diff
@@ -251,12 +258,9 @@ def _execute_shell_command(self, command: str, sandbox: Sandbox | None = None) -
             raise ToolExecutionException(f"Error during shell command execution: {output.stderr}", recoverable=True)
         return output.stdout
 
-    def execute(self, input_data: dict[str, Any], config: RunnableConfig | None = None, **kwargs) -> dict[str, Any]:
+    def execute(self, input_data: dict[str, Any], config: RunnableConfig | None = None, **_) -> dict[str, Any]:
         """Executes the requested action based on the input data."""
-        logger.debug(f"Tool {self.name} - {self.id}: started with input data {input_data}")
-
         config = ensure_config(config)
-        self.run_on_node_execute_run(config.callbacks, **kwargs)
```
acoola, line 259: Why did you remove this one?

dynamiq/nodes/tools/function_tool.py
```diff
@@ -102,6 +96,13 @@ def function_tool(func: Callable[..., T]) -> type[FunctionTool[T]]:
     :param func: Function to be converted into a tool.
     :return: A FunctionTool subclass that wraps the provided function.
     """
+    def create_input_schema(func) -> type[BaseModel]:
+        signature = inspect.signature(func)
+        params_dict = {param.name: (param.annotation, param.default) for param in signature.parameters.values()}
+        return create_model(
```
acoola, line 102: Looks good, but please double-check if this works properly with all cases.

dynamiq/nodes/tools/http_api_call.py
```diff
@@ -16,6 +18,25 @@ class ResponseType(str, enum.Enum):
     JSON = "json"
 
 
+class HttpApiCallInputSchema(BaseModel):
+    data: dict = Field(default={}, description="Parameter to provide main json payload.")
```
acoola, line 22: It's just payload, not always JSON.

dynamiq/nodes/tools/python.py
```diff
@@ -42,6 +43,11 @@ def restricted_import(name, globals=None, locals=None, fromlist=(), level=0):
     return importlib.import_module(name)
 
 
+class PythonInputSchema(BaseModel):
+    class Config:
+        extra = "allow"
```
acoola, line 48: Please use pudantic V2 suggested model config
```suggestion
    model_config = ConfigDict(extra="allow")
```

dynamiq/nodes/tools/scale_serp.py
```diff
@@ -10,6 +10,21 @@
...
+    url: str = Field(default="", description="Parameter to provide a search url.")
+    limit: str = Field(
+        default="", description="Parameter to specify the number of results to return, by default is set to 10."
+    )
+
+    @model_validator(mode="after")
+    def validate_query_url(self):
+        """Validate that either query or url is specified"""
+        if not self.url and not self.query:
+            raise ValueError("Either query or url has to be specified.")
```
acoola, line 24:
```suggestion
            raise ValueError("Either 'query' or 'url' has to be specified.")
```

dynamiq/nodes/tools/zenrows.py
```diff
@@ -9,20 +9,22 @@
 from dynamiq.utils.logger import logger
 
 
+class ZenRowsInputSchema(BaseModel):
+    url: str = Field(..., description="Parameter to provide a url of the page to scrape.")
```
acoola, line 13: `url` is required by this. Is this expected? In such case there is no sense in ZenRowsTool.url

dynamiq/nodes/node.py
```diff
@@ -217,6 +217,7 @@ class Node(BaseModel, Runnable, ABC):
     _output_references: NodeOutputReferences = PrivateAttr()
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
+    _input_schema: ClassVar[type[BaseModel] | None] = None
```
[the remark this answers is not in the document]
acoola: It can't be changed or overridden on instance creation
If you are meaning changing on class level, the both could be changed if user would like to break something.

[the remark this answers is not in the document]
acoola:
```
from pydantic import BaseModel, Field

class B(BaseModel):
    b: str = Field(default="bee", frozen=True, description="Bee is b", title="B")

class A(BaseModel):
    a: ClassVar[B] = B
    _a: ClassVar[B] = B

a = A(a="test", _a="test")
```
```
>>> print(a)

>>> a.a
<class '__main__.B'>
>>> a._a
<class '__main__.B'>
```

dynamiq/nodes/tools/zenrows.py
maksDev123: Yes it is required as it is the only parameter. `url` can still be passed but be empty, so here behaviour is the same.

maksDev123: I am not sure what is was used for to provide ZenRowsTool.url

dynamiq/nodes/node.py
maksDev123: Should I stick to second approach?

dynamiq/nodes/tools/zenrows.py
olbychos: The URL can be a parameter from the input or from the node initialization.

dynamiq/nodes/node.py
```diff
@@ -225,6 +227,13 @@ def __init__(self, **kwargs):
 
         self._output_references = NodeOutputReferences(node=self)
 
+    @field_validator("name")
+    def sanitize_name(s: str):
```
acoola, line 231: Do we have any issues with naming? E.g I would to use `[Andrii] OpenAI` name. Any reason why I can't?

dynamiq/nodes/node.py
```diff
@@ -490,12 +516,21 @@ def run(
         try:
             transformed_input = self.transform_input(input_data=input_data, depends_result=depends_result)
 
+            try:
```
acoola, line 519: Please move this to `def transform_input`

dynamiq/nodes/node.py
```diff
@@ -490,12 +516,21 @@ def run(
...
+                transformed_input = self.validate_input_schema(transformed_input)
+            except NodeException as e:
+                raise RecoverableAgentException(f"Error occurred while parsing input: {e}.")
+
+            if isinstance(transformed_input, BaseModel):
```
acoola, line 524: Please handle different data types inside `run_on_node_start`

dynamiq/nodes/node.py
```diff
@@ -550,9 +583,11 @@ def execute_with_retry(
         n_attempt = self.error_handling.max_retries + 1
         for attempt in range(n_attempt):
             merged_kwargs = merge(kwargs, {"execution_run_id": uuid4()})
-            self.run_on_node_execute_start(
-                config.callbacks, input_data, **merged_kwargs
-            )
+
+            if isinstance(input_data, BaseModel):
```
acoola, line 587: Please handle different data types inside `run_on_node_execute_start`

dynamiq/nodes/tools/e2b_sandbox.py
```diff
@@ -81,6 +68,24 @@ def generate_file_description(file: bytes | io.BytesIO, length: int = 20) -> str
     return f"File starting with: {file_content.hex()}"
 
 
+class E2BInterpreterInputSchema(BaseModel):
+    packages: str = Field(default="", description="Parameter to provide packages to install.")
+    shell_command: str = Field(default="", description="Parameter to provide shell command to execute.")
+    python: str = Field(default="", description="Parameter to provide python code to execute.")
+    files: Any = Field(
```
acoola, line 75: Shouldn't files be a `list[io.BytesIO]` instead of Any?

dynamiq/nodes/tools/function_tool.py
```diff
@@ -17,15 +17,15 @@ class FunctionTool(Node, Generic[T]):
     """
 
     group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
-    name: str = Field(default="Function Tool")
+    name: str = Field(default="function-tool")
```
acoola, line 20: Field is obsolete here:
```suggestion
    name: str ="function-tool"
```

dynamiq/nodes/tools/http_api_call.py
```diff
@@ -32,7 +53,7 @@ class HttpApiCall(ConnectionNode):
         response_type(ResponseType|str): The type of response content.
     """
 
-    name: str = "The name of the API call tool"
+    name: str = "Api call tool"
```
acoola, line 56: Why some nodes name using human like words in naming and few nodes renamed to `tool-name` style?

dynamiq/nodes/tools/zenrows.py
acoola: So this need to fix

dynamiq/nodes/tools/human_feedback.py
```diff
@@ -137,16 +141,11 @@ def execute(self, input_data: dict[str, Any], config: RunnableConfig | None = No
...
         config = ensure_config(config)
         self.run_on_node_execute_run(config.callbacks, **kwargs)
 
-        if "input" not in input_data:
-            raise ToolExecutionException(
-                "Input data must contain an 'input' key with the prompt for the user.", recoverable=True
-            )
-
-        input_text = input_data["input"]
+        input_text = input_data.question
```
acoola, line 148: Is this change from "input" to "question" safe?

dynamiq/nodes/node.py
maksDev123: Yes because:

1) On UI we can only provide name in this format
2) Function calling require this format.
3) It transforms to appropriate format automatically.
4) I guess it is better to keep names in this format, easier to read in traces.

dynamiq/nodes/tools/human_feedback.py
acoola: I see this change break at least few examples

dynamiq/nodes/tools/firecrawl.py
```diff
@@ -30,17 +30,21 @@ class ExtractorOptions(BaseModel):
     extraction_schema: dict | None = None
 
 
+class FirecrawlInputSchema(BaseModel):
+    url: str = Field(..., description="Parameter to specify the url of the page to be scraped.")
```
acoola, line 34: This field is required, so `FirecrawlTool.url` became unusable. Please make it oprional

maksDev123: It functions as previously. In case url will be empty we will stick to FirecrawlTool.url

maksDev123:
> I think we need to have such schemas for all nodes.
> 
> Now we have this in all node, which make it hard to use:
> 
> ```python
> def execute(
>     self, input_data: dict[str, Any], config: RunnableConfig | None = None, **kwargs
> ) -> dict[str, list[Any]]:
> ```
> 
> It would be great to have something like this:
> 
> ```python
> def execute(
>     self, input_data: LLMDocumentRankerInputSchema, config: RunnableConfig | None = None, **kwargs
> ) -> LLMDocumentRankerOutputSchema:
> ```



> I think we need to have such schemas for all nodes.
> 
> Now we have this in all node, which make it hard to use:
> 
> ```python
> def execute(
>     self, input_data: dict[str, Any], config: RunnableConfig | None = None, **kwargs
> ) -> dict[str, list[Any]]:
> ```
> 
> It would be great to have something like this:
> 
> ```python
> def execute(
>     self, input_data: LLMDocumentRankerInputSchema, config: RunnableConfig | None = None, **kwargs
> ) -> LLMDocumentRankerOutputSchema:
> ```



> I think we need to have such schemas for all nodes.
> 
> Now we have this in all node, which make it hard to use:
> 
> ```python
> def execute(
>     self, input_data: dict[str, Any], config: RunnableConfig | None = None, **kwargs
> ) -> dict[str, list[Any]]:
> ```
> 
> It would be great to have something like this:
> 
> ```python
> def execute(
>     self, input_data: LLMDocumentRankerInputSchema, config: RunnableConfig | None = None, **kwargs
> ) -> LLMDocumentRankerOutputSchema:
> ```

Works this way now

dynamiq/nodes/tools/zenrows.py
```diff
@@ -9,33 +9,34 @@
...
     group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
-    name: str = "Scraper-ZenRows"
+    name: str = "Zenrows Scraper Tool"
     description: str = (
         "A tool for scraping web pages, powered by ZenRows. "
         "You can use this tool to scrape the content of a web page."
-        "Input should be a dictionary with a key 'input' containing the URL of the page to scrape."
     )
     connection: ZenRows
-    url: str | None = None
```
acoola, line 28: Please revert this one. It usable by lib and also exists on UI. Also please fix `ZenRowsInputSchema.url` to make it optional

dynamiq/nodes/tools/firecrawl.py
maksDev123: Yes, If it is missed, but not when it present but empty.

maksDev123: I made this parameter in schema optional.

acoola (approved):
Great job!

PR #30, opened 2024-11-05
feat: add baseline streaming for agents

olbychos:
add baseline streaming support for react agent with ability to stream intermediate steps/final answer only, for simple/reflection agent


dynamiq/nodes/agents/base.py
```diff
@@ -57,7 +57,9 @@ class Agent(Node):
     llm: Node = Field(..., description="LLM used by the agent.")
     group: NodeGroup = NodeGroup.AGENTS
     error_handling: ErrorHandling = ErrorHandling(timeout_seconds=600)
-    streaming: StreamingConfig = StreamingConfig()
+    streaming_mode: StreamingMode = Field(
```
delamainer: I would keep it as `streaming`. Essentially, this is a still node's steaming configuration, but with a different set of parameters.

olbychos: Already updated this!

delamainer: Do we use the same event name for the final output and intermediate steps?

TrachukT (approved):

dynamiq/types/streaming.py
```diff
@@ -65,12 +65,14 @@ class StreamingConfig(BaseModel):
         timeout (float | None): Timeout for streaming. Defaults to None.
         input_queue (Queue | None): Input queue for streaming. Defaults to None.
         input_queue_done_event (Event | None): Event to signal input queue completion. Defaults to None.
+        is_final (bool): Whether the streaming is final. Defaults to False.
```
acoola, line 68: Sorry, both the name of param and description are not clear.

olbychos: Option for stream all steps or only final version

acoola: Let's use mode with Enum as you used previously

examples/use_case_streaming_agents/react/backend.py
```diff
@@ -0,0 +1,57 @@
...
+    return agent
+
+
+def generate_agent_response(agent: ReActAgent, user_input: str):
+    """
+    Processes the user input using the agent. Supports both streaming and non-streaming responses.
+    """
+    if agent.streaming.enabled:
+        streaming_handler = StreamingIteratorCallbackHandler()
+        agent.run(
```
acoola, line 42: To unblock this operation, don't wait result of it and start streaming immediately you could run it in separate thhread. Im similar way we managed streaming in async:
https://github.com/dynamiq-ai/dynamiq/blob/main/examples/chainlit/chainlit_agent.py#L76
https://github.com/dynamiq-ai/dynamiq/blob/main/examples/chainlit/chainlit_agent.py#L26

PR #36, opened 2024-11-11
feat: graph agent and orchestrator

maksDev123:


olbychos (changes requested):
Please revert the adaptive orchestrator to its previous state and keep the graph orchestrator as a separate node.

dynamiq/nodes/agents/orchestrators/adaptive_manager.py
```diff
@@ -7,9 +7,10 @@
 2. Break down the task into subtasks
 3. Delegate subtasks to appropriate specialized agents
 4. Synthesize results from specialized agents
-5. Request clarification from the user when needed
+5. Request clarification from the user when needed (If appropriate tools are available)
```
olbychos, line 10: Don't assume that an LLM can recognize the "appropriate tool." It's better to rely on "human feedback" for sure.

olbychos (commented):
I left a brief comment. The testing for the trip planner example failed, but the other two were successful. @acoola, could you please review the graph as well?

dynamiq/nodes/agents/react.py
```diff
@@ -287,7 +287,7 @@ def tracing_intermediate(self, loop_num, formatted_prompt, llm_generated_output)
         ).model_dump()
 
     def convert_string_to_dict(string: str) -> dict:
-        string = re.sub(r'\\+', r'\\', string)
```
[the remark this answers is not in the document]
maksDev123: Yep, seems it is not used, removed it

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+    """Raised when next state was not found."""
+    pass
+
+
+class StateInputSchema(BaseModel):
+    context: dict[str, Any] = Field(..., description="Previous context objects.")
+    chat_history: list[dict[str, Any]] = Field(..., description="Previous chat history.")
+
+
+class State(Node):
```
acoola, line 30: Let's maybe move it to other utils nodes

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+    """Represents single state of graph flow
+
+    Attributes:
```
acoola, line 33: Please update dosctrings

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+    def init_components(self, connection_manager: ConnectionManager = ConnectionManager()) -> None:
+        """
+        Initialize components of the orchestrator.
+
+        Args:
+            connection_manager (ConnectionManager, optional): The connection manager. Defaults to ConnectionManager.
+        """
+        super().init_components(connection_manager)
+
+        for task in self.tasks:
```
acoola, line 99: It seems need to init manager, too if it exists

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+            if task.is_postponed_component_init:
+                task.init_components(connection_manager)
+
+    def _submit_task(self, task, global_context, chat_history, config, **kwargs) -> str:
```
acoola, line 103: Please add annotations and fix return annotation and docstings

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+        for d in context_list:
+            for key, value in d.items():
+                if key in merged_dict:
+                    if merged_dict[key] != value:
+                        raise OrchestratorError(f"Error: multiple changes of variable {key} are detected.")
+                merged_dict[key] = value
+
+        return merged_dict
+
+    def task_description(self, task: Agent):
```
acoola, line 87: Please add return annotation

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+        return ("Name:", task.name, "Role:", task.role)
```
acoola, line 88: It looks a little bit weird tuple. Please use better naming, as description in most cases mean some text/string

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+            str: The result of the task execution.
+
+        Raises:
+            OrchestratorError: If an error occurs during the execution process.
+
+        """
+
+        run_depends = []
+        if isinstance(task, Agent):
+            manager_result = self.manager.run(
```
acoola, line 121: `self.manager` is optional

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+        if not isinstance(context, dict):
+            raise OrchestratorError(
+                f"Error: Task returned invalid data format. Expected a dictionary got {type(context)}"
+            )
+
+        if "result" not in context:
+            raise OrchestratorError("Error: Task returned dictionary with no 'result' key in it.")
+
+        if "history" in context:
+            context.pop("history")
```
acoola, line 184: `context.pop("history", None)` could be used without  if statement

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+        if self.manager.is_postponed_component_init:
+            self.manager.init_components(connection_manager)
+
+        for state in self.states.values():
+            if state.is_postponed_component_init:
+                state.init_components(connection_manager)
+            state.manager = self.manager
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
acoola, line 294: `*args` in unexpected hete

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+                raise ValueError(f"State with name {state_name} is not present")
+
+    def add_conditional_edge(self, source: str, destinations: list[str], condition: Callable | Python) -> None:
+        """
+        Adds conditional edge to the graph.
+
+        Args:
+            source (str): Name of source state.
+            destinations (list[str]): Name of destinations states.
+            path_func (Callable): Function that will determine next state.
```
acoola, line 375: Please fix docsstring

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+                if next_state not in self.states:
+                    raise StateNotFoundError(f"State with name {next_state} was not found.")
+
+                return self.states[next_state]
+            else:
+                return self.get_next_state_by_manager(state, config)
+        else:
+            return self.states[state.next_states[0]]
+
+    def states_descriptions(self, states: list[State]) -> str:
```
acoola, line 479: It seems function works with `str`, but in annotation present `State`

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+                self.context = self.context | output["context"]
+                self._run_depends = [NodeDependency(node=state).to_dict()]
+                self._chat_history = output["chat_history"]
+
+            state = self._get_next_state(state, config=config, **kwargs)
+
+    def setup_streaming(self) -> None:
+        """Setups streaming for orchestrator."""
+        self.manager.streaming = self.streaming
+        for state in self.states:
```
acoola, line 528: Here you iterate over dict keys, not values. Please fix

dynamiq/nodes/agents/orchestrators/adaptive.py
```diff
@@ -55,25 +49,15 @@ class AdaptiveOrchestrator(Node):
     Attributes:
         manager (ManagerAgent): The managing agent responsible for overseeing the orchestration process.
         agents (List[BaseAgent]): List of specialized agents available for task execution.
-        objective (Optional[str]): The main objective of the orchestration.
+        input_tasks (Optional[str]): The main objective of the orchestration.
```
acoola, line 52: `input_tasks` missed, so please remove

dynamiq/nodes/agents/orchestrators/orchestrator.py
```diff
@@ -0,0 +1,143 @@
...
+    Attributes:
+        manager (ManagerAgent): The managing agent responsible for overseeing the orchestration process.
+        agents (List[BaseAgent]): List of specialized agents available for task execution.
+        objective (Optional[str]): The main objective of the orchestration.
+    """
+
+    name: str | None = "Orchestrator"
+    group: NodeGroup = NodeGroup.AGENTS
+    manager: AgentManager
+    _chat_history: list = []
```
acoola, line 36: It seems `_chat_history` obsolete here

dynamiq/nodes/agents/orchestrators/orchestrator.py
```diff
@@ -0,0 +1,143 @@
...
+    input: str = Field(default="", description="The main objective of the orchestration.")
+
+
+class Orchestrator(Node, ABC):
+    """
+    Orchestrates the execution of complex tasks using multiple specialized agents.
+
+    Attributes:
+        manager (ManagerAgent): The managing agent responsible for overseeing the orchestration process.
+        agents (List[BaseAgent]): List of specialized agents available for task execution.
```
acoola, line 29: Please clean up docstrings

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+                }
+            )
+
+            global_context = global_context | context
+
+        elif len(self.tasks) > 1:
+
+            contexts = []
+
+            with ThreadPoolExecutor() as executor:
```
acoola, line 221: Like this idea, but I see that currently, Agents don't support concurrent/parallel run of the same instance. 
We have a few instance attributes (self._intermediate_steps, self._run_depends, self._chat_history, etc) that operates with mutable objects and also reset during `execute` call. So few calls or `run`/`execute` break logic.

- Let's use consecutive calls for now.
- Please also revert tracing changes as well. We need to improve that not just for Agent purposes with minimal breaking changes

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,531 @@
...
+        agents (List[BaseAgent]): List of specialized agents available for task execution.
+        objective (Optional[str]): The main objective of the orchestration.
+        max_loops (Optional[int]): Maximum number of transition between states.
+    """
+
+    name: str | None = "GraphOrchestrator"
+    manager: GraphAgentManager
+    initial_state: str = START
+
+    states: dict[str, State] = {}
```
acoola, line 273: Lets use list of `states` similar to other usages: agent/state `tasks`,  flow `nodes`, workflow `flows` and so on.
You could create `_state_by_id` on init for internal usage (similar to Flow)

acoola:
Please rebase/merge main

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+
+    name: str | None = "GraphOrchestrator"
+    manager: GraphAgentManager
+    initial_state: str = START
+    context: dict[str, Any] = {}
+    states: list[State] = []
+    max_loops: int = 15
+
+    def init_components(self, connection_manager: ConnectionManager = ConnectionManager()) -> None:
```
acoola, line 49:
```suggestion
    def init_components(self, connection_manager: ConnectionManager = ConnectionManager | None = None) -> None:
        connection_manager = connection_manager or ConnectionManage()
```

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,316 @@
...
+        if self.manager.is_postponed_component_init:
+            self.manager.init_components(connection_manager)
+
+        for state in self.states:
+            if state.is_postponed_component_init:
+                state.init_components(connection_manager)
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.states.extend(
```
acoola, line 67: Make sense to check if `start` and `end` states exists before add them

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,316 @@
...
+
+        Returns:
+            dict: A dictionary representation of the instance.
+        """
+        data = super().to_dict(**kwargs)
+        data["manager"] = self.manager.to_dict(**kwargs)
+        data["states"] = [state.to_dict(**kwargs) for state in self.states]
+        return data
+
+    def add_node(self, name: str, tasks: list[Node | Callable]) -> None:
```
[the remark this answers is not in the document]
acoola: As we are operating with States, so maybe make sense to add `def add_state` with the same functionality. Or rename `states: list[State]` to `nodes: list[State]`?

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,319 @@
...
+                raise ValueError(f"State with name {state_name} is not present")
+
+    def add_conditional_edge(self, source: str, destinations: list[str], condition: Callable | Python) -> None:
+        """
+        Adds conditional edge to the graph.
+
+        Args:
+            source (str): Name of source state.
+            destinations (list[str]): Names of destination states.
+            path_func (Callable | Python): Condition that will determine next state.
```
acoola, line 160: Please update docstring

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,319 @@
...
+            names (list[str]): names of states to validate.
+
+        Raises:
+            ValueError: If state with specified name is not present.
+        """
+        for state_name in names:
+            if state_name not in self._state_by_id:
+                raise ValueError(f"State with name {state_name} is not present")
+
+    def add_conditional_edge(self, source: str, destinations: list[str], condition: Callable | Python) -> None:
```
acoola, line 153: Why here we have `destinations: list[str]`, and in `def add_edge` `destination: str`?

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -0,0 +1,319 @@
...
+
+        Returns:
+            dict: A dictionary representation of the instance.
+        """
+        data = super().to_dict(**kwargs)
+        data["manager"] = self.manager.to_dict(**kwargs)
+        data["states"] = [state.to_dict(**kwargs) for state in self.states]
+        return data
+
+    def add_state(self, state_id: str, tasks: list[Node | Callable]) -> None:
```
acoola, line 94: Let's rename it little bit to smth like `add_state_by_tasks`.
`add_state` is a general and for me, it should looks like `def add_state(self, state: State)`. In such case I could define own name, description and so on.

dynamiq/nodes/agents/orchestrators/graph.py
maksDev123: `add_edge`  accepts source and destination to create edge. This is used in case we want to go to destination right after source is finished. In `add_conditional_edge` providing destinations is not obligatory for functionality, but critical to ability to draw this graph. In case of conditional edge next state is determined by passed function or Python Node.

dynamiq/nodes/agents/orchestrators/graph.py
maksDev123: Agree, but it seems that it gets a bit wordy and less user-friendly. I do not see any need to add separate function which accepts State objects. Also it is implemented in the same way in LangGraph.

acoola: It's ok if we add a little bit more comparing to langraph. How I could add mine `State` to graph as I'm able to init State by my self? Please add this functionality

dynamiq/nodes/agents/orchestrators/graph.py
acoola: I mean, why does conditional edge support multiple destinations and edge just one?
If there is some specific - please describe it in docstring.

dynamiq/nodes/agents/orchestrators/graph.py
acoola: We have `condition`, not `path_func`

acoola (approved):

acoola:
Great tremendous work

dynamiq/nodes/agents/orchestrators/orchestrator.py
```diff
@@ -0,0 +1,141 @@
...
+            dict[str, Any]: The result of the orchestration process.
+        """
+
+        self.reset_run_state()
+        config = ensure_config(config)
+        self.run_on_node_execute_run(config.callbacks, **kwargs)
+
+        input_task = input_data.input or self.objective
+
+        logger.debug(f"LinearOrchestrator {self.id}: starting the flow with input_task:\n```{input_task}```")
```
olbychos, line 126: Just fix please naming
`LinearOrchestrator` -> `Orchestrator`

olbychos (approved):
Let's make this stuff alive on UI.
Left small comment about logs naming

PR #48, opened 2024-11-18
feat: allow relative import for python node

olbychos:


TrachukT (approved):

examples/tools/use_python_node.py
```diff
@@ -0,0 +1,137 @@
+from pydantic import ConfigDict
```
acoola, line 1: Please extend python node tests with the similar examples: https://github.com/dynamiq-ai/dynamiq/blob/main/tests/integration/nodes/tools/test_python.py

acoola (approved):

PR #57, opened 2024-11-19
fix: union types handling for agent stuff

olbychos:


dynamiq/nodes/agents/react.py
```diff
@@ -587,24 +595,33 @@ def generate_structured_output_schemas(self):
...
+                case "int":
+                    return "integer"
+                case "float":
+                    return "float"
+                case _:
+                    return "string"
+        elif hasattr(param_type, "__origin__") and param_type.__origin__ is Union:
+            # Handle Union types by using the first type in the Union
+            first_type = param_type.__args__[0]
+            return "string" if not hasattr(first_type, "__name__") else first_type.__name__
```
maksDev123, line 613: It seems that in case this condition is false value for type is not validated.

dynamiq/nodes/agents/react.py
```diff
@@ -587,24 +595,33 @@ def generate_structured_output_schemas(self):
...
-            case "bool":
-                return "boolean"
-            case "int":
-                return "integer"
-            case "float":
-                return "float"
-        return "string"
+    def filter_format_type(param_type: str | type) -> str:
+        """Filters proper type for a function calling schema."""
+        if isinstance(param_type, str):
```
maksDev123, line 600: Also it seems it will never get here, as you are passing class type not str.

dynamiq/nodes/agents/react.py
olbychos: Great, I've fixed this.

dynamiq/nodes/agents/react.py
maksDev123: Maybe just change conditions in order.

olbychos: Should be updated

maksDev123: It seems that it still have the same behaviour. Or I miss something.

olbychos: Here are tests results:
```
Testing with string inputs:
Input: 'bool', Output: boolean, Expected: 'boolean'
Input: 'int', Output: integer, Expected: 'integer'
Input: 'float', Output: float, Expected: 'float'
Input: 'some_random_string', Output: string, Expected: 'string'

Testing with Python types:
Input: int, Output: int, Expected: 'int'
Input: str, Output: str, Expected: 'str'
Input: float, Output: float, Expected: 'float'
Input: bool, Output: bool, Expected: 'bool'
Input: list, Output: list, Expected: 'list'
Input: dict, Output: dict, Expected: 'dict'

Testing with Union inputs:
Input: Union[int, str], Output: int, Expected: 'int'
Input: Union[None, list], Output: list, Expected: 'list'
Input: Union[float, None], Output: float, Expected: 'float'

Testing with complex Union inputs:
Input: Union[None, Union[int, str]], Output: int, Expected: 'int'
Input: Union[Union[str, int], list], Output: str, Expected: 'str'

Testing with fallback cases:
Input: CustomType, Output: CustomType, Expected: 'CustomType'
Input: object, Output: object, Expected: 'object'

Testing with edge cases:
Input: None, Output: string, Expected: 'string'
Input: 123, Output: string, Expected: 'string'
```

maksDev123: Yes, but here you pass self.filter_format_type(field.annotation) not string.

maksDev123: Yes but expected for (Input: int, Output: int, Expected: 'int') Expected should be 'integer' not int.

maksDev123: Also it seems we never pass string to a function  (you pass filter_format_type(field.annotation))

olbychos:
```
type_mapping = {
    int: "integer",
    float: "float",
    bool: "boolean",
    str: "string",
    list: "list",
    dict: "dict",
}
```
This mapping can help address naming issues, so please double-check it. Regarding strings input, it would be good to keep this behavior.

maksDev123: `type_mapping = {
    int: "integer",
    float: "number",
    bool: "boolean",
    str: "string",
}
`
Regarding list and dict, it has to map to `array` and `object`, but we also need to define special structured for them. So their behaviour would be different from first 4. I guess it is fine now to have only first 4 for now.

dynamiq/nodes/agents/react.py
```diff
@@ -543,13 +543,21 @@ def _extract_final_answer_xml(self, llm_output: str) -> str:
...
-                for name, field in tool.input_schema.model_fields.items()
-                if not field.json_schema_extra or field.json_schema_extra.get("is_accessible_to_agent", True)
-            ]
+            params = []
+            for name, field in tool.input_schema.model_fields.items():
+                if not field.json_schema_extra or field.json_schema_extra.get("is_accessible_to_agent", True):
+                    # Handle Union types
+                    if hasattr(field.annotation, "__origin__") and field.annotation.__origin__ is Union:
+                        type_names = [t.__name__ for t in field.annotation.__args__ if hasattr(t, "__name__")]
+                        type_str = " | ".join(type_names)
```
acoola, line 555: This is a little bit clear. Plase check if works fine for you.
Don't forget to make proper imports:
```
import types
from typing import get_origin
```
```suggestion
                    if get_origin(field.annotation) in (Union, types.UnionType):
                        type_str = str(field.annotation)
```

acoola (approved):

PR #59, opened 2024-11-20
feat: add deepseek llm, fix perplexity connection

olbychos:


acoola (approved):

tests/integration/nodes/llms/test_deepseek.py
```diff
@@ -0,0 +1,89 @@
...
+                ),
+            ],
+        ),
+    )
+
+    return wf
+
+
+@pytest.mark.parametrize(
+    ("model", "expected_model"),
```
vitalii-dynamiq, line 43: changes to .env are missing

PR #69, opened 2024-11-27
feat: add `exa` tool

olbychos:


dynamiq/nodes/tools/exa_search.py
```diff
@@ -0,0 +1,169 @@
...
+    """Schema for Exa search input parameters."""
+
+    query: str = Field(description="The search query string.")
+    get_contents: bool = Field(
+        default=False, description="If true, retrieve full content, highlights, and summaries for search results."
+    )
+    use_autoprompt: bool = Field(
+        default=False, description="If true, query will be converted to a Exa query. Default false."
+    )
+    type: Literal["keyword", "neural", "auto"] = Field(
```
acoola, line 22: Please use enum

dynamiq/nodes/tools/exa_search.py
```diff
@@ -0,0 +1,169 @@
...
+        default="neural", description="The Type of search: 'keyword', 'neural', or 'auto'."
+    )
+    category: str | None = Field(
+        default=None, description="A data category to focus on (e.g., company, research paper, news article)."
+    )
+    num_results: int = Field(default=10, ge=1, le=100, description="Number of search results to return. Default 10.")
```
acoola, line 28: In analog we are using `limit`

dynamiq/nodes/tools/exa_search.py
```diff
@@ -0,0 +1,169 @@
...
+from dynamiq.nodes.node import ConnectionNode, ensure_config
+from dynamiq.runnables import RunnableConfig
+from dynamiq.utils.logger import logger
+
+
+class ExaInputSchema(BaseModel):
+    """Schema for Exa search input parameters."""
+
+    query: str = Field(description="The search query string.")
+    get_contents: bool = Field(
```
acoola, line 16: Let's use smth like `include_full_content` to be consistent

dynamiq/nodes/tools/exa_search.py
```diff
@@ -0,0 +1,169 @@
...
+        payload = {
+            "query": payload.pop("query"),
+            "useAutoprompt": payload.pop("use_autoprompt", False),
+            "type": payload.pop("type", "neural"),
+            "numResults": payload.pop("num_results", 10),
+            **{self.to_camel_case(k): v for k, v in payload.items()},
+        }
+
+        if get_contents:
+            payload["contents"] = {
```
acoola, line 126: Let's define extra models for such `contents` data and defining it with such defaults in `ExaInputSchema`. So the user could update such params in case it would like.

olbychos: I've simplified it, due to agents stuff

acoola: Ok, let's do now not just for Agents stuff.

olbychos: If I create a new schema for content, we won't be able to generate and use these schemas with the agents, as this could cause problems with the schema tool's description generation. 
For user behavior, it would be sufficient to add a choice: either receive additional content or not.

dynamiq/connections/connections.py
```diff
@@ -929,3 +930,32 @@ class DeepSeek(BaseApiKeyConnection):
...
+        api_key (str): The API key for authentication, fetched from the environment variable 'EXA_API_KEY'.
+        method (Literal[HTTPMethod.POST]): HTTP method used for the request, defaults to HTTPMethod.POST.
+        headers (dict[str, Any]): HTTP headers for the request.
+    """
+
+    type: Literal[ConnectionType.Exa] = ConnectionType.Exa
+    url: str = Field(default="https://api.exa.ai/search")
+    api_key: str = Field(default_factory=partial(get_env_var, "EXA_API_KEY"))
+    method: Literal[HTTPMethod.POST] = HTTPMethod.POST
+    headers: dict[str, Any] = Field(default_factory=dict)
```
delamainer, line 951: Why do we need `url`, `method` and `headers` here?

delamainer: `URL`, `method`, and `headers` are always the same.

delamainer (changes requested):

dynamiq/connections/connections.py
olbychos: Structure is the same as for our other tools like `Tavily`, `Zenrows` , etc

delamainer: There is no point in following the same structure and bringing some unused fields to connection.

delamainer: I haven't checked the `Tavily`, `Zenrows` API docs, but I think we might do that incorrectly because, usually, it's how APIs are defined.

delamainer (changes requested):
Update the connections.

dynamiq/connections/connections.py
olbychos: I've removed `headers` but kept `url`/`method`, you can use either `url` for `scrape` and so on

delamainer: We cannot use different URLs. The node is working with the exact URL and exact method.

dynamiq/connections/connections.py
```diff
@@ -929,3 +930,30 @@ class DeepSeek(BaseApiKeyConnection):
...
+        url (str): The URL of the Exa API.
+        api_key (str): The API key for authentication, fetched from the environment variable 'EXA_API_KEY'.
+        method (Literal[HTTPMethod.POST]): HTTP method used for the request, defaults to HTTPMethod.POST.
+    """
+
+    type: Literal[ConnectionType.Exa] = ConnectionType.Exa
+    url: str = Field(default="https://api.exa.ai/search")
+    api_key: str = Field(default_factory=partial(get_env_var, "EXA_API_KEY"))
+    method: Literal[HTTPMethod.POST] = HTTPMethod.POST
+
```
delamainer, line 950:
```python3
class Exa(BaseApiKeyConnection):
    type: Literal[ConnectionType. Exa] = ConnectionType. Exa
    api_key: str = Field(default_factory=partial(get_env_var, "EXA_API_KEY"))
...
```

This is all we need.

delamainer (changes requested):

dynamiq/connections/connections.py
olbychos: This would be inconsistent with the behavior of other tools. WDYT, @acoola?

delamainer: Other tools and their connections are poorly implemented. For example `TavilyTool` is [designed](https://github.com/dynamiq-ai/dynamiq/blob/b572490638fb200659a1dc7c1254c1fcaa8a9504/dynamiq/nodes/tools/tavily.py#L14) to work with an [exact endpoint](https://docs.tavily.com/docs/rest-api/api-reference#endpoint-post-search) that has an exact method + exact payload + exact response. Why would you want to overwrite the method or headers?

delamainer: The same is true for other tools. Overriding a `method` or an `URL` or `headers` only can break things.

delamainer: Here is the [endpoint](https://github.com/dynamiq-ai/dynamiq/blob/b572490638fb200659a1dc7c1254c1fcaa8a9504/dynamiq/nodes/tools/tavily.py#L126) we want to use. And this endpoint expects `POST`, not anything else.

olbychos: Regarding Tavily, if we want to add a new node for extraction or make changes to our existing node, we need to update the URL in the connection and use the same connection as we do with Cohere for LLMs and embedders.

https://docs.tavily.com/docs/rest-api/api-reference#endpoint-post-extract

delamainer: We will not have to update the URL (aka base URL), we will have to update the path only.

delamainer: And the path is defined on the node level, not in the connection.

delamainer: For example, the connection has two fields, `url` + `api_key` (this is more than enough), and even `api_key` should be enough.

`TavilySearchTool` will use `/search` path.
`TavilyExtractTool` will use `/extract` path.

olbychos: Our base HTTP API or HTTP connections depend on the `url` parameter.

delamainer: This is why we need to use `BaseApiKeyConnection` or `HttpApiKey`.

dynamiq/nodes/tools/exa_search.py
```diff
@@ -0,0 +1,177 @@
...
+    limit: int = Field(default=10, ge=1, le=100, description="Number of search results to return. Default 10.")
+    include_domains: list[str] | None = Field(default=None, description="List of domains to include in the search.")
+    exclude_domains: list[str] | None = Field(default=None, description="List of domains to exclude from the search.")
+    include_text: list[str] | None = Field(default=None, description="Strings that must be present in webpage text.")
+    exclude_text: list[str] | None = Field(
+        default=None, description="Strings that must not be present in webpage text."
+    )
+
+
+class ExaTool(ConnectionNode):
```
vitalii-dynamiq, line 45: why we need Tool in the name?

dynamiq/connections/connections.py
acoola: Please inherit from `Http` instead of `HttpApiKey`

acoola: if url unchanged please use Literal instead of str

dynamiq/connections/connections.py
acoola: We need to inherit from `Http` and define proper attributes that need for connection in connection

acoola (changes requested):
Please revert hardcode from def execute and move attributes that need for proper connection to service to Connection. Please use inheritance from Http in the same way as used for other similar tools with Literals where this need

dynamiq/nodes/tools/exa_search.py
olbychos: Can use ‘ExaSearch’ instead of this

acoola: Historically we are using Tool suffix for such nodes

dynamiq/connections/connections.py
```diff
@@ -929,3 +930,30 @@ class DeepSeek(BaseApiKeyConnection):
...
+    Attributes:
+        type (Literal[ConnectionType.Exa]): The type of connection, which is always 'Exa'.
+        url (str): The URL of the Exa API.
+        method (Literal[HTTPMethod.POST]): HTTP method used for the request, defaults to POST.
+        api_key (str): The API key for authentication, fetched from the environment variable 'EXA_API_KEY'.
+    """
+
+    type: Literal[ConnectionType.Exa] = ConnectionType.Exa
+    url: str = Field(default="https://api.exa.ai")
```
acoola, line 947:
```suggestion
    url: Literal["https://api.exa.ai"] = Field(default="https://api.exa.ai")
```

acoola (approved):
Great job

PR #70, opened 2024-11-27
refactor: change vector store configs

TrachukT:

PR #74, opened 2024-11-28
feat: integrate PGVector vector database

tyaroshko:


olbychos (commented):
Tests should be reviewed, and any conflicts should be resolved.

MaksymAI (approved):

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
```
acoola, line 1: Please use Poetry 1.8.3, or revert this one line change

dynamiq/components/retrievers/pgvector.py
```diff
@@ -0,0 +1,74 @@
...
+        self.vector_store = vector_store
+        self.filters = filters or {}
+        self.top_k = top_k
+
+    def run(
+        self,
+        query_embedding: list[float],
+        exclude_document_embeddings: bool = True,
+        top_k: int | None = None,
+        filters: dict[str, Any] | None = None,
```
acoola, line 46: Please check improvements from this one PR: https://github.com/dynamiq-ai/dynamiq/pull/70.
Need to add support of
```
content_key: str | None = None,
embedding_key: str | None = None,
```

dynamiq/connections/connections.py
```diff
@@ -932,6 +933,58 @@ def connect(self):
...
+        """
+        return {
+            "host": self.host,
+            "port": self.port,
+            "database": self.database,
+            "user": self.user,
+            "password": self.password,
+        }
+
+    def __str__(self):
```
acoola, line 984: If you added this for proper logging, please hide sensitive info (password)

dynamiq/nodes/writers/pgvector.py
```diff
@@ -0,0 +1,79 @@
...
+            config (RunnableConfig, optional): Configuration for the execution.
+            **kwargs: Additional keyword arguments.
+
+        Returns:
+            dict: A dictionary containing the count of upserted documents.
+        """
+        config = ensure_config(config)
+        self.run_on_node_execute_run(config.callbacks, **kwargs)
+
+        documents = input_data["documents"]
```
acoola, line 72: Please add support of `content_key` and `embedding_key` similar to Milvus (check parent classes too):
https://github.com/dynamiq-ai/dynamiq/blob/c547e44a804fccd94d01954be4060115c97c0b82/dynamiq/nodes/writers/milvus.py#L75

dynamiq/storages/vector/pgvector/pgvector.py
```diff
@@ -0,0 +1,551 @@
...
+        query = SQL(
+            """
+            CREATE TABLE IF NOT EXISTS {schema_name}.{table_name} (
+                id VARCHAR(128) PRIMARY KEY,
+                content TEXT,
+                metadata JSONB,
+                embedding vector({dimension})
+            );
+            """
+        ).format(
```
acoola, line 177: Please use native pcycopg params binding instead of create full str query with params by self. This helps to avoid SQL injections

dynamiq/storages/vector/pgvector/pgvector.py
```diff
@@ -0,0 +1,551 @@
...
+
+class PGVectorStore:
+    """Vector store using pgvector."""
+
+    def __init__(
+        self,
+        connection: PGVector | str | None = None,
+        client: Optional["PGVector"] = None,
+        create_extension: bool = True,
+        table_name: str = "dynamiq_vector_store",
```
acoola, line 69: I think `table` and `schema` names should be configurable in Nodes with default values (writer and retriever). So users could specify their own. Also make sense to be able to handle this in runtime similar to `content_key` and `embedding_key`.

tyaroshko: Should the usage for `table` and `schema` be the same as for `documents` in writer or `query_embedding`, `filters`, and `top_k` in retriever? Meaning if I should be able to extract the `table` and `schema` keys from the `input_data` dict and pass them to the vector store?

vitalii-dynamiq: table and schema should be configurable in the UI in the same way index is configured in Pinecone

dynamiq/connections/connections.py
```diff
@@ -932,6 +933,58 @@ def connect(self):
         pass
 
 
+class PGVector(BaseConnection):
```
delamainer, line 936: Should this be `PostgreSQL`?

delamainer: `pgvector` is just an extension for `PostgreSQL`

dynamiq/connections/connections.py
```diff
@@ -932,6 +933,58 @@ def connect(self):
...
+    type: Literal[ConnectionType.PGVector] = ConnectionType.PGVector
+    host: str = Field(default_factory=partial(get_env_var, "PGVECTOR_HOST", "localhost"))
+    port: str | int = Field(default_factory=partial(get_env_var, "PGVECTOR_PORT", "5432"))
+    database: str = Field(default_factory=partial(get_env_var, "PGVECTOR_DATABASE", "db"))
+    user: str = Field(default_factory=partial(get_env_var, "PGVECTOR_USER", "postgres"))
+    password: str = Field(default_factory=partial(get_env_var, "PGVECTOR_PASSWORD", "password"))
```
delamainer, line 942: What about `SSLMODE`?

dynamiq/connections/connections.py
tyaroshko: I guess we can rename it to `PostgreSQL` if it makes sense.

dynamiq/nodes/writers/pgvector.py
```diff
@@ -0,0 +1,83 @@
...
+
+from dynamiq.connections import PGVector
+from dynamiq.nodes.node import NodeGroup, VectorStoreNode, ensure_config
+from dynamiq.runnables import RunnableConfig
+from dynamiq.storages.vector.base import BaseWriterVectorStoreParams
+from dynamiq.storages.vector.pgvector.pgvector import PGVectorStore
+from dynamiq.utils.logger import logger
+
+
+class PGVectorDocumentWriter(VectorStoreNode, BaseWriterVectorStoreParams):
```
acoola, line 11: There need to add extra base class and use instead of BaseWriterVectorStoreParams:
https://github.com/dynamiq-ai/dynamiq/blob/9134b277128f2f59b2b9884ae726a787e8afaa1e/dynamiq/storages/vector/milvus/milvus.py#L16-L17

This automatically add `content_key` and `embedding_key` to node attributes. Currently `embedding_key` missed

dynamiq/connections/connections.py
tyaroshko: I think it uses `prefer` mode for `SSLMODE` by default (meaning `first try an SSL connection; if that fails, try a non-SSL connection`), but, if needed, I can add it to env vars and the connection string.

tyaroshko: Also, are there any more params we should add to the connection (I'm using [this documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS) for reference)?

dynamiq/storages/vector/pgvector/filters.py
```diff
@@ -0,0 +1,398 @@
...
+    Raises:
+        VectorStoreFilterException: If the value is of an unsupported type.
+    """
+    if isinstance(value, str):
+        try:
+            datetime.fromisoformat(value)
+        except (ValueError, TypeError) as exc:
+            msg = "Can't compare strings using operators '>', '>=', '<', '<='. "
+            raise VectorStoreFilterException(msg) from exc
+    if type(value) in [list, Jsonb]:
```
acoola, line 224: Is it possible to use `isinstance(value, (list, Jsonb))` too?

tyaroshko: Sure, fixed that.

dynamiq/nodes/writers/pgvector.py
tyaroshko: Thanks for the clarification, added the `PGVectorStoreParams` in the latest commit.

dynamiq/connections/connections.py
delamainer: I think we already have the bare minimum.

dynamiq/connections/connections.py
delamainer: Let's rename it.

dynamiq/connections/connections.py
```diff
@@ -932,6 +933,44 @@ def connect(self):
         pass
 
 
+class PGVector(BaseConnection):
+    type: Literal[ConnectionType.PGVector] = ConnectionType.PGVector
+    host: str = Field(default_factory=partial(get_env_var, "PGVECTOR_HOST", "localhost"))
```
delamainer, line 938: Why `port` is a string?

delamainer: It should be int (as in the rest of the connections).

dynamiq/connections/connections.py
tyaroshko: Is it fine to leave everything else as `PGVector` and only rename the connection to `PostgreSQL`?

delamainer: Yeah, only the connection. The rest (nodes) are fine.

dynamiq/connections/connections.py
delamainer: Is there a need for `str | int`?

dynamiq/storages/vector/pgvector/pgvector.py
```diff
@@ -0,0 +1,648 @@
...
+            create_if_not_exist (bool): Whether to create the table and index if they do not exist. Defaults to True.
+            content_key (Optional[str]): The field used to store content in the storage. Defaults to 'content'.
+            embedding_key (Optional[str]): The field used to store embeddings in the storage. Defaults to 'embedding'.
+        """
+        if vector_function not in PGVectorVectorFunction:
+            raise ValueError(f"vector_function must be one of {PGVectorVectorFunction.list()}")
+        if index_method is not None and index_method not in PGVectorIndexMethod:
+            raise ValueError(f"index_method must be one of {PGVectorIndexMethod.list()}")
+
+        if isinstance(connection, str):
```
acoola, line 111: It seems we should connect this just in case `if client is None`. client in you case is already created connection.

dynamiq/storages/vector/pgvector/pgvector.py
```diff
@@ -0,0 +1,648 @@
...
+        self.create_extension = create_extension
+        if self.create_extension:
+            self._conn.execute("CREATE EXTENSION IF NOT EXISTS vector")
+            self._conn.commit()
+
+        register_vector(self._conn)
+
+        self.client = client if client else self._conn
+
+        self.table_name = table_name if table_name else DEFAULT_TABLE_NAME
```
acoola, line 129:
```suggestion
        self.table_name = table_name or DEFAULT_TABLE_NAME
```
or better in _init_ use `table_name: str = DEFAULT_TABLE_NAME `.

The same for schema_name

dynamiq/connections/connections.py
tyaroshko: No, I don't think that there is a need for both types to be supported. I'll remove the `str` for `port`.

dynamiq/storages/vector/pgvector/pgvector.py
```diff
@@ -0,0 +1,648 @@
...
+
+        if (
+            self.index_method == PGVectorIndexMethod.IVFFLAT
+            and self.vector_function == PGVectorVectorFunction.L1_DISTANCE
+        ):
+            msg = "IVFFLAT index does not support L1 distance metric"
+            raise VectorStoreException(msg)
+
+        if create_if_not_exist:
+            with self._get_connection() as conn:
```
acoola, line 146: Why do we again try to connect and don't reuse self.client or self._conn?

dynamiq/connections/connections.py
tyaroshko: Thanks, will do.

dynamiq/storages/vector/pgvector/pgvector.py
tyaroshko: `_get_connection()` will just yield `self._conn` if it is already established and only tries to connect when the connection is closed or absent (as far as I understand it).

dynamiq/storages/vector/pgvector/pgvector.py
tyaroshko: Got it.

dynamiq/storages/vector/pgvector/pgvector.py
```diff
@@ -0,0 +1,648 @@
...
+
+        with self._get_connection() as conn:
+            with conn.cursor(row_factory=dict_row) as cur:
+                result = self._execute_sql_query(sql_query, params, cursor=cur)
+                records = result.fetchall()
+
+                documents = self._convert_query_result_to_documents(records)
+                return documents
+
+    def __del__(self):
```
acoola, line 645: Please add a separate `def close` method for this purposes. We are trying to close all connection in ConnectionManeger by calling this method if it present:
https://github.com/dynamiq-ai/dynamiq/blob/9134b277128f2f59b2b9884ae726a787e8afaa1e/dynamiq/connections/managers.py#L134-L142

dynamiq/nodes/writers/pgvector.py
acoola: Please add `table_name` and `schema_name` as well, as this extra attributes that make sense for each of PGVector node.

poetry.lock
tyaroshko: Done

dynamiq/connections/connections.py
tyaroshko: Removed the redundant `__str__` method and added the `conn_params` property instead, and removed sensitive info (password) from logging.

dynamiq/nodes/retrievers/pgvector.py
```diff
@@ -0,0 +1,102 @@
...
+            config (RunnableConfig, optional): The configuration for the execution.
+            **kwargs: Additional keyword arguments.
+
+        Returns:
+            dict[str, Any]: A dictionary containing the retrieved documents.
+        """
+        config = ensure_config(config)
+        self.run_on_node_execute_run(config.callbacks, **kwargs)
+
+        query_embedding = input_data["embedding"]
```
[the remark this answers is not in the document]
tyaroshko: Done

dynamiq/storages/vector/pgvector/pgvector.py
tyaroshko: Does it mean that the `__del__()` method is redundant here and I don't even have to add a separate `close()` method to close the connection as it will be closed automatically by the `ConnectionManager`?

dynamiq/storages/vector/pgvector/pgvector.py
tyaroshko: Done.

dynamiq/storages/vector/pgvector/pgvector.py
```diff
@@ -0,0 +1,648 @@
...
+
+from dynamiq.connections import PGVector
+from dynamiq.storages.vector.exceptions import VectorStoreException
+from dynamiq.storages.vector.pgvector.filters import _convert_filters_to_query
+from dynamiq.types import Document
+from dynamiq.utils.logger import logger
+from dynamiq.storages.vector.base import BaseWriterVectorStoreParams
+
+
+class PGVectorEnum(str, Enum):
```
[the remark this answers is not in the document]
tyaroshko: Thanks, fixed this issue.

dynamiq/nodes/writers/pgvector.py
tyaroshko: Finished with the `PGVectorStoreParams` class, and updated it with relevant fields for the nodes.

dynamiq/nodes/writers/pgvector.py
tyaroshko: Added the relevant parameters to base class for `PGVectorStoreParams`.

dynamiq/connections/connections.py
tyaroshko: Renamed the PGVector connection instances to PostgreSQL.

dynamiq/connections/connections.py
tyaroshko: @delamainer Should I still add the `SSLMODE` parameter to the connection? It would mean that more environment variables will be needed to provide local paths to relevant SSL certificates. Also, it would be harder to construct the connection string as its form will depend on which option was selected for the `SSLMODE` param (basically, a couple of if-statements should be added in this case).

delamainer: Let's omit it.

delamainer: Note: update the environment variable names too.

tyaroshko: Meaning `PGVECTOR_HOST` ->`POSTGRESQL_HOST` and so on?

tyaroshko: Renamed the environment variables for the `PGVector` connection to `POSTGRES_...` in the `.env.example`, `connections.py`, and RAG example for `pgvector`.

dynamiq/storages/vector/pgvector/pgvector.py
tyaroshko: Added the improved handling of connections for `PGVectorStore` aligned with your logic. Thanks, @acoola!

dynamiq/storages/vector/pgvector/pgvector.py
```diff
@@ -0,0 +1,687 @@
...
+}
+
+DEFAULT_TABLE_NAME = "dynamiq_vector_store"
+DEFAULT_SCHEMA_NAME = "public"
+
+
+class PGVectorStoreParams(BaseWriterVectorStoreParams):
+    table_name: str = DEFAULT_TABLE_NAME
+    schema_name: str = DEFAULT_SCHEMA_NAME
+    create_if_not_exist: bool = True
```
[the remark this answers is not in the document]
tyaroshko: @acoola should I have separate classes for reader and writer params? The first one (for the reader) called `PGVectorStoreParams` will be inherited from `BaseVectorStoreParams` and the other one (for the writer) will be inherited from `PGVectorStoreParams` and `BaseWriterVectorStoreParams`. Or is there another way to do it?

dynamiq/storages/vector/pgvector/pgvector.py
acoola: let's add like this:
```
    def close(self):
        if hasattr(self, "_conn") and self._conn is not None and not self._conn.closed:
            self._conn.close()

    def __del__(self):
        """Close the connection when the object is deleted."""
        self.close()
```

dynamiq/storages/vector/pgvector/pgvector.py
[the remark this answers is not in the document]
delamainer: It's not related to this.

But I think there are a couple of problems with inheritance. One more such example: `BaseVectorStoreParams` has `index_name` field, which is not needed for `pgvector` but all writers and retrievers inherit this class (including `pgvector`). There are several such examples in our package regarding the poor inheritance.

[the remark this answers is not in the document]
tyaroshko: Done

dynamiq/storages/vector/pgvector/pgvector.py
tyaroshko: Thanks, it's done now.

dynamiq/storages/vector/pgvector/pgvector.py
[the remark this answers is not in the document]
tyaroshko: @delamainer I mean, the index name is indeed applicable to `pgvector`, however, it does not have the same meaning as in other writers/retrievers (the alternative in `pgvector` is to set `table` or `schema` attributes, but we do also have the ability to change the name of the index created on a specific table). But I do agree that the inheritance here is far from ideal.

acoola (approved):
Great job

PR #75, opened 2024-12-01
fix: update input schemas for nodes

maksDev123:


dynamiq/nodes/llms/base.py
```diff
@@ -305,4 +330,6 @@ def execute(
             self._handle_streaming_completion_response if self.streaming.enabled else self._handle_completion_response
         )
 
-        return handle_completion(response=response, messages=messages, config=config, input_data=input_data, **kwargs)
+        return handle_completion(
```
acoola, line 333: Why is this shown as diff?

dynamiq/nodes/node.py
```diff
@@ -339,7 +339,9 @@ def validate_input_schema(self, input_data: dict[str, Any]) -> dict[str, Any] |
 
         if self.input_schema:
             try:
-                return self.input_schema(**input_data)
+                return self.input_schema.model_validate(
+                    input_data, context=kwargs | self.get_context_for_input_schema()
+                )
             except Exception as e:
                 raise RecoverableAgentException(f"Input data validation failed: {e}")
```
acoola, line 346: Here we raise `RecoverableAgentException`, which may make sense for tools, but just that used in Agents. For other cases, we should raise default error

acoola (approved):

acoola:
Great job

PR #82, opened 2024-12-06
feat: add json parser for llm evaluator

MaksymAI:


dynamiq/components/evaluators/llm_evaluator.py
```diff
@@ -333,7 +407,7 @@ def is_valid_json_and_has_expected_keys(
             bool: True if valid, False otherwise.
         """
         try:
-            parsed_output = json.loads(received)
+            parsed_output = self.parse_llm_json_output(received)
```
acoola, line 410: You are just trying to extract correct json on validation, but still return raw LLM output (potentially with some trash)

MaksymAI: Sorry, could you please explain what do you mean one more time? This method returns the loaded json

dynamiq/utils/json_parser.py
```diff
@@ -0,0 +1,86 @@
...
+    Args:
+        s (str): The input string containing the JSON object.
+
+    Returns:
+        Optional[str]: The extracted JSON string, or None if not found.
+    """
+    nesting = 0
+    start = None
+    for i, char in enumerate(s):
+        if char == "{":
```
acoola, line 18: It seems here you handle just object `"{}"` and don't handle list `"[]"`, `"null"` and few other types (int, float, etc)

MaksymAI: Could you explain that one more time, please?

This function extracts everything that starts with '{' until it finds the closing '}'. So, for instance, if the response looks like this:

```
Here is a JSON
{ "questions": ["Question 1", "Question 2"], 
  "answer": None,
  "contexts": [{"text": "text"}]
}
```

This function will extract:

```
{ "questions": ["Question 1", "Question 2"], 
  "answer": None,
  "contexts": [{"text": "text"}]
}
```

Then, the following function will clean it and attempt to load it as JSON.

Or do you mean we need to extract JSON that is a list of dictionaries as well? Kind of like starting from a list?
```
[{"question": "text"}, {"question": "text 2"}]
```
image

MaksymAI: image

acoola (approved):
Great job

PR #87, opened 2024-12-10
fix: add default value for date in ranker

maksDev123:


dynamiq/nodes/rankers/recency.py
```diff
@@ -73,6 +73,7 @@ class TimeWeightedDocumentRanker(Node):
     max_days: int = 3600
     min_coefficient: float = 0.9
     date_field: str = "date"
+    default_date: str = "01 January, 2022"
```
vitalii-dynamiq, line 76: why do we hardcore old date?

dynamiq/nodes/rankers/recency.py
```diff
@@ -110,6 +111,7 @@ def execute(
         ranked_documents = self.adjust_similarity_scores(
             documents,
             date_field=self.date_field,
+            default_date=self.default_date,
```
MaksymAI, line 114: The default date will impact the ranking. For example, if we set the default date to the current date, all documents without a date would be considered more recent than those with a specific date, which would create hidden logic. On the other hand, if we set the default date to a very old date, documents without a date would be considered older than others.

I suggest the following logic: if a document doesn't have a date, do not update its score. If a document has a date, update the score based on its recency. What do you think?

dynamiq/nodes/rankers/recency.py
maksDev123: Changed logic to update score only in case date is provided in metadata.

acoola (approved):

PR #91, opened 2024-12-13
feat: add node specific callbacks

maksDev123:
> [!IMPORTANT]
> Introduce node-specific callbacks to enhance event handling in nodes, updating `NodeCallbackHandler`, `GraphOrchestrator`, and `Node` class, with new tests and examples.
> 
>   - **Node Callbacks**:
>     - Introduce `NodeCallbackHandler` in `base.py` for node-specific events.
>     - Add node-specific callback methods like `on_node_start`, `on_node_end`, etc.
>   - **Graph Orchestrator**:
>     - Update `GraphOrchestrator` in `graph.py` to support node-specific callbacks in `add_state_by_tasks` and `add_conditional_edge`.
>   - **Node Class**:
>     - Modify `Node` class in `node.py` to include `callbacks` attribute and integrate node-specific callbacks in execution flow.
>   - **Testing**:
>     - Add `test_callbacks.py` to test node-specific callback functionality.
>     - Update `test_cohere.py` to reflect changes in output structure.
>   - **Examples**:
>     - Add `writer_assistant.py` example demonstrating the use of node-specific callbacks in a workflow.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 14600796eec89132e507fd9666213c475a9d7f8d. It will automatically update as commits are pushed.</sup>


vitalii-dynamiq:
what's the point of these? why only for graph orchestrator?

maksDev123:
> what's the point of these? why only for graph orchestrator?

This allows creating callbacks for specific nodes. It can be used for HITL (and not only). With that we can develop pipeline more easily where workflow does not have finish to send reply, allowing creating similar to CopilotKIT examples (inside this callback can be communication with frontend). This callbacks can be used for any node thus allowing controller HITL for all agents and orchestrators.

dynamiq/nodes/node.py
```diff
@@ -511,7 +512,7 @@ def run(
 
             transformed_input = self.transform_input(input_data=input_data, depends_result=depends_result)
 
-            self.run_on_node_start(config.callbacks, transformed_input, **merged_kwargs)
+            transformed_input |= self.run_on_node_start(config.callbacks, transformed_input, **merged_kwargs)
```
acoola, line 515: The callback shouldn't modify input/output. I such cases, that is some other transformers. We already have different type of transformers.
Callbacks are a way just to call smth without affecting the initial flow and data.

maksDev123: Refactored, leaving Node callbacks that does not change anything.

acoola (approved):

PR #93, opened 2024-12-16
feat: add cohere reranker

olbychos:


dynamiq/nodes/rankers/cohere.py
```diff
@@ -0,0 +1,89 @@
...
+
+            return reranked_documents[: self.top_k]
+
+        except Exception as e:
+            logger.error(f"Node {self.name} - {self.id}: Error during reranking: {str(e)}")
+            raise
+
+    def init_components(self, connection_manager=None):
+        """Initialize the reranker component with the Cohere connection."""
+        super().init_components(connection_manager)
```
acoola, line 87: Cohere reranker don't have any components (like other nodes, llms, embedders and so on). So this method is obsolete here. Connection is mandatory in node definition too.

dynamiq/nodes/rankers/cohere.py
```diff
@@ -0,0 +1,89 @@
...
+                return {"documents": []}
+
+            document_texts = [doc.content for doc in documents]
+
+            logger.debug(f"Node {self.name} - {self.id}: Reranking {len(documents)} documents")
+
+            response = rerank(model=self.model, query=query, documents=document_texts, top_n=self.top_k)
+
+            reranked_documents = []
+            print(response)
```
acoola, line 71: Please remove print

dynamiq/nodes/rankers/cohere.py
```diff
@@ -0,0 +1,89 @@
+from typing import Any, ClassVar, Literal
+
+from litellm import rerank
```
acoola, line 3: Please add `rerank` as private attr on init similar to completion and other litellm helpers. Please optimize import in a similar way to speed up loading.

dynamiq/nodes/rankers/cohere.py
```diff
@@ -0,0 +1,93 @@
...
+                doc = documents[result.get("index")]
+                doc.score = result.get("relevance_score")
+                if doc.score > self.threshold:
+                    reranked_documents.append(doc)
+
+            logger.debug(f"Node {self.name} - {self.id}: Successfully reranked {len(reranked_documents)} documents")
+
+            return reranked_documents[: self.top_k]
+
+        except Exception as e:
```
acoola, line 91: I don't think we need this extra try..except here as we don't do any specific logic. We already have logging on high level in case of error.

acoola (approved):

PR #94, opened 2024-12-16
feat: add YAML dump logic

acoola:
- update connection types
- add YAML dumper
- update Retriever tool
- update ScaleSerp connection and node
- update node dump functionality
- add dump logic to flows and workflows

### DIFF
--- a/dynamiq/serializers/dumpers/yaml.py
+++ b/dynamiq/serializers/dumpers/yaml.py
@@ -1,0 +1,202 @@
+import enum
+from os import PathLike
+from typing import IO, Any
+
+from dynamiq.connections import BaseConnection
+from dynamiq.flows import Flow
+from dynamiq.nodes import Node
+from dynamiq.serializers.types import WorkflowYamlData
+from dynamiq.utils.logger import logger
+
+
+class WorkflowYAMLDumperException(Exception):
+    pass
+
+
+class WorkflowYAMLDumper:
+    """Dumper class for parsing workflow components and save them to YAML."""
+
+    @classmethod
+    def get_updated_node_data(
+        cls,
+        node_data: dict,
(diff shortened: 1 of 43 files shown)


dynamiq/serializers/dumpers/yaml.py
```diff
@@ -0,0 +1,202 @@
...
+            A dictionary of newly created nodes with dependencies.
+        """
+        updated_node_init_data = {}
+        for param_name, param_data in node_data.items():
+            if param_name == "depends":
+                updated_node_init_data[param_name] = [
+                    {"node": dep["node"]["id"], "option": dep["option"]} for dep in param_data
+                ]
+
+            elif param_name == "connection":
```
delamainer, line 44: What about `prompt`? Or are we always going to dump it inline?

acoola: Currently just supports inline prompts.

delamainer (approved):

dynamiq/workflow/workflow.py
```diff
@@ -93,6 +91,57 @@ def from_yaml_file_data(
...
+        Returns:
+            WorkflowYamlData: Data for the YAML dump.
+        """
+        from dynamiq.serializers.loaders.yaml import WorkflowYamlData
+
+        return WorkflowYamlData(
+            workflows={self.id: self},
+            flows={self.flow.id: self.flow},
+            nodes={node.id: node for node in self.flow.nodes},
+            connections={},
```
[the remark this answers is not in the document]
acoola: We need to go over each node and each component of the node (embedders, retrievals, agents and so on) recursively. 
This has already been done in Yaml dumper so currently didn't add same logic into workflow. But potentially could be added into Workflow too.

PR #99, opened 2024-12-19
refactor: predefined metrics update

MaksymAI:
Unify parameter names across predefined metrics: `questions`, `answers`, `contexts`, and `ground_truth_answers`. 

The only exception for now is the `ContextPrecision` metric, which expects a list of lists `(list[list[str]])` instead of a list of strings `(list[str])`; therefore, it is currently named `contexts_list`. The approach that will allow for the unification of this name will be proposed in a separate PR later.

### DIFF
--- a/dynamiq/evaluations/metrics/answer_correctness.py
+++ b/dynamiq/evaluations/metrics/answer_correctness.py
@@ -111,21 +111,21 @@ class RunInput(BaseModel):
     Input model for running the evaluator.
 
     Attributes:
-        questions (List[str]): The list of questions.
-        answers (List[str]): The list of answers.
-        ground_truths (List[str]): The list of ground truth texts.
+        question (List[str]): The list of questions.
+        answer (List[str]): The list of answers.
+        ground_truth_answer (List[str]): The list of ground truth answers.
         verbose (bool): Flag to enable verbose logging.
     """
 
-    questions: list[str]
-    answers: list[str]
-    ground_truths: list[str]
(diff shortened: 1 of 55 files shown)


dynamiq/evaluations/metrics/answer_correctness.py
```diff
@@ -111,21 +111,21 @@ class RunInput(BaseModel):
...
+        question (List[str]): The list of questions.
+        answer (List[str]): The list of answers.
+        ground_truth_answer (List[str]): The list of ground truth answers.
         verbose (bool): Flag to enable verbose logging.
     """
 
-    questions: list[str]
-    answers: list[str]
-    ground_truths: list[str]
+    question: list[str]
```
acoola, line 120: A little bit weird renaming and this list of questions.

dynamiq/evaluations/metrics/context_recall.py
```diff
@@ -199,41 +199,41 @@ def _initialize_evaluator(self):
...
         )
 
         final_scores = []
 
-        for idx in range(len(input_data.questions)):
-            question = input_data.questions[idx]
-            context = input_data.contexts[idx]
-            answer = input_data.answers[idx]
+        for idx in range(len(input_data.question)):
+            single_question = input_data.question[idx]
```
acoola, line 229: Please use `questions` and `question` naming. This is clear and transparent way

acoola (approved):
Great job

PR #115, opened 2025-01-13
fix: update message convertion

maksDev123:
> [!IMPORTANT]
> Update `format_messages` in `prompts.py` to convert single `VisionMessageTextContent` items from a list to a string.
> 
>   - **Behavior**:
>     - In `format_messages` in `prompts.py`, single `VisionMessageTextContent` items are now converted from a list to a string if they are the only item in `out_msg_content`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for e7e5da1b378b1a9633028498e81833a378355c0c. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/prompts/prompts.py
+++ b/dynamiq/prompts/prompts.py
@@ -244,6 +244,10 @@ class Prompt(BasePrompt):
                         )
                     else:
                         raise ValueError(f"Invalid content type: {content.type}")
+
+                if len(out_msg_content) == 1:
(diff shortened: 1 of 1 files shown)


maksDev123:
This PR simplifies VisionMessage to Message in case it has only text. 
This is done as not all models support VisionMessage like type of content.

PR #125, opened 2025-01-22
feat: add coda node

olbychos:
add Coda node, with a list of actions with docs, pages


vitalii-dynamiq:
I think this should be in the tools repo

PR #130, opened 2025-02-01
feat: add support for o3 models & update litellm

olbychos:
> [!IMPORTANT]
> Add support for O3 models, update `litellm`, and enhance `OpenAI` class for O-series models with new example and tests.
> 
>   - **Behavior**:
>     - Add `update_completion_params()` in `base.py` to allow subclasses to modify completion parameters.
>     - Modify `execute()` in `base.py` to use `update_completion_params()`.
>     - Add `is_o_series_model` and override `update_completion_params()` in `openai.py` to handle O-series models.
>   - **Examples**:
>     - Add `adaptive_article_o3.py` to demonstrate adaptive orchestrator with O3 models.
>   - **Dependencies**:
>     - Update `litellm` version to `1.59.12` in `pyproject.toml`.
>   - **Tests**:
>     - Add `api_key` to `test_flow.py` and `test_ollama.py` to ensure correct API usage.
>     - Update `test_flow.py` to handle new parameter updates in `OpenAI` class.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 6b675d54ef4dd0a5c50f6ddaacbb44ba74c676d8. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/nodes/converters/csv.py
+++ b/dynamiq/nodes/converters/csv.py
@@ -21,7 +21,7 @@ class CSVConverterInputSchema(BaseModel):
         files (list[BytesIO | bytes] | None): List of file objects or bytes containing CSV data.
         delimiter (str): Character used to separate fields in the CSV files. Defaults to comma.
         content_column (str): Name of the column to use as the main document content.
-        metadata_columns (list[str]): Column names to extract as metadata for each document.
+        metadata_columns (list[str] | None): Column names to extract as metadata for each document.
     """
 
     file_paths: list[str] | None = Field(
@@ -30,10 +30,20 @@ class CSVConverterInputSchema(BaseModel):
     files: list[BytesIO | bytes] | None = Field(
         default=None, description="List of file objects or bytes representing CSV files."
     )
-    delimiter: str = Field(default=",", description="Delimiter used in the CSV files.")
-    content_column: str = Field(..., description="Name of the column that will be used as the document's main content.")
-    metadata_columns: list[str] = Field(
-        default_factory=list, description="List of column names to extract as metadata for each document."
+    delimiter: str | None = Field(
+        default=None,
+        description="Delimiter used in the CSV files. If not provided, the Node's configured delimiter is used."
+    )
+    content_column: str | None = Field(
+        default=None, description="Name of the column that will be used as the document's main content."
+    )
+    metadata_columns: list[str] | None = Field(
+        default=None,
+        description="Optional list of column names to extract as metadata for each document. Can be None.",
+    )
+    metadata: dict | list | None = Field(
+        default=None,
+        description="External metadata to be merged with metadata extracted from CSV rows."
     )
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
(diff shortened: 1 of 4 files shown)


dynamiq/nodes/llms/openai.py
```diff
@@ -21,3 +28,87 @@ def __init__(self, **kwargs):
         if kwargs.get("client") is None and kwargs.get("connection") is None:
             kwargs["connection"] = OpenAIConnection()
         super().__init__(**kwargs)
+
+    def is_o_family(self) -> bool:
```
vitalii-dynamiq, line 32: can you please check how it's done in litellm? I guess it's better to do something a bit more generic like is_reasoning_model

dynamiq/nodes/llms/openai.py
```diff
@@ -21,3 +28,87 @@ def __init__(self, **kwargs):
...
+        """
+        config = ensure_config(config)
+        prompt = prompt or self.prompt or Prompt(messages=[], tools=None)
+        messages = prompt.format_messages(**dict(input_data))
+        base_tools = prompt.format_tools(**dict(input_data))
+        self.run_on_node_execute_run(callbacks=config.callbacks, prompt_messages=messages, **kwargs)
+
+        params = self.connection.conn_params
+        if self.client and not isinstance(self.connection, HttpApiKey):
+            params = {"client": self.client}
```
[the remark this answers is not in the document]
vitalii-dynamiq: @olbychos needs to be addressed

dynamiq/nodes/llms/openai.py
```diff
@@ -21,3 +28,88 @@ def __init__(self, **kwargs):
...
+            "top_p": self.top_p,
+            "seed": self.seed,
+            "presence_penalty": self.presence_penalty,
+            "frequency_penalty": self.frequency_penalty,
+            "response_format": response_format,
+            "drop_params": True,
+            **params,
+        }
+
+        if self.is_o_series_model():
```
acoola, line 101: Instead of copy/paste logic of base LLM, please add into BaseLLM extra method like `def update_completion_params` that by default don't do any action and just return passed params. In `OpenAI` please override it with the update `completion_params` logic,

dynamiq/nodes/llms/openai.py
```diff
@@ -21,3 +28,88 @@ def __init__(self, **kwargs):
...
         super().__init__(**kwargs)
+
+    def is_o_series_model(self) -> bool:
+        """Determine if the model belongs to the O-series (e.g. o1 or o3).
+
+        Returns:
+            bool: True if the model is an O-series model, otherwise False.
+        """
+        model_lower = self.model.lower()
+        return "o1" in model_lower or "o3" in model_lower
```
acoola, line 39: Please define some ClassVar constant tuple like `o_series_model_prefixes`. After go over this tuple and check if the model starts with this prefix (not just check for substring).

dynamiq/nodes/llms/openai.py
```diff
@@ -21,3 +28,88 @@ def __init__(self, **kwargs):
         if kwargs.get("client") is None and kwargs.get("connection") is None:
             kwargs["connection"] = OpenAIConnection()
         super().__init__(**kwargs)
+
+    def is_o_series_model(self) -> bool:
```
acoola, line 32: It seems this could be `@cached_property` as we don't need to recheck each time as we don't update the model dynamically.

poetry.lock
```diff
@@ -1,11 +1,13 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
acoola, line 1: Please use poetry 1.8.3

vitalii-dynamiq (approved):

poetry.lock
```diff
@@ -13,13 +13,13 @@ files = [
 
 [[package]]
 name = "accelerate"
-version = "0.34.2"
+version = "1.3.0"
```
acoola, line 16: Please update and lock just the desired lib (litellm). Updating all dependencies is so dangerous
Use `poetry lock --no-update` to lock updated deps based on toml file

acoola (approved):

PR #158, opened 2025-02-24
feat: add single run for metric

MaksymAI:
> [!IMPORTANT]
> Add single-run evaluation methods for BLEU, ROUGE, and string similarity metrics, enhancing granularity of evaluations.
> 
>   - **Behavior**:
>     - Add `run_single()` method to `BleuScoreEvaluator`, `RougeScoreEvaluator`, `ExactMatchEvaluator`, `StringPresenceEvaluator`, and `StringSimilarityEvaluator` for single pair evaluation.
>     - Update `run()` methods to utilize `run_single()` for batch processing.
>   - **Models**:
>     - Introduce `RunSingleInput` and `RunSingleOutput` models for single-run evaluations in `bleu_score.py`, `rouge_score.py`, and `string_metrics.py`.
>   - **Examples**:
>     - Update `bleu_score.py`, `rouge_score.py`, and `string_metrics.py` in `examples/evaluation/metrics/` to demonstrate single-run evaluations.
>     - Add single-run evaluation examples to `python_evaluator.py`.
>   - **Tests**:
>     - Update `test_bleu_score.py` to validate new BLEU score calculations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for c6bd39b1d98d8aa62662f71c4bf92037f0436141. It will automatically update as commits are pushed.</sup>


vitalii-dynamiq:
@MaksymAI will it break catalyst?

MaksymAI:
> @MaksymAI will it break catalyst?

No, it doesn't contain any breaking changes for Catalyst.

All previous input/output functions remain the same, and everything will work without any additional adjustments.

This PR adds a new method, `run_single`, which will allow users to run evaluations for a single example without the need to run metrics for the full batch. 
This new method is currently not used in Catalyst, but it can be utilized for the new API for metric testing in the playground on a single example, before we run the evaluation on the full dataset.

dynamiq/evaluations/metrics/bleu_score.py
```diff
@@ -78,41 +95,54 @@ def _initialize_bleu(self) -> None:
...
+            answer (str): The candidate answer.
+
+        Returns:
+            float: The computed BLEU score (as a fraction, e.g., 0.75 for 75%).
         """
-        Compute BLEU scores for each reference-response pair.
+        # Validate inputs using the Pydantic model.
+        single_input = RunSingleInput(ground_truth_answer=ground_truth_answer, answer=answer)
+
+        ref_sentences = single_input.ground_truth_answer.split(". ")
```
maksDev123, line 116: Maybe add splits by "." not ". ".

maksDev123 (approved):

dynamiq/evaluations/metrics/bleu_score.py
```diff
@@ -1,75 +1,92 @@
...
+    Single-run output model for BLEU score evaluation.
+
+    Attributes:
+        score (float): The computed BLEU score.
+    """
+
+    score: float
+
+
 class BleuScoreEvaluator(BaseEvaluator):
```
maksDev123, line 68: We might need to do some preprocessing for text to make score more informative.
    
    gt_single = "The cat sits on the mat."
    ans_single = "The cat, sits on the mat"
    
    BLEU Score (Single): 0.49
    
Gives 0.85.

maksDev123:
Looks good to me, everything works. Just not sure about pre-processing.

dynamiq/evaluations/metrics/bleu_score.py
```diff
@@ -1,75 +1,92 @@
...
+    Attributes:
+        ground_truth_answer (str): The reference answer.
+        answer (str): The candidate answer.
+    """
+
+    ground_truth_answer: str
+    answer: str
+
+
+class RunSingleOutput(BaseModel):
```
anastasiasenyk, line 60: It looks like `RunSingleOutput` class has never been used. Should it be incorporated into `run_single()`, similar to `run()`? 
Same to rouge_score.py

examples/evaluation/metrics/rouge_score.py
```diff
@@ -16,22 +16,32 @@ def main():
...
+    print("\nSingle Evaluation:")
+    gt_single = "The cat sits on the mat."
+    ans_single = "The cat sits on the mat."
+    single_score = rouge_evaluator.run_single(ground_truth_answer=gt_single, answer=ans_single)
+    print("Ground Truth Answer:", gt_single)
+    print("System Answer:", ans_single)
+    print(f"ROUGE Score ({rouge_evaluator.rouge_type} - {rouge_evaluator.measure_type}):", single_score)
+
 
 if __name__ == "__main__":
```
anastasiasenyk, line 46: Getting an import error when running this example file because of the line `from rouge_score import rouge_scorer` in `dynamiq/evaluations/metrics/rouge_score.py`. 
Python is attempting to import rouge_scorer from the current file rather than the intended module.
Maybe it's possible to rename the file?

dynamiq/evaluations/metrics/bleu_score.py
MaksymAI: Updated splitting logic into sentences

anastasiasenyk (commented):
Looking good! Just minor comments

examples/evaluation/metrics/rouge_score.py
MaksymAI: Let's keep it as it is for now; I haven't encountered such errors before. I don't want to change the file name, as it will introduce a breaking change. For now, let's keep it as it is, and if needed, we can rename it later.

anastasiasenyk (approved):

PR #166, opened 2025-02-26
feat: add base url for OpenAI connection

TrachukT:
> [!IMPORTANT]
> Add configurable base URL for OpenAI connection in `connections.py` and update tests in `test_flow.py`.
> 
>   - **Behavior**:
>     - Add `url` attribute to `OpenAI` in `connections.py`, defaulting to `https://api.openai.com/v1`.
>     - Modify `connect()` in `OpenAI` to use `base_url` parameter.
>     - Add `conn_params` property to return connection parameters including `api_base`.
>   - **Tests**:
>     - Update `test_flow.py` to check `api_base` parameter in `mock_llm_executor` calls.
>     - Ensure tests verify the correct URL is used in OpenAI connections.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 2cb9b9f508cc1fae629563b5d0e306cf54ac3ff8. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/connections/connections.py
+++ b/dynamiq/connections/connections.py
@@ -212,13 +212,6 @@ class OpenAI(BaseApiKeyConnection):
     def conn_params(self) -> dict:
         """
         Returns the parameters required for connection.
-
-        Returns:
-            dict: A dictionary containing
-
-                -the API key with the key 'api_key'.
-
-                -the base url with the key 'api_base'.
         """
         return {
             "api_base": self.url,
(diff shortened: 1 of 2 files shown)


dynamiq/connections/connections.py
```diff
@@ -201,10 +203,28 @@ def connect(self) -> "OpenAIClient":
...
 
+    @property
+    def conn_params(self) -> dict:
+        """
+        Returns the parameters required for connection.
+
+        Returns:
+            dict: A dictionary containing
+
+                -the API key with the key 'api_key'.
```
acoola, line 219: We don't specify all keys in dicts in case of return. Here all clear without this info in docstring

acoola (approved):

PR #170, opened 2025-03-01
feat: streaming websockets support

maksDev123:


vitalii-dynamiq:
@maksDev123 is this ready for review?

PR #209, opened 2025-03-24
feat: add one agent class for react and simple

olbychos:
> [!IMPORTANT]
> Consolidates `ReActAgent` and `SimpleAgent` into a single `Agent` class, updating examples and tests accordingly.
> 
>   - **Agent Consolidation**:
>     - Merges `ReActAgent` and `SimpleAgent` into a single `Agent` class in `base.py`.
>     - Removes `react.py` and `simple.py` as they are no longer needed.
>   - **Examples and Tests**:
>     - Updates all example scripts to use the new `Agent` class instead of `ReActAgent` and `SimpleAgent`.
>     - Modifies test cases in `test_memory.py` to use the `Agent` class.
>   - **Dependencies**:
>     - Updates `pyproject.toml` to reflect changes in dependencies, including version updates for `openai` and `litellm`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for e74e3f1c4edff63875fcc191bdb507ad1b1d6148. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/nodes/agents/base.py
+++ b/dynamiq/nodes/agents/base.py
@@ -862,7 +862,7 @@ class Agent(Node):
                     if call_info["function"]["name"] == "provide_final_answer":
                         final_ans = call_info["function"]["arguments"]["answer"]
                         if self.streaming.enabled:
-                            self.stream_content(final_ans, source=self.name, step="final_answer", config=config)
+                            self.stream_content(final_ans, source=self.name, step="final_answer", config=config, **kwargs)
                         return final_ans
                 raise ValueError("Expected 'provide_final_answer' call not found.")
             raise ValueError("No tool calls in FUNCTION_CALLING mode.")
@@ -873,7 +873,7 @@ class Agent(Node):
                 data = json.loads(content)
                 final_ans = data["answer"]
                 if self.streaming.enabled:
-                    self.stream_content(final_ans, source=self.name, step="final_answer", config=config)
+                    self.stream_content(final_ans, source=self.name, step="final_answer", config=config, **kwargs)
                 return final_ans
             except (json.JSONDecodeError, KeyError) as e:
                 raise ValueError(f"Invalid JSON or missing 'answer' in STRUCTURED_OUTPUT: {e}")
@@ -889,7 +889,7 @@ class Agent(Node):
                 final_text = content.strip()
 
         if self.streaming.enabled:
-            self.stream_content(final_text, source=self.name, step="final_answer", config=config)
+            self.stream_content(final_text, source=self.name, step="final_answer", config=config, **kwargs)
         return final_text
 
     def _run_react(
@@ -951,7 +951,7 @@ class Agent(Node):
                     if self.streaming.enabled and self.streaming.mode == StreamingMode.ALL:
                         chunk_content = {"thought": thought, "action": func_name, "action_input": action_input}
                         self.stream_content(
-                            chunk_content, source=self.name, step=f"reasoning_{loop_idx}", config=config
+                            chunk_content, source=self.name, step=f"reasoning_{loop_idx}", config=config, **kwargs
                         )
 
                     try:
@@ -1017,7 +1017,7 @@ class Agent(Node):
             self.log_reasoning(thought, action, action_input, loop_idx)
             if self.streaming.enabled and self.streaming.mode == StreamingMode.ALL:
                 chunk_content = {"thought": thought, "action": action, "action_input": action_input}
-                self.stream_content(chunk_content, source=self.name, step=f"reasoning_{loop_idx}", config=config)
+                self.stream_content(chunk_content, source=self.name, step=f"reasoning_{loop_idx}", config=config, **kwargs)
 
             if action:
                 try:


dynamiq/nodes/agents/base.py
```diff
@@ -55,9 +70,216 @@
 {%- endif %}
 """
 
+REACT_BLOCK_TOOLS = """
```
maksymbuleshnyi, line 73: Maybe we can keep our prompts separated, file gets too big.

dynamiq/nodes/agents/base.py
```diff
@@ -458,259 +1014,532 @@ def _run_llm(self, messages: list[Message | VisionMessage], config: RunnableConf
...
+            output_text = llm_result.output.get("content", "")
+            tool_calls = llm_result.output.get("tool_calls", {})
+
+            self._intermediate_steps[loop_idx] = AgentIntermediateStep(
+                input_data={"prompt": self._prompt.messages},
+                model_observation=AgentIntermediateStepModelObservation(initial=output_text or str(tool_calls)),
+            ).model_dump()
+
+            if self.inference_mode == InferenceMode.FUNCTION_CALLING and tool_calls:
+                for call_id, call_info in tool_calls.items():
```
maksymbuleshnyi, line 1168: call_id is not used.

dynamiq/nodes/agents/base.py
```diff
@@ -458,259 +1014,532 @@ def _run_llm(self, messages: list[Message | VisionMessage], config: RunnableConf
...
+                        )
+
+                    try:
+                        tool = self._get_tool(func_name)
+                        tool_result = self._run_tool(tool, action_input, config, **kwargs)
+
+                    except RecoverableAgentException as e:
+                        tool_result = f"{type(e).__name__}: {str(e)}"
+
+                    obs_msg = f"\nObservation: {tool_result}\n"
```
maksymbuleshnyi, line 1195: No logging for intermediate steps. It would be great to have it.

dynamiq/nodes/agents/base.py
```diff
@@ -458,259 +1014,532 @@ def _run_llm(self, messages: list[Message | VisionMessage], config: RunnableConf
...
+
+            output_text = llm_result.output.get("content", "")
+            tool_calls = llm_result.output.get("tool_calls", {})
+
+            self._intermediate_steps[loop_idx] = AgentIntermediateStep(
+                input_data={"prompt": self._prompt.messages},
+                model_observation=AgentIntermediateStepModelObservation(initial=output_text or str(tool_calls)),
+            ).model_dump()
+
+            if self.inference_mode == InferenceMode.FUNCTION_CALLING and tool_calls:
```
maksymbuleshnyi, line 1167: Does not handle empty tool calls.

examples/components/agents/streaming/intermediate_streaming/agent.py
maksymbuleshnyi, line 1: Please revert. Or there is some reason for removal?

maksymbuleshnyi:
For some reason agent fails with STRUCTURED_OUTPUT inference mode (Scale Serp tool).

```LLM 'OpenAI LLM' failed: litellm.APIConnectionError: APIConnectionError: OpenAIException - issubclass() arg 1 must be a class```

File: dynamiq/nodes/agents/base.py [hunk 5]
```diff
@@ -162,7 +169,13 @@ def validate_input_fields(self, context):
 
 
 class Agent(Node):
```
anastasiasenyk, line 171: If we're keeping the ReflectionAgent, wouldn't it be better to keep React + Simple as a separate class? 

- Having both Agent and ReflectionAgent without a clear distinction could be confusing.

- Also, if we add more agent types in the future, keeping redundant functions in the base class might make maintenance more difficult. 

What do you think?

anastasiasenyk (commented):
Hey, have we decided to remove the ReflectionAgent as well? If so, you'll also need to delete its file from `dynamiq/nodes/agents/reflection.py`. Also, could you update the `README.md` to reflect the new changes?

dynamiq/nodes/agents/base.py
vitalii-dynamiq: I think we keep one agent. For simplicity

anastasiasenyk (commented):
Could you please look into the following issue:
I'm trying to run orchestrators (Adaptive, Linear) with agents, and when the LLM is called in the `_run_react` or `_run_simple` function, it throws an error that eventually results in:  
```Agentdue to error: dynamiq.nodes.node.Node.run() got multiple values for keyword argument 'run_depends'```

File: dynamiq/nodes/agents/base.py [hunk 6]
```diff
@@ -229,81 +280,288 @@ def to_dict(self, **kwargs) -> dict:
...
+                    },
+                    "required": ["thought", "answer"],
+                },
+            },
+        }
+        self.format_schema.append(final_answer_schema)
+
+        for tool in self.tools:
+            try:
+                properties = {}
```
maksymbuleshnyi, line 420: This function calling function generation is not full comparing to what we had.

dynamiq/nodes/agents/base.py
```diff
@@ -458,259 +732,588 @@ def _run_llm(self, messages: list[Message | VisionMessage], config: RunnableConf
...
+            "\n------------------------------------------\n"
+            f"Agent {self.name}: Loop {loop_num}:\n"
+            f"Thought: {thought}\n"
+            f"Action: {action}\n"
+            f"Action Input: {action_input}"
+            "\n------------------------------------------"
+        )
 
 
 class AgentManagerInputSchema(BaseModel):
```
maksymbuleshnyi, line 1319: Lets put Manager to a separate file, it will be easier to follow.

dynamiq/nodes/agents/base.py
```diff
@@ -458,259 +732,588 @@ def _run_llm(self, messages: list[Message | VisionMessage], config: RunnableConf
...
+        )
+
+        if self.inference_mode == InferenceMode.FUNCTION_CALLING:
+            tool_calls = llm_result.output.get("tool_calls", {})
+            if tool_calls:
+                for _, call_info in tool_calls.items():
+                    if call_info["function"]["name"] == "provide_final_answer":
+                        final_ans = call_info["function"]["arguments"]["answer"]
+                        if self.streaming.enabled:
+                            self.stream_content(final_ans, source=self.name, step="final_answer", config=config)
```
maksymbuleshnyi, line 865: Inside `stream_content` kwargs have to be passed. Streaming requires run_id, which is passed through kwargs.

maksymbuleshnyi: fixed

PR #221, opened 2025-03-31
fix: transformer validation

maksymbuleshnyi:
> [!IMPORTANT]
> Fixes logic and improves error logging in `validate_input_transformer` in `graph_state.py`.
> 
>   - **Behavior**:
>     - Fixes logic in `validate_input_transformer` in `graph_state.py` to handle cases where `input_transformer` is not defined or lacks `path` and `selector`.
>     - Improves error logging in `validate_input_transformer` to include task name and exception details.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for db6a987bb26de4f6d15c6d409280e0259ecbe05b. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/nodes/agents/orchestrators/graph_state.py
+++ b/dynamiq/nodes/agents/orchestrators/graph_state.py
@@ -129,7 +129,7 @@ class GraphState(Node):
             bool: Whether input data is correct.
         """
 
-        if task.input_transformer.path or task.input_transformer.selector:
+        if task.input_transformer and (task.input_transformer.path or task.input_transformer.selector):
             try:
                 output = task.transform(input_data, task.input_transformer, task.id)
                 task.validate_input_schema(output, **kwargs)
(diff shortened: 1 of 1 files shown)


dynamiq/nodes/agents/orchestrators/graph_state.py
```diff
@@ -129,7 +129,7 @@ def validate_input_transformer(self, task: Node, input_data: dict[str, Any], **k
             bool: Whether input data is correct.
         """
 
-        if task.input_transformer:
+        if task.input_transformer and (task.input_transformer.path or task.input_transformer.selector):
```
MaksymAI, line 132: Please add logging of the exception reasoning for why the validation failed. I think such information could be helpful in understanding what's going on and would provide greater transparency regarding what should be improved.

MaksymAI: Also, if you execute the try clause, maybe it's worth moving this if statement under the try clause as well?

acoola (approved):

PR #230, opened 2025-04-07
fix: update message prefix

olbychos:
> [!IMPORTANT]
> Removes `prefix` attribute from `Message` and updates message formatting logic in `BaseLLM` and `Mistral` classes.
> 
>   - **Behavior**:
>     - Removes setting of `prefix` attribute for `Message` in `get_all()`, `search()`, and `get_agent_conversation()` in `memory.py`.
>     - Introduces `get_messages()` in `base.py` and `mistral.py` to format messages, setting `prefix` for last `ASSISTANT` message.
>   - **Models**:
>     - Removes `prefix` attribute from `Message` in `prompts.py`.
>   - **Tests**:
>     - Updates `test_callbacks.py` to remove `prefix` from test message data.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for b95c4026f7b4d8094f5da921516dbe8fbfe75b40. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/nodes/llms/base.py
+++ b/dynamiq/nodes/llms/base.py
@@ -140,11 +140,16 @@ class BaseLLM(ConnectionNode):
         """Provides context for input schema that is required for proper validation."""
         return {"instance_prompt": self.prompt}
 
-    def format_messages_for_provider(self, messages: list[dict]) -> list[dict]:
+    def get_messages(
+        self,
+        prompt,
+        input_data,
+    ) -> list[dict]:
         """
         Format and filter message parameters based on provider requirements.
         Override this in provider-specific subclasses.
         """
+        messages = prompt.format_messages(**dict(input_data))
         return messages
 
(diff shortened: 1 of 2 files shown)


dynamiq/nodes/llms/base.py
```diff
@@ -140,6 +140,13 @@ def get_context_for_input_schema(self) -> dict:
         """Provides context for input schema that is required for proper validation."""
         return {"instance_prompt": self.prompt}
 
+    def format_messages_for_provider(self, messages: list[dict]) -> list[dict]:
```
acoola, line 143: Let's maybe call this `def get_messages` and move 
`return prompt.format_messages(**dict(input_data))` here as default?
For Mistral also make extra formatting

PR #245, opened 2025-04-21
feat: add MCP server tool support

anastasiasenyk:
> [!IMPORTANT]
> Adds support for MCP servers as tools, including new connection types, tool management classes, examples, and tests.
> 
>   - **Behavior**:
>     - Adds `MCPSse`, `MCPStreamableHTTP`, and `MCPStdio` classes in `connections.py` for different MCP server connection types.
>     - Introduces `MCPTool` and `MCPServer` classes in `mcp.py` to manage and execute tools from MCP servers.
>     - Updates `Agent` class in `base.py` to handle MCP tools, expanding tools list with subtools from `MCPServer`.
>   - **Examples**:
>     - Adds `dag_mcp_server.yaml` and `dag_mcp_tool.yaml` for workflow examples using MCP servers and tools.
>     - Provides Python scripts in `dag_yaml_mcp.py` and `use_mcp_adapter_tool.py` to demonstrate setting up and using MCP tools.
>     - Includes `math_server.py` and `weather_server.py` as example MCP servers.
>   - **Tests**:
>     - Adds `test_mcp_tool.py` to test MCP tool functionality, including tool initialization, execution, and filtering.
>   - **Dependencies**:
>     - Adds `mcp` and `datamodel-code-generator` to `pyproject.toml` dependencies.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for f99b5e30feb2292ec13bd8cc6d6984c635f81f19. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,247 @@
...
+
+from dynamiq.nodes import Node, NodeGroup
+from dynamiq.nodes.agents.exceptions import ToolExecutionException
+from dynamiq.nodes.node import ensure_config
+from dynamiq.runnables import RunnableConfig
+from dynamiq.utils import is_called_from_async_context
+from dynamiq.utils.logger import logger
+
+
+class SseServerParameters(BaseModel):
```
vitalii-dynamiq, line 18: SSE

dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,247 @@
...
+                f"Tool '{self.name}' failed to call tool from the MCP server."
+                f"Error: {str(e)}. Please analyze the error and take appropriate action.",
+                recoverable=True,
+            )
+
+        logger.info(f"Tool {self.name} - {self.id}: finished with result:\n{str(result)[:200]}...")
+        return {"content": result}
+
+
+class MCPAdapterTool(Node):
```
vitalii-dynamiq, line 174: what's the difference between MCPTool and MCPAdapterTool?

anastasiasenyk: MCP servers can host multiple tools. To handle this, I introduced the MCPAdapterTool, which establishes the initial connection to an MCP server. It then initialises individual MCPTool instances for each tool available on the server. 
This way, the ReAct agent uses each tool as a separate entity with independent input schemas, names, and descriptions.

dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,247 @@
...
+    def get_mcp_tools(self) -> list[MCPTool]:
+        """
+        Synchronously fetches and initializes MCP tools if not already available.
+
+        Returns:
+            list[MCPTool]: A list of initialized MCPTool instances.
+        """
+        if is_called_from_async_context():
+            with ThreadPoolExecutor() as executor:
+                future = executor.submit(lambda: asyncio.run(self.get_mcp_tools_async()))
```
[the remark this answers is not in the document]
anastasiasenyk: This is done to support calling the agent from an async context and not to block the main event loop.

Since the agent prepares all prompt templates during initialisation and before its own execute(), we need the actual result of `get_mcp_tools_async()`, not a coroutine or task, so it can be used immediately within the agent’s synchronous logic.

dynamiq/nodes/tools/mcp_adapter.py
vitalii-dynamiq: maybe MCPServerTool? just curious if there's a better name that is easy to understand for a new person

anastasiasenyk: I agree, MCPServer / MCPServerAdapter / MCPServerTool sounds better.

dynamiq/nodes/tools/mcp_adapter.py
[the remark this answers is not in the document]
acoola: Need to check this, maybe there could be a more elegant solution

dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,248 @@
...
+        Asynchronously fetches and initializes MCP tools if not already available.
+
+        Returns:
+            list[MCPTool]: A list of initialized MCPTool instances.
+        """
+        if not self.mcp_tools:
+            await self.initialize_tools()
+        return self.mcp_tools
+
+    def execute(self, **kwargs):
```
acoola, line 241: Can we also here return list of initialized MCP tools? Maybe under some dict key

dynamiq/connections/connections.py
```diff
@@ -1232,3 +1233,62 @@ def conn_params(self) -> dict:
             "api_base": self.url,
             "api_key": self.api_key,
         }
+
+
+class MPC(BaseConnection):
+    # SSE
```
acoola, line 1239: Let's add proper Pydantic model validator to validate mandatory params for each of connection types. Let's do this on connection init, not on client init

dynamiq/connections/connections.py
```diff
@@ -1232,3 +1233,62 @@ def conn_params(self) -> dict:
...
+        default="utf-8", description="The text encoding used when sending/receiving messages to the server."
+    )
+    encoding_error_handler: Literal["strict", "ignore", "replace"] = Field(
+        default="strict", description="The text encoding error handler."
+    )
+
+    def connect(self):
+        pass
+
+    async def get_client(self):
```
acoola, line 1262: It seems this logic returns just client, so it could be moved to `def connect` as it done in the rest connections

examples/components/tools/mcp_server_as_tool/use_mcp_adapter_tool.py
```diff
@@ -0,0 +1,189 @@
...
+
+def use_local_server_with_token():
+    """
+    Runs a local MCP server using an access token via NPX subprocess.
+
+    Returns:
+        result (str): The result of executing the workflow.
+    """
+    stdio_connection = MPCConnection(
+        command="npx", args=["-y", "tavily-mcp@0.1.4"], env={"TAVILY_API_KEY": "tvly-token"}
```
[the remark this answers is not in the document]
anastasiasenyk: for example purposes

dynamiq/nodes/agents/base.py
```diff
@@ -298,6 +299,8 @@ class Agent(Node):
     role: str | None = ""
     _prompt_blocks: dict[str, str] = PrivateAttr(default_factory=dict)
     _prompt_variables: dict[str, Any] = PrivateAttr(default_factory=dict)
+    _mcp_servers: list[Node] = PrivateAttr(default_factory=list)
```
acoola, line 302: it seems this would be more clear and should work
```suggestion
    _mcp_servers: list[MCPServerAdapter] = PrivateAttr(default_factory=list)
```

dynamiq/nodes/agents/base.py
```diff
@@ -298,6 +299,8 @@ class Agent(Node):
...
+    _exclude_tools_ids: list[str] = PrivateAttr(default_factory=list)
```
acoola, line 303: Let use here more clear naming, like `_mcp_servers_tools` or smth like this.

dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,258 @@
...
+      _mcp_tools (dict[str, MCPTool]): Internal dict of initialized MCP tools.
+    """
+
+    group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
+    name: str = "MCP Adapter Tool"
+    description: str = "The tool used to initialize available MCP tools based on provided server parameters."
+    connection: MPCConnection
+
+    tool_filter_names: list[str] = field(default_factory=list)
+    selection_mode: ToolSelectionMode = ToolSelectionMode.SELECT
```
acoola, line 188: let's gather both with the same prefix and maybe use `include/exclude` as we used such prefixes for other tools params
```suggestion
    tool_filter_mode: ToolSelectionMode = ToolSelectionMode.SELECT
```

dynamiq/connections/connections.py
```diff
@@ -1256,10 +1256,7 @@ class MPC(BaseConnection):
         default="strict", description="The text encoding error handler."
     )
 
-    def connect(self):
-        pass
-
-    async def get_client(self):
+    async def connect(self):
```
acoola: you don't have any async here, so just use `def connect` and after `async with self.connection.connect() as (read, write):`

dynamiq/nodes/agents/base.py
acoola: It seems my comment was unclear, _mcp_servers_tools it's about renaming `_exclude_tools_ids`. `mcp_servers` was OK
```suggestion
    _mcp_servers: list[MCPServerAdapter] = PrivateAttr(default_factory=list)
    _msp_servers_tools_ids: list[str] = PrivateAttr(default_factory=list)
```

dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,258 @@
...
+from dynamiq.connections import MCP as MPCConnection
+from dynamiq.nodes import NodeGroup
+from dynamiq.nodes.agents.exceptions import ToolExecutionException
+from dynamiq.nodes.node import ConnectionNode, ensure_config
+from dynamiq.runnables import RunnableConfig
+from dynamiq.utils import is_called_from_async_context
+from dynamiq.utils.logger import logger
+
+
+class ToolSelectionMode(Enum):
```
acoola, line 25:
```suggestion
class ToolFilterMode(Enum):
```

examples/components/core/dag/dag_mcp_server.yaml
```diff
@@ -0,0 +1,78 @@
+connections:
+  2195a94d-981c-46d5-b453-3759e3c25b1e:
```
acoola, line 2: let's in example use some human-readable string ids like `openai-conn`, `mcp-math-server-conn` and so on. The same for nodes/flows/etc. This would help read/verify node results/tracing, etc

dynamiq/connections/connections.py
```diff
@@ -1232,3 +1233,124 @@ def conn_params(self) -> dict:
             "api_base": self.url,
             "api_key": self.api_key,
         }
+
+
+class MCP(BaseConnection):
```
delamainer, line 1238: I would prefer to see two different connections for MCP. One for stdio and one for SSE.

dynamiq/nodes/agents/base.py
```diff
@@ -307,6 +310,18 @@ def __init__(self, **kwargs):
         self._intermediate_steps: dict[int, dict] = {}
         self._run_depends: list[dict] = []
         self._prompt = Prompt(messages=[])
+
+        expanded_tools = []
+        for tool in self.tools:
```
delamainer, line 315: I'm not really sure that we want to do this in `__init__`, as this will increase the cold start for the workflow. Because every time we initialize the workflow (and we do that on each invocation), we will have to make a call to the MCP server(s).

dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,258 @@
...
+                f"Tool '{self.name}' failed to call tool from the MCP server."
+                f"Error: {str(e)}. Please analyze the error and take appropriate action.",
+                recoverable=True,
+            )
+
+        logger.info(f"Tool {self.name} - {self.id}: finished with result:\n{str(result)[:200]}...")
+        return {"content": result}
+
+
+class MCPServerAdapter(ConnectionNode):
```
delamainer, line 168: What does `Adapter` mean here?

dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+    A tool that manages connections to MCP servers and initializes MCP tools.
+
+    Attributes:
+      group (Literal[NodeGroup.TOOLS]): Node group.
+      name (str): Node name.
+      description (str): Node description.
+      connection (MPCConnection): The connection module with parameters needed to the MCP server.
+      tool_filter_names (list[str]): Names of tools to include or exclude.
+      tool_filter_mode (ToolFilterMode): Strategy for tool filtering (INCLUDE or EXCLUDE).
```
delamainer, line 178: What about `include_tools` and `exclude_tools`? This seems simpler to use.

delamainer: - `len(include_tools) == 0 && len(exclude_tools) == 0` — all tools are available
- `len(include_tools) > 0 && len(exclude_tools) > 0` — all tools from `include_tools` excluding tools listed in `exclude_tools`

delamainer (changes requested):

dynamiq/connections/connections.py
acoola: Yes make sense for me too, talked about this

dynamiq/nodes/agents/base.py
acoola: Yes, this definitely makes sense for large workflows that could have few agents or complex structure. Unfortunately we don't have solution for now for some lazy resource/component access or load. Currently we check access to indexes, memory resources, and so on on init to prevent any potential issues with access and so on.
Anyway would be great to have option to define some resources/components for lazy access/load.

dynamiq/nodes/tools/mcp_adapter.py
```diff
@@ -0,0 +1,258 @@
...
+    @staticmethod
+    def get_input_schema(schema_dict) -> type[BaseModel]:
+        """
+        Creates an input schema based on provided MCP schema.
+
+        Args:
+            schema_dict (dict[str, Any]): A JSON schema dictionary describing the tool's expected input.
+        """
+        schema_dict = rename_keys_recursive(schema_dict, {"type_": "type"})
+        with TemporaryDirectory() as tmpdir:
```
acoola, line 91: Let's use memory resources for this (StringIO or similar things from io lib), as any manipulation with the file system is not a very good idea (of course, there are a lot of exceptions, but here don't see limitations).

dynamiq/nodes/tools/mcp_adapter.py
anastasiasenyk: I used the Adapter suffix to indicate that this class adapts an MCP server implementation (from the MCP library) to work within the Dynamiq library. I will remove this suffix to align it with other tools.

dynamiq/nodes/tools/mcp_adapter.py
anastasiasenyk: I think that's a good idea. I'll refactor it.

dynamiq/nodes/tools/mcp_adapter.py
anastasiasenyk: Unfortunately, it would no be possible since the `generate()` function from the `datamodel_code_generator` library expects an actual file path on the filesystem for the output. 
The alternative would be to manually parse the JSON schema, but this would require handling not only basic types but also nested/custom types.

dynamiq/nodes/agents/base.py
anastasiasenyk: Unfortunately, it is not possible to postpone this particular code snippet inside the agent's __init__ method.
Before executing this code, we don’t know which tools will be available inside the server. However, we need this information to create the agent's prompt with the correct tool descriptions. So, to delay execution, we would also have to delay the creation of the agent's prompt.

dynamiq/connections/connections.py
```diff
@@ -1232,3 +1234,64 @@ def conn_params(self) -> dict:
             "api_base": self.url,
             "api_key": self.api_key,
         }
+
+
+class MCPSse(BaseConnection):
+    url: str = Field(..., description="The SSE endpoint URL to connect to.")
+    headers: dict[str, Any] | None = Field(default=None, description="Optional headers to include in the SSE request.")
+    timeout: float = Field(default=5.0, description="Timeout in seconds for establishing the initial connection.")
+    sse_read_timeout: float = Field(default=60 * 5, description="Timeout for reading SSE messages (in seconds).")
```
delamainer, line 1243: Probably we don't need this `sse_` prefix.

delamainer (approved):

dynamiq/connections/connections.py
delamainer: Let's keep it as it is.

dynamiq/connections/connections.py
```diff
@@ -1244,3 +1246,64 @@ def conn_params(self) -> dict:
...
+    def connect(self):
+        """
+        Establishes an SSE connection.
+
+        Returns:
+            Async context manager for the SSE client.
+        """
+        from mcp.client.sse import sse_client
+
+        return sse_client(
```
delamainer, line 1266: We need to use `streamablehttp_client` from [here](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/streamable_http.py#L411).

dynamiq/connections/connections.py
```diff
@@ -1244,3 +1246,64 @@ def conn_params(self) -> dict:
             "api_base": self.url,
             "api_key": self.api_key,
         }
+
+
+class MCPSse(BaseConnection):
```
delamainer, line 1251:
```python3
class MCP(BaseConnection):
    ...
```

dynamiq/connections/connections.py
delamainer: SSE transport is being superseded by [Streamable HTTP transport](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http)!

delamainer (changes requested):

dynamiq/connections/connections.py
anastasiasenyk: Since older versions of SSE servers are not directly compatible with the Streamable HTTP client, I left MCPSse and added a new MCPStreamableHTTP connection.

dynamiq/connections/connections.py
anastasiasenyk: Since older versions of SSE servers are not directly compatible with the Streamable HTTP client, I left MCPSse and added a new MCPStreamableHTTP connection.

delamainer (dismissed):

acoola (approved):
Tremendous job

PR #251, opened 2025-04-24
feat: graph orchestrator parameters

maksymbuleshnyi:


dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -46,6 +47,7 @@ class GraphOrchestrator(Orchestrator):
     context: dict[str, Any] = {}
     states: list[GraphState] = []
     max_loops: int = 15
+    summarize_execution: bool = True
```
acoola, line 50: We already have `use_summarizer` in agents, so maybe we could reuse this parameter if it fit?
Would be great to limit the number of similar params with different names in different components.

maksymbuleshnyi: Decided to remove summarisation from Graph Orchestrator at all

dynamiq/nodes/agents/orchestrators/graph.py
```diff
@@ -38,6 +38,7 @@ class GraphOrchestrator(Orchestrator):
         initial_state (str): State to start from.
         objective (Optional[str]): The main objective of the orchestration.
         max_loops (Optional[int]): Maximum number of transition between states.
+        summarize_execution (Optional[bool]): Determines whether to call Manager to generate final output.
```
vitalii-dynamiq, line 41: @maksymbuleshnyi so are we removing it?

anastasiasenyk (approved):

acoola (approved):

PR #252, opened 2025-04-24
feat: update output configuration for non-streaming logic

TrachukT:
> [!IMPORTANT]
> Introduces `ResponseTransformer` enum to handle different response formats in `BaseLLM`, updating `_handle_completion_response()` to support detailed response metadata.
> 
>   - **Behavior**:
>     - Adds `ResponseTransformer` enum in `types.py` to specify response formats for LLM nodes.
>     - Updates `_handle_completion_response()` in `base.py` to use `ResponseTransformer` for response formatting.
>     - Supports `DEFAULT` and `CHAT_COMPLETION` formats, with `DEFAULT` returning plain text and `CHAT_COMPLETION` returning full response metadata.
>   - **Classes**:
>     - Adds `response_transformer` attribute to `BaseLLM` in `base.py` with default `ResponseTransformer.DEFAULT`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 61d0cd6e6c34c1d8a8a76ca117578813c27856bc. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/nodes/llms/base.py
+++ b/dynamiq/nodes/llms/base.py
@@ -79,6 +79,11 @@ class BaseLLM(ConnectionNode):
         - InferenceMode.STRUCTURED_OUTPUT: Produces structured JSON output.
         - InferenceMode.FUNCTION_CALLING: Structured output for tools (functions) to be called.
         dict[str, Any] | type[BaseModel] | None: schema_ for structured output. Defaults to empty dict.
+        response_transformer (ResponseTransformer): Specifies the format of the response returned by the node.
+        - ResponseTransformer.DEFAULT: Returns only the generated content (default), as a plain text string.
+        - ResponseTransformer.CHAT_COMPLETION: Returns the full response object in chat completion format, including
+            metadata such as token usage, role, and message structure.
+
     """
 
     MODEL_PREFIX: ClassVar[str | None] = None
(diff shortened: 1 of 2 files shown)


dynamiq/nodes/llms/base.py
```diff
@@ -79,6 +79,11 @@ class BaseLLM(ConnectionNode):
         - InferenceMode.STRUCTURED_OUTPUT: Produces structured JSON output.
         - InferenceMode.FUNCTION_CALLING: Structured output for tools (functions) to be called.
         dict[str, Any] | type[BaseModel] | None: schema_ for structured output. Defaults to empty dict.
+        response_transformer (ResponseTransformer): Specifies the format of the response returned by the node.
+        - ResponseTransformer.DEFAULT: Returns only the generated content (default), as a plain text string.
```
acoola, line 83: We don't need to add description about options and what they mean here. Please add description in Enum. The same for `inference_mode`

dynamiq/nodes/llms/base.py
```diff
@@ -102,6 +107,7 @@ class BaseLLM(ConnectionNode):
     schema_: dict[str, Any] | type[BaseModel] | None = Field(
         None, description="Schema for structured output or function calling.", alias="schema"
     )
+    response_transformer: ResponseTransformer = ResponseTransformer.DEFAULT
```
acoola, line 110: Let's make it Optional. And apply just when it passes. I think we also should remove the DEFAULT option as by default we return our structure (and DEFAULT it's not clear)

PR #291, opened 2025-05-20
feat: add Stagehand tool

anastasiasenyk:


dynamiq/nodes/tools/stagehand_tool.py
```diff
@@ -0,0 +1,220 @@
...
+                raise ToolExecutionException(f"Invalid action type: {input_data.action_type}", recoverable=True)
+
+        except Exception as e:
+            raise ToolExecutionException(f"Error message: {e}", recoverable=True)
+
+        logger.info(f"Tool {self.name} - {self.id}: finished with result:\n{str(result)[:200]}...")
+
+        return {"content": result}
+
+    def __del__(self):
```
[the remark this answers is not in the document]
acoola: Agree with this

pyproject.toml
```diff
@@ -58,6 +58,9 @@ elasticsearch = "~8.12.0"
 numpy = "~1.26.0"
 mcp = "1.8.0"
 datamodel-code-generator = "0.30.0"
+stagehand-py = "~0.3.6"
+browserbase = "~1.4.0"
+rich = "~14.0.0"
```
acoola, line 63: Are this libs not included in stagehand-py? If not, please add inline comments that they are mandatory for stagehand

dynamiq/nodes/tools/stagehand_tool.py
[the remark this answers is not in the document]
anastasiasenyk: My concern is whether users should be required to explicitly call the .close() method in all workflows. Should this be handled internally as part of the library’s logic?

dynamiq/nodes/tools/stagehand_tool.py
```diff
@@ -0,0 +1,189 @@
...
+    session: Stagehand | None = None
+
+    def start_session(self):
+        """
+        Initializes a Stagehand session.
+
+        Raises:
+            ToolExecutionException: If a required API key is missing or the model is unknown.
+        """
+        stagehand = self.connection.connect()
```
acoola, line 131: This is already done automatically by ConnectionNode functionality. Please recheck and just use `self.client`

dynamiq/nodes/tools/stagehand_tool.py
```diff
@@ -0,0 +1,189 @@
...
+    group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
+    name: str = "Stagehand Tool"
+    description: str = DESCRIPTION_STAGEHAND
+    connection: StagehandConnection
+    model_name: str
+
+    input_schema: ClassVar[type[StagehandInputSchema]] = StagehandInputSchema
+    session: Stagehand | None = None
+
+    def start_session(self):
```
acoola, line 124: It seems instead of `start_session` and `session` param need to override `init_components` method to additionally init client like:
```
    def init_components(self, connection_manager: ConnectionManager | None = None):
        """
        Initialize components for the node.

        Args:
            connection_manager (ConnectionManager, optional): The connection manager. Defaults to ConnectionManager.
        """
        connection_manager = connection_manager or ConnectionManager()
        super().init_components(connection_manager)
        self.client.model_name = self.model_name
        self.client.init()
```
After just use `self.client`

acoola (approved):
Great job

PR #294, opened 2025-05-25
feat: dry run

MaksymAI:
> [!IMPORTANT]
> Introduces a dry run feature with cleanup capabilities across nodes and vector stores, allowing simulation of operations without permanent changes.
> 
>   - **Behavior**:
>     - Introduces `DryRunConfig` in `types/dry_run.py` to configure dry run behavior, including options to delete documents and collections.
>     - Adds `_cleanup_dry_run()` and `_cleanup_dry_run_async()` methods in `flow.py` for cleaning up resources after dry runs.
>     - Implements `dry_run_cleanup()` in `node.py` and `operators.py` for node-level cleanup.
>   - **Vector Stores**:
>     - Integrates `DryRunMixin` into vector store classes like `ChromaVectorStore`, `ElasticsearchVectorStore`, `MilvusVectorStore`, `PGVectorStore`, `PineconeVectorStore`, `QdrantVectorStore`, and `WeaviateVectorStore` to support dry run operations.
>     - Adds tracking and cleanup of documents and collections in vector stores.
>   - **Examples**:
>     - Adds `dry_run_example.py` to demonstrate dry run functionality with `WeaviateDocumentWriter` and `Map` nodes.
>   - **Misc**:
>     - Updates `RunnableConfig` in `runnables/base.py` to include `dry_run` configuration.
>     - Adds `dry_run_cleanup()` to `Writer` class in `writers/base.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for c4f51f423ad01995c0997a95746f0abda130a71f. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/nodes/agents/react.py
+++ b/dynamiq/nodes/agents/react.py
@@ -1,6 +1,7 @@
 import json
 import re
 import types
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from enum import Enum
 from typing import Any, Callable, Union, get_args, get_origin
 
@@ -19,11 +20,7 @@ from dynamiq.nodes.agents.exceptions import (
     TagNotFoundError,
     XMLParsingError,
 )
-from dynamiq.nodes.agents.utils import (
-    SummarizationConfig,
-    ToolCacheEntry,
-    XMLParser,
-)
+from dynamiq.nodes.agents.utils import SummarizationConfig, ToolCacheEntry, XMLParser
 from dynamiq.nodes.llms.gemini import Gemini
 from dynamiq.nodes.node import Node, NodeDependency
 from dynamiq.nodes.types import Behavior, InferenceMode
@@ -1345,11 +1342,12 @@ class ReActAgent(Agent):
 
                 if action and self.tools:
                     tool_result = None
-                    tool_files = []
+                    tool_files: Any = []
                     tool = None
 
                     if self.inference_mode == InferenceMode.XML and self.parallel_tool_calls_enabled:
-                        tool_result = self._execute_tools(tools_data, config, **kwargs)
+                        execution_output = self._execute_tools(tools_data, config, **kwargs)
+                        tool_result, tool_files = self._separate_tool_result_and_files(execution_output)
 
                     elif self.inference_mode == InferenceMode.DEFAULT and self.parallel_tool_calls_enabled:
                         if action == "multiple_tools":
@@ -1379,7 +1377,8 @@ class ReActAgent(Agent):
                                 **kwargs,
                             )
 
-                            tool_result = self._execute_tools(tools_data, config, **kwargs)
+                            execution_output = self._execute_tools(tools_data, config, **kwargs)
+                            tool_result, tool_files = self._separate_tool_result_and_files(execution_output)
 
                             action_input_json = json.dumps(action_input)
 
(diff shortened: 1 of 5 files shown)


dynamiq/storages/vector/dry_run.py
```diff
@@ -0,0 +1,78 @@
...
+    Attributes:
+        cleanup_documents: If True, the ingested documents will be cleaned up after the dry run.
+        cleanup_collection: If True, the created collection will be cleaned up after the dry run.
+    """
+
+    cleanup_documents: bool = Field(default=True, description="Cleanup the ingested documents")
+    cleanup_collection: bool = Field(default=False, description="Cleanup the created collection")
+
+
+class DryRunMixin:
```
delamainer: Maybe it would be better to add a cleanup function to the base node?

dynamiq/storages/vector/dry_run.py
```diff
@@ -0,0 +1,78 @@
...
+    This class defines all parameters needed to configure dry run behavior
+    across different vector store implementations.
+
+    Attributes:
+        cleanup_documents: If True, the ingested documents will be cleaned up after the dry run.
+        cleanup_collection: If True, the created collection will be cleaned up after the dry run.
+    """
+
+    cleanup_documents: bool = Field(default=True, description="Cleanup the ingested documents")
+    cleanup_collection: bool = Field(default=False, description="Cleanup the created collection")
```
delamainer: It's better to use more explicit names:
- `delete_documents`
- `delete_collection`

Also, I don't like the fact that in some places we call such things an `index`, in others `collection`.

dynamiq/storages/vector/dry_run.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if dry_run_config.cleanup_documents and self._tracked_documents:
+            try:
+                self.delete_documents(list(self._tracked_documents))
+                logger.info(f"Cleaned up {len(self._tracked_documents)} tracked documents")
+                self._tracked_documents = []
+            except Exception as e:
+                logger.error(f"Failed to clean up tracked documents: {e}")
+
+        if dry_run_config.cleanup_collection and self._tracked_collection:
```
delamainer: I don't think deleting the collection is the right thing to do. It seems too dangerous. What if it had some data before? Instead, we should delete **only** inserted documents.

We might delete collection/table/etc only if the `create_if_not_exist` was set to `true` and the node itself created the collection/table/etc during its execution.

dynamiq/flows/flow.py
```diff
@@ -234,6 +266,14 @@ def run_sync(self, input_data: Any, config: RunnableConfig = None, **kwargs) ->
                     self._results.update(results)
                     self._ts.done(*results.keys())
 
+                    for node in ready_nodes:
+                        if (
+                            isinstance(node.node, BaseVectorStore)
+                            or isinstance(node.node, Writer)
+                            and node.node not in self._dry_run_nodes
+                        ):
+                            self._dry_run_nodes.append(node.node)
```
delamainer: Will this include nested nodes? For example `map` node has a nested node, what if the nested node is a vector store node?

delamainer (changes requested):

dynamiq/storages/vector/dry_run.py
delamainer: If the `node.dry_run.enabled` is `true`, call `node.dry_run_clean_up()` (or whatever the function name you like). Additionally, this function should be responsible for calling cleanup on its nested nodes.

For example, on a map node, when we call `the map.dry_run_clean_up()`, this method also calls `map.node.dry_run_clean_up()`.

dynamiq/runnables/base.py
```diff
@@ -30,6 +31,7 @@ class RunnableConfig(BaseModel):
     cache: CacheConfig | None = None
     max_node_workers: int | None = None
     nodes_override: dict[str, NodeRunnableConfig] = {}
+    dry_run_config: DryRunConfig | None = None
```
acoola: Please use similar style as for other keys (without config suffix)
```suggestion
    dry_run: DryRunConfig | None = None
```

dynamiq/storages/vector/dry_run.py
```diff
@@ -0,0 +1,78 @@
...
+    def dry_run_cleanup(self, dry_run_config: DryRunConfig) -> None:
+        """Clean up tracked resources based on configuration.
+
+        Args:
+            dry_run_config: Configuration for dry run behavior.
+        """
+
+        if dry_run_config.cleanup_documents and self._tracked_documents:
+            try:
+                self.delete_documents(list(self._tracked_documents))
```
acoola: I don't like that here you rely on some functionality that could be missed. To fix this:
- specify here `self.delete_documents` and `self.delete_collection` with potential pass/raise NotImplemented
- in child classed specify proper functionality (take care about order of parent classes in inheritance)

dynamiq/storages/vector/dry_run.py
acoola: please improve this functionality

dynamiq/flows/flow.py
```diff
@@ -299,6 +341,14 @@ async def run_async(self, input_data: Any, config: RunnableConfig = None, **kwar
...
                         self._results.update(results)
                         self._ts.done(*results.keys())
+
+                        for node in nodes_to_run:
+                            if (
+                                isinstance(node.node, BaseVectorStore)
+                                or isinstance(node.node, Writer)
+                                and node.node not in self._dry_run_nodes
+                            ):
+                                self._dry_run_nodes.append(node.node)
```
acoola: Already recommended by @delamainer - let's add the ability for each node to dry_run by self taking into account specific functionality, potentially other nodes as components, etc.

So in Flow you should just call `node.dry_run_cleanup` per each node after processing. Also need to take into account ability to use node as `node.run` without flow/workflow

We are specifying in the same way the `to_dict`, `clone` and other node related functionality.

dynamiq/storages/vector/dry_run.py
acoola: This mixing could be added to Node. After each node could specify own `dry_run_cleanup`. base functionality could be here

dynamiq/flows/flow.py
acoola: Let's also take into account that this is async version, so let's run it properly (asyncio.to_thread)

dynamiq/storages/vector/dry_run.py
tyaroshko: Updated the dry run config to use `enabled` flag and also added the mixin to the base Node to make sure we can explicitly define `dry_run_cleanup()` methods for each nodes.

dynamiq/storages/vector/dry_run.py
tyaroshko: Updated the naming for the `cleanup_documents` and `cleanup_collection` attribute.

dynamiq/storages/vector/dry_run.py
tyaroshko: This is mostly handled by the way we use `self._tracked_collection` in sub-classes – as we write to this attribute only if `create_if_not_exist` is set to `True` **and** there was no collection with such name beforehand. Therefore, it should be fine now, with a couple more updates I made lately.

dynamiq/flows/flow.py
tyaroshko: Added a separate definition of `dry_run_cleanup` for both `Map` node and `VectorStoreWriter` node.

dynamiq/storages/vector/dry_run.py
tyaroshko: Added proper definitions of missing methods in the base mixin class with `raise NotImplemented`.

dynamiq/flows/flow.py
tyaroshko: Changed the code structure to allow the explicit definitions of `dry_run_cleanup` for each sub-node in the workflow.

dynamiq/storages/vector/dry_run.py
```diff
@@ -0,0 +1,96 @@
+from pydantic import BaseModel, ConfigDict, Field
+
+from dynamiq.utils.logger import logger
+
+
+class DryRunConfig(BaseModel):
```
acoola: Let's move this config to more proper place as theoretically it not just about Vector stores (we have types, base and so on). Please check for proper place
Let's also move `class DryRunMixin` also in more proper way as it used by base Node

dynamiq/storages/vector/dry_run.py
```diff
@@ -0,0 +1,96 @@
...
+        self._tracked_collection: str | None = None
+
+    def delete_documents(self, document_ids: list[str] | None = None, delete_all: bool = False) -> None:
+        """Delete documents by their IDs.
+
+        Args:
+            document_ids: List of document IDs to delete.
+            delete_all: Whether to delete all documents.
+        """
+        raise NotImplementedError("Subclass must implement delete_documents method")
```
acoola: I think here we need to use a pass, as not all nodes should specify `delete_documents` logic. The same is for `def delete_collection`

dynamiq/storages/vector/dry_run.py
```diff
@@ -0,0 +1,96 @@
...
+        enabled (bool): Whether dry run is enabled. Defaults to False.
+        delete_documents: Whether to delete the ingested documents after the dry run. Defaults to True.
+        delete_collection: Whether to delete the created collection after the dry run. Defaults to False.
+    """
+
+    enabled: bool = False
+    delete_documents: bool = Field(default=True, description="Delete the ingested documents")
+    delete_collection: bool = Field(default=False, description="Delete the created collection")
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
acoola: We don't need here this config and used just bool params

acoola (commented):
Great jobs, small comments

acoola (dismissed):

acoola (approved):

PR #306, opened 2025-06-04
feat: change logic for tool passing

TrachukT:
> [!IMPORTANT]
> The PR updates the LLM tool handling logic to accept tools as either `Tool` instances or dictionaries, with changes in `BaseLLM`, `Prompt`, and an example script.
> 
>   - **Behavior**:
>     - `tools` attribute in `BaseLLM` and `Prompt` now accepts `list[Tool | dict]` instead of `list[Tool]`.
>     - `_get_response_format_and_tools()` in `base.py` updated to handle `Tool` instances and dictionaries.
>     - Example in `function_calling.py` updated to use dictionary format for tools.
>   - **Misc**:
>     - Removed unused imports `Tool`, `ToolFunction`, and `ToolFunctionParameters` from `function_calling.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 3722d91cf9ccd17db9fe0f3e85839f90aa0fdef5. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


dynamiq/nodes/llms/base.py
```diff
@@ -292,9 +292,8 @@ def _handle_streaming_completion_response(
 
     def _get_response_format_and_tools(
         self,
-        input_data: BaseLLMInputSchema,
         prompt: Prompt | None = None,
-        tools: list[Tool] | None = None,
+        tools: list[Tool | dict] | None = None,
```
acoola, line 296: Why we removed `dict` here?

dynamiq/nodes/llms/base.py
```diff
@@ -328,6 +327,9 @@ def _get_response_format_and_tools(
                         raise ValueError("Schema must be provided when using FUNCTION_CALLING inference mode")
                     tools = tools or schema
 
+        if tools:
+            tools = [tool.model_dump() if hasattr(tool, "model_dump") else tool for tool in tools]
```
acoola: Better to check with `isinstance` instead of `hasattr`

dynamiq/prompts/prompts.py
```diff
@@ -345,7 +345,4 @@ def format_messages(self, **kwargs) -> list[dict]:
         return out
 
     def format_tools(self, **kwargs) -> list[dict] | None:
-        out = None
-        if self.tools:
-            out = [tool.model_dump() for tool in self.tools]
-        return out
+        pass
```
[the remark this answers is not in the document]
acoola: This is a breaking change. Don't see the reason why this should be removed. This could be used by users or other services

PR #323, opened 2025-06-25
feat: add GeminiSTT node

TrachukT:


dynamiq/nodes/audio/gemini.py
```diff
@@ -0,0 +1,127 @@
...
+            messages=[
+                {
+                    "role": "user",
+                    "content": [
+                        {"type": "text", "text": "Generate a transcript of the speech."},
+                        {
+                            "type": "file",
+                            "file": {
+                                "file_data": f"data:{mime_type};base64,{encoded_data}",
+                            },
```
tyaroshko, line 123: Let's maybe move the prompt string `Generate a transcript of the speech`. to a module-level constant (something like `TRANSCRIPTION_PROMPT`) at the top of the file. We may also want to think about whether we want to eventually expose the transcription prompt to the user in the UI – then it would be a good idea to actually make it configurable in the node params.

anastasiasenyk: I agree. It would also be a good idea to allow users of the library to modify the prompt, and give them the ability to add more detailed instructions about specific language or domains. 
For example, to add a specific parameter to a prompt.

anastasiasenyk (approved):

PR #328, opened 2025-07-04
fix: replace typing.Self for Python 3.10 compatibility

anastasiasenyk:
> [!IMPORTANT]
> Replace `Self` with string literals for Python 3.10 compatibility and skip a test due to API key limitations.
> 
>   - **Compatibility**:
>     - Replace `Self` with string literals in `depends_on()` in `node.py` and `from_exception()` in `base.py` for Python 3.10 compatibility.
>   - **Testing**:
>     - Skip `test_react_agent_google_models` in `test_agent_llms.py` due to API key limitations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 42f3e5344afe89088d966c709c697d3bc7e3c2c6. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


dynamiq/nodes/node.py
```diff
@@ -6,11 +6,12 @@
 from datetime import datetime
 from functools import cached_property
 from queue import Empty
-from typing import Any, Callable, ClassVar, Self, Union
+from typing import Any, Callable, ClassVar, Union
 from uuid import uuid4
 
 from jinja2 import Template
 from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, computed_field, create_model, model_validator
+from typing_extensions import Self
```
acoola: `typing-extensions` is a 3pt lib that was missed in our toml file. Let's not add an extra dependency into toml and let's use like `-> "Node"` for Node class

PR #333, opened 2025-07-07
feat: update python node to pass multiple params in the code

TrachukT:
> [!IMPORTANT]
> Add `use_multiple_params` to `Python` node to allow input data as multiple parameters, with updated tests.
> 
>   - **Behavior**:
>     - Add `use_multiple_params` attribute to `Python` class in `python.py` to control input data passing method.
>     - If `use_multiple_params` is `True`, input data is unpacked into multiple parameters; otherwise, passed as a single dictionary.
>   - **Tests**:
>     - Update `test_python.py` to use `@pytest.mark.parametrize` for testing both single dictionary and multiple parameter passing.
>     - Add tests for various scenarios including math, random, and dynamiq imports with both input passing methods.
>     - Ensure security tests block unauthorized imports and executions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 315ee6ac06d86446337652f98f2ed61de8cdda53. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -7578,22 +7578,23 @@ optree = ["optree (>=0.13.0)"]
 
 [[package]]
 name = "tornado"
-version = "6.4.2"
+version = "6.5.1"
 description = "Tornado is a Python web framework and asynchronous networking library, originally developed at FriendFeed."
 optional = false
-python-versions = ">=3.8"
+python-versions = ">=3.9"
 files = [
-    {file = "tornado-6.4.2-cp38-abi3-macosx_10_9_universal2.whl", hash = "sha256:e828cce1123e9e44ae2a50a9de3055497ab1d0aeb440c5ac23064d9e44880da1"},
-    {file = "tornado-6.4.2-cp38-abi3-macosx_10_9_x86_64.whl", hash = "sha256:072ce12ada169c5b00b7d92a99ba089447ccc993ea2143c9ede887e0937aa803"},
-    {file = "tornado-6.4.2-cp38-abi3-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:1a017d239bd1bb0919f72af256a970624241f070496635784d9bf0db640d3fec"},
-    {file = "tornado-6.4.2-cp38-abi3-manylinux_2_5_i686.manylinux1_i686.manylinux_2_17_i686.manylinux2014_i686.whl", hash = "sha256:c36e62ce8f63409301537222faffcef7dfc5284f27eec227389f2ad11b09d946"},
-    {file = "tornado-6.4.2-cp38-abi3-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:bca9eb02196e789c9cb5c3c7c0f04fb447dc2adffd95265b2c7223a8a615ccbf"},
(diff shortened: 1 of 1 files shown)


tests/integration/nodes/tools/test_python.py
```diff
@@ -26,7 +26,7 @@ def run(input_data):
     }
 """
     python_node = Python(code=python_code)
-    input_data = {"name": "Alice", "age": 30}
+    input_data = {"input_data": {"name": "Alice", "age": 30}}
```
acoola: Why did you add this extra "input_data" key? Let's have test with existing data - as this is real case. I don't add extra key in my usage

dynamiq/nodes/tools/python.py
```diff
@@ -185,7 +185,7 @@ def guarded_write(obj, value=None):
             restricted_globals = compile_and_execute(self.code, restricted_globals)
             if "run" not in restricted_globals:
                 raise ValueError("The 'run' function is not defined in the provided code.")
-            result = restricted_globals["run"](dict(input_data))
+            result = restricted_globals["run"](**dict(input_data))
```
acoola: Previously, we had such a solution, but @vitalii-dynamiq pointed out that it was incorrect, as we can't connect nodes directly. 
https://github.com/dynamiq-ai/dynamiq/pull/19/files#diff-a2d6654c17e0f0ab8a471245196e617bef4d35535eac28e83f5a137db43831f9R151

Now, we will disconnect this direct connection, and we should be able to support connections via transformers? Or both?

tests/integration/nodes/tools/test_python.py
```diff
@@ -65,7 +65,7 @@ def run(input_data):
 def test_python_node_without_input():
     """Test Python node functionality without specific input data."""
     python_code = """
-def run(input_data):
+def run(input_data={}):
```
acoola: Let's skip using mutable as defaults

tests/integration/nodes/tools/test_python.py
```diff
@@ -15,7 +15,7 @@
 def test_python_node_with_input():
     """Test Python node with specific input data for name and age calculation."""
     python_code = """
-def run(input_data):
+def run(input_data:dict):
```
acoola:
```suggestion
def run(input_data: dict):
```

tests/integration/nodes/tools/test_python.py
```diff
@@ -51,7 +51,7 @@ def run(input_data):
     }
 """
     python_node = Python(code=python_code)
-    input_data = {}  # Empty input to trigger default values
+    input_data = {"input_data": {}}  # Empty input to trigger default values
```
acoola: Why did you change this? With new param it should work as before

tests/integration/nodes/tools/test_python.py
```diff
@@ -87,8 +87,7 @@ def test_python_node_with_math_import():
     python_code = """
 import math
 
-def run(input_data):
-    radius = input_data.get('radius', 1)
+def run(radius:int):
```
acoola:
```suggestion
def run(radius: int):
```

tests/integration/nodes/tools/test_python.py
```diff
@@ -98,7 +97,7 @@ def run(input_data):
         'pi_used': math.pi
     }
 """
-    python_node = Python(code=python_code)
+    python_node = Python(code=python_code, use_multiple_params=True)
```
acoola: Please add tests for both cases use_multiple_params=True/False, not just update existing

PR #350, opened 2025-07-25
feat: add validation for Pinecone index name

tyaroshko:
> [!IMPORTANT]
> Add validation for Pinecone index names in `pinecone.py` and corresponding tests in `test_pinecone.py`.
> 
>   - **Validation**:
>     - Add `validate_pinecone_index_name()` in `pinecone.py` to enforce index name pattern `^[a-z](?:[a-z0-9]*(-[a-z0-9]+)*)?$`.
>     - Apply validation in `PineconeVectorStore.__init__()` and `PineconeVectorStoreParams` via `field_validator`.
>   - **Testing**:
>     - Add `test_pinecone.py` with `test_valid_index_names()` and `test_invalid_index_names()` to verify index name validation logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 83b3362866c361dd090336935cbda97b8d06bf8d. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/storages/vector/pinecone/pinecone.py
+++ b/dynamiq/storages/vector/pinecone/pinecone.py
@@ -4,7 +4,7 @@ from copy import copy
 from functools import partial
 from typing import TYPE_CHECKING, Any, Optional
 
-from pydantic import Field
+from pydantic import Field, field_validator
 
 from dynamiq.connections import Pinecone
 from dynamiq.storages.vector.base import BaseVectorStore, BaseVectorStoreParams, BaseWriterVectorStoreParams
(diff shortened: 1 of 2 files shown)


dynamiq/storages/vector/pinecone/pinecone.py
```diff
@@ -83,7 +87,7 @@ def __init__(
                 connection = Pinecone()
             self.client = connection.connect()
 
-        self.index_name = index_name
+        self.index_name = self._fix_and_validate_index_name(index_name)
```
acoola: What about nodes and tracing? It seems we still show the old name. Is it OK to silently change index_name? How user should know about this?

tyaroshko: You're right, it would a good idea to at least introduce some kind of warning when we try to fix an invalid index name. Otherwise, I can remove the part when we fix the name and just try to throw the error on our side (because previously we did not handle the case where the index name is invalid, therefore, we were receiving the error message directly from Pinecone). 

Overall, we just had a [similar implementation for Weaviate](https://github.com/dynamiq-ai/dynamiq/blob/main/dynamiq/storages/vector/weaviate/weaviate.py#L82), and I thought it would make sense to also recreate this for other vector store nodes as well.

acoola (approved):

PR #375, opened 2025-08-20
feat: add intermediate streaming callback for ReAct agent

tyaroshko:
> [!IMPORTANT]
> Adds `ReActAgentStreamingParserCallback` for real-time LLM streaming in ReAct agent, supporting multiple inference modes and streaming steps.
> 
>   - **Behavior**:
>     - Adds `ReActAgentStreamingParserCallback` to `streaming.py` for parsing LLM streaming output in real-time.
>     - Supports `DEFAULT`, `XML`, `FUNCTION_CALLING`, and `STRUCTURED_OUTPUT` inference modes.
>     - Handles `reasoning` and `answer` steps, streaming them via `stream_content()`.
>   - **Agent Changes**:
>     - Integrates `ReActAgentStreamingParserCallback` in `ReActAgent` in `react.py`.
>     - Temporarily enables LLM streaming and attaches callback during execution.
>     - Removes callback post-execution and restores original streaming state.
>   - **Testing**:
>     - Adds `agents_test.py` for testing various LLM configurations, inference modes, and streaming modes.
>     - Uses Streamlit for UI to display test results and configurations.
>     - Includes `run_agent_async()` in `agent.py` for asynchronous agent execution.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 3de956a9806617d129c3f41ef08ce2b096c21744. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/nodes/agents/react.py
+++ b/dynamiq/nodes/agents/react.py
@@ -1091,10 +1091,19 @@ class ReActAgent(Agent):
                                 else:
                                     self.log_reasoning(thought, "multiple_tools", str(tools_data), loop_num)
 
+                                tools_data_for_streaming = [
+                                    {
+                                        "name": tool.get("name", ""),
+                                        "type": self.tool_by_names.get(tool.get("name", "")).type,
+                                    }
+                                    for tool in tools_data
+                                    if tool.get("name", "") and self.tool_by_names.get(tool.get("name", ""))
+                                ]
+
                                 self.stream_reasoning(
                                     {
                                         "thought": thought,
-                                        "tools": tools_data,
+                                        "tools": tools_data_for_streaming,
                                         "loop_num": loop_num,
                                     },
                                     config,
@@ -1216,7 +1225,7 @@ class ReActAgent(Agent):
                                     {
                                         "thought": thought,
                                         "action": action,
-                                        "tool": tool,
+                                        "tool": {"name": tool.name, "type": tool.type},
                                         "action_input": action_input,
                                         "loop_num": loop_num,
                                     },
@@ -1243,7 +1252,7 @@ class ReActAgent(Agent):
                                 {
                                     "thought": thought,
                                     "action": action,
-                                    "tool": tool,
+                                    "tool": {"name": tool.name, "type": tool.type},
                                     "action_input": action_input,
                                     "loop_num": loop_num,
                                 },


dynamiq/nodes/agents/react.py
```diff
@@ -1256,18 +1191,6 @@ def _run_agent(
                                         "Please correct the action field or state that you cannot answer the question. "
                                     )
 
-                                self.stream_reasoning(
```
maksymbuleshnyi: Tool and action seems not to be send in new implementation

maksymbuleshnyi: This will break intermediate streaming

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+            return
+
+        # Skip streaming if in FINAL mode and not in answer step
+        if self.agent.streaming.mode == StreamingMode.FINAL and step != StreamingState.ANSWER:
+            return
+
+        # Format content based on the step type
+        if step == "reasoning":
+            content_to_stream = {
+                "thought": content,
```
maksymbuleshnyi: Here also should be "tools", "action_input" and proper way to handle multiple tools.

dynamiq/callbacks/streaming.py
```diff
@@ -1,14 +1,62 @@
...
+class JSONStreamingField:
+    THOUGHT = "thought"
+    ACTION = "action"
+    ACTION_INPUT = "action_input"
+    ANSWER = "answer"
+
+
+class StreamingState:
+    REASONING = "reasoning"
+    ANSWER = "answer"
```
acoola: Please use str enums

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+class ReActAgentStreamingParserCallback(BaseCallbackHandler):
+    """ReAct Agent callback that parses LLM streaming output in real time and streams structured chunks.
+
+    This callback attaches to the underlying LLM node (group == 'llms'), incrementally parses the
+    provider streaming chunks to detect logical sections like reasoning and final answer based on
+    the selected inference mode, and forwards the relevant segments via the `stream_content()` API
+    with appropriate `step` labels (reasoning/answer) included.
+    """
+
+    def __init__(self, agent_instance: "ReActAgent", config, loop_num: int, **kwargs):
```
acoola:
```suggestion
    def __init__(self, agent: "ReActAgent", config, loop_num: int, **kwargs):
```

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+        self.loop_num = loop_num
+        self.kwargs = kwargs
+
+        # Aggregate streamed text from the LLM in the current loop for proper tracing inside the agent
+        self.accumulated_content: str = ""
+
+        self._buffer: str = ""
+        self._current_state: str | None = None
+        self._state_start_index: int = 0
+        self._last_emit_index: int = 0
```
acoola: is `self._state_start_index` or `self._last_emit_index` relate to self._current_state? Is yes please use the same prefix like `_current_state` or `_state` for all related

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+            field_name (str): The name of the field to search for.
+            start_index (int): The index to start searching from.
+
+        Returns:
+            int: The index of the first character inside the opening quote of the string field value,
+                 or -1 if not found.
+        """
+        key = f'"{field_name}"'
+        position = input_string.find(key, start_index)
+        if position == -1:
```
acoola, line 627: Please use `None` instead of `-1` if index is not found. And where it need check for `is None`

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+        """
+        if not content:
+            return
+
+        # Skip streaming if in FINAL mode and not in answer step
+        if self.agent.streaming.mode == StreamingMode.FINAL and step != StreamingState.ANSWER:
+            return
+
+        # Format content based on the step type
+        if step == "reasoning":
```
acoola: Please use enum value here instead of hardcode

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+            content_to_stream = {
+                "thought": content,
+                "loop_num": self.loop_num,
+            }
+        elif step == "answer":
```
acoola: Please use enum value here instead of hardcode

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+        if not content:
+            return
+
+        # Skip streaming if in FINAL mode and not in answer step
+        if self.agent.streaming.mode == StreamingMode.FINAL and step != StreamingState.ANSWER:
+            return
+
+        # Format content based on the step type
+        if step == "reasoning":
+            content_to_stream = {
```
acoola: It would be great to use Pydantic model here instead of just dict

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+        arguments_text = ""
+        function_name = None
+
+        choices = chunk.get("choices") or []
+        if choices:
+            delta = choices[0].get("delta", {})
+            tool_calls = delta.get("tool_calls")
+
+            if tool_calls and len(tool_calls) > 0:
+                tool_call = tool_calls[0]
```
acoola, line 410: Why we process just the first tool call? Or do we always have just 1?

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+        self.agent.stream_content(
+            content=content_to_stream,
+            source=self.agent.name,
+            step=step,
+            config=self.config,
+            **(self.kwargs | {"loop_num": self.loop_num}),
+        )
+
+    def _process_default_mode(self, final_answer_only: bool) -> None:
+        tags = {
```
acoola: I don't understand reason of this dict. You could just use `find(DefaultModeTag.THOUGHT)` and  also update this `for tag_text, name in ((tags["action"], "action"), (tags["answer"], "answer")):` to proper usage just DefaultModeTag enum

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+            "thought": DefaultModeTag.THOUGHT,
+            "action": DefaultModeTag.ACTION,
+            "answer": DefaultModeTag.ANSWER,
+        }
+
+        if self._current_state is None:
+            start = self._last_emit_index
+            idx_thought = self._buffer.find(tags["thought"], start) if not final_answer_only else -1
```
acoola: Please use None instead of -1 as it seems -1 is not a real position

dynamiq/callbacks/streaming.py
```diff
@@ -218,3 +266,521 @@ async def __aiter__(self) -> AsyncIterator[StreamingEventMessage]:
...
+        if safe_end > self._last_emit_index:
+            self._emit(self._buffer[self._last_emit_index : safe_end], step=StreamingState.ANSWER)
+            self._last_emit_index = safe_end
+
+    def _process_xml_mode(self, final_answer_only: bool) -> None:
+        open_thought = XMLModeTag.OPEN_THOUGHT
+        close_thought = XMLModeTag.CLOSE_THOUGHT
+        open_answer = XMLModeTag.OPEN_ANSWER
+        close_answer = XMLModeTag.CLOSE_ANSWER
+        open_action = XMLModeTag.OPEN_ACTION
```
acoola: Just use Enum values instead if this extra variables

dynamiq/nodes/agents/react.py
```diff
@@ -924,50 +923,75 @@ def _run_agent(
...
-                    response_format=self._response_format,
-                    config=config,
-                    **kwargs,
-                )
+                # Create a streaming callback
+                streaming_callback = None
+                original_streaming_enabled = self.llm.streaming.enabled
+
+                if self.streaming.enabled:
+                    streaming_callback = ReActAgentStreamingParserCallback(
```
acoola, line 931: As far as I understand, we are always using this callback underhood and shouldn't specify in workflow run config?

dynamiq/callbacks/streaming.py
tyaroshko: We have parallel tool calls enabled only for DEFAULT and XML inference modes. In FUNCTION_CALLING mode (where we use this `_extract_function_calling_text()` method), there would be only one tool call on each step –  either an actual tool call or a `provide_final_answer()` function.

dynamiq/callbacks/streaming.py
tyaroshko: Does it make sense to do so if `find()` will return -1 anyway whenever it doesn't find the string we were searching for?

dynamiq/nodes/agents/react.py
tyaroshko: Yes, we want to use this callback by default whenever streaming is enabled.

acoola (approved):
Great job

PR #377, opened 2025-08-20
feat: update error handling for Zenrows and ScaleSerp

TrachukT:


dynamiq/nodes/tools/zenrows.py
```diff
@@ -77,21 +77,20 @@ def execute(self, input_data: ZenRowsInputSchema, config: RunnableConfig = None,
...
-            response.raise_for_status()
-            scrape_result = response.text
-        except Exception as e:
-            logger.error(f"Tool {self.name} - {self.id}: failed to get results. Error: {e}")
+        response = self.client.request(
+            method=self.connection.method,
+            url=self.connection.url,
+            params={**self.connection.params, **params},
+        )
+        if response.status_code >= 400:
```
acoola: There could be other unexpected errors - connections, timeouts, etc. So make sense them also handle with except Exception as e. But for 4xx errors (maybe 3xx, 5xx) make sense to use our error message

PR #393, opened 2025-09-09
feat: agents as tools

vitalii-dynamiq:


olbychos:
I had more slighter implementation and tested with yaml/tracing, propose to look over that PR and merge it, in a bit simpler way.

PR #436, opened 2025-10-10
fix: update action parsing pattern in Agent

tyaroshko:
> [!IMPORTANT]
> Update action parsing in `Agent` to use `regex` for handling nested JSON structures and add `regex` as a dependency.
> 
>   - **Behavior**:
>     - Update action parsing pattern in `_parse_action()` in `agent.py` to handle nested JSON structures using `regex`.
>     - Replace `re.search` with `regex.search` for action pattern matching in `agent.py`.
>   - **Dependencies**:
>     - Add `regex` library to `pyproject.toml` for advanced regex capabilities.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for ea77c7c54e5b8f6a3ce439702b6adc937ba87e05. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pyproject.toml
```diff
@@ -63,6 +63,7 @@ databricks-sql-connector = "~4.0.3"
 orjson = "3.10.18"
 charset-normalizer = "^3.4.2"
 tenacity = "9.1.2"
+regex = "^2025.9.18"
```
acoola, line 66: why we need add extra pkg?

dynamiq/nodes/agents/agent.py
```diff
@@ -798,13 +799,13 @@ def _parse_action(self, output: str) -> tuple[str | None, str | None, dict | lis
             thought_match = re.search(thought_pattern, output, re.DOTALL)
             thought = thought_match.group(1).strip() if thought_match else None
 
-            action_pattern = r"Action:\s*(.*?)\nAction Input:\s*((?:[\[{][\s\S]*?[\]}]))"
+            action_pattern = r"Action:\s*(.*?)\nAction Input:\s*(\{(?:[^{}]|(?R))*\})"
```
olbychos, line 802: What about the XML parser - does it parse things correctly?

tyaroshko: I didn't find any issues with it while testing agents, so I think XML parser should be fine.

PR #442, opened 2025-10-14
feat: add agent file processing

maksymbuleshnyi:
> [!IMPORTANT]
> Enhance `FileReadTool` with intelligent file type detection and content extraction using converters and LLMs for various file types.
> 
>   - **Behavior**:
>     - `FileReadTool` in `file_tools.py` now detects file types and uses converters for text extraction.
>     - Supports PDF, DOCX, PPTX, HTML, TXT, IMAGE, and more.
>     - For large files, returns chunked content.
>     - Uses LLM for image and PDF processing if supported.
>   - **Converters**:
>     - Adds `detect_file_type()` function in `file_tools.py` for file type detection.
>     - Introduces `DEFAULT_FILE_TYPE_TO_CONVERTER_CLASS_MAP` for mapping file types to converters.
>     - `LLMImageConverter` and `LLMPDFConverter` in `llm_text_extractor.py` handle image and PDF text extraction.
>   - **File Types**:
>     - New `FileType` enum and `EXTENSION_MAP` in `file_types.py` for file type management.
>   - **Tests**:
>     - Updates `test_file_tools.py` to test new file reading and writing functionalities.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for eaeafec829ba8aa8fb5b330177075ec5d6b7e88f. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/flows/flow.py
+++ b/dynamiq/flows/flow.py
@@ -268,14 +268,13 @@ class Flow(BaseFlow):
 
                 while self._ts.is_active():
                     ready_nodes = self._get_nodes_ready_to_run(input_data=input_data)
-                    if ready_nodes:
-                        results = run_executor.execute(
-                            ready_nodes=ready_nodes,
-                            config=config,
-                            **(merged_kwargs | {"parent_run_id": run_id}),
-                        )
-                        self._results.update(results)
-                        self._ts.done(*results.keys())
+                    results = run_executor.execute(
+                        ready_nodes=ready_nodes,
+                        config=config,
+                        **(merged_kwargs | {"parent_run_id": run_id}),
+                    )
+                    self._results.update(results)
+                    self._ts.done(*results.keys())
 
                     # Wait for ready nodes to be processed and reduce CPU usage
                     time.sleep(0.003)
--- a/dynamiq/nodes/node.py
+++ b/dynamiq/nodes/node.py
@@ -1006,7 +1006,7 @@ class Node(BaseModel, Runnable, DryRunMixin, ABC):
             raise error
         finally:
             if executor is not None:
-                executor.shutdown(wait=False)
+                executor.shutdown()
 
     def execute_with_timeout(
         self,
--- a/tests/integration/nodes/tools/test_mcp_tool.py
+++ b/tests/integration/nodes/tools/test_mcp_tool.py
@@ -120,7 +120,7 @@ def mcp_server_tool(sse_server_connection, mock_mcp_tools):
 @pytest.mark.asyncio
 async def test_get_mcp_tools(mcp_server_tool, sse_server_connection):
     with patch.object(MCPServer, "initialize_tools", new=AsyncMock()) as mock_init:
-        tools = await mcp_server_tool.get_mcp_tools_async()
+        tools = mcp_server_tool.get_mcp_tools()
         mock_init.assert_not_awaited()
 
         expected_tools = [


dynamiq/nodes/tools/file_tools.py
```diff
@@ -1,23 +1,78 @@
...
+
+
+DEFAULT_FILE_TYPE_TO_CONVERTER_CLASS_MAP = {
+    FileType.PDF: PyPDFConverter,
+    FileType.DOCUMENT: DOCXFileConverter,
+    FileType.PRESENTATION: PPTXFileConverter,
+    FileType.HTML: HTMLConverter,
+    FileType.TEXT: TextFileConverter,
+    FileType.MARKDOWN: TextFileConverter,
+    FileType.IMAGE: LLMImageConverter,
```
vitalii-dynamiq, line 74: which LLM it will use?

dynamiq/nodes/tools/file_tools.py
```diff
@@ -1,23 +1,78 @@
...
+        return None
+
+    except Exception as e:
+        logger.warning(f"File type detection failed for {filename}: {str(e)}")
+        return None
+
+
+DEFAULT_FILE_TYPE_TO_CONVERTER_CLASS_MAP = {
+    FileType.PDF: PyPDFConverter,
+    FileType.DOCUMENT: DOCXFileConverter,
```
vitalii-dynamiq: document is too broad, no? same for presentation. they might be in other formats

dynamiq/nodes/tools/file_tools.py
```diff
@@ -41,34 +96,40 @@ class FileWriteInputSchema(BaseModel):
 
 class FileReadTool(Node):
     """
-    A tool for reading files from storage.
+    A tool for reading files from storage with intelligent file processing.
 
-    This tool can be passed to ReAct agents to read files
-    from the configured storage backend. For large files, it automatically
-    returns chunked content showing first, middle, and last parts.
+    This tool can be passed to ReAct agents to read files from the configured storage backend.
```
vitalii-dynamiq: let's not use ReAct anymore

dynamiq/nodes/tools/file_tools.py
```diff
@@ -41,34 +96,40 @@ class FileWriteInputSchema(BaseModel):
...
 
     Attributes:
         group (Literal[NodeGroup.TOOLS]): The group to which this tool belongs.
         name (str): The name of the tool.
         description (str): A brief description of the tool.
         file_store (FileStore): File storage to read from.
+        llm (BaseLLM): LLM that will be used to process files.
         max_size (int): Maximum size in bytes before chunking (default: 10000).
         chunk_size (int): Size of each chunk in bytes (default: 1000).
+        converter_mapping (dict[FileType, Node]): Mapping of file types to converters.
```
vitalii-dynamiq, line 114: how do we deal with cases when an LLM itself can process pptx, docx, and many other formats

vitalii-dynamiq (approved):

dynamiq/nodes/tools/file_tools.py
maksymbuleshnyi: The same as for agent

dynamiq/nodes/tools/file_tools.py
maksymbuleshnyi: We do not use llms in this implementation in such cases. if LLM can process images on contrary we do use our LLM image converter.

dynamiq/nodes/tools/file_tools.py
maksymbuleshnyi: Agree, changed

dynamiq/nodes/tools/file_tools.py
maksymbuleshnyi: Currently as I see there are problems with detecting if model in litellm supports specific file format, except for images and pdfs.

vitalii-dynamiq:
@olbychos @tyaroshko @delamainer please review

tyaroshko (approved):
LGTM 

(but we'll probably need to update this PR when we make `dynamiq-extra` public in order to avoid code duplication for file type mapping and detection)

dynamiq/nodes/tools/file_tools.py
```diff
@@ -1,23 +1,78 @@
...
+    """
+    try:
+        if not filename and hasattr(file, "name"):
+            filename = file.name
+
+        if not filename:
+            logger.warning("No filename provided for file type detection")
+            return None
+
+        file_ext = os.path.splitext(filename)[1][1:] if filename else ""
```
delamainer, line 47: Why do we use extensions instead of MIME types?

maksymbuleshnyi:
Image 20 10 2025 at 17 40


Example of trace

dynamiq/nodes/tools/file_tools.py
maksymbuleshnyi: I used similar implementation to what we have in MultifileConverter. Potentially we can extend to using MIME types too.

dynamiq/nodes/tools/file_tools.py
```diff
@@ -1,23 +1,78 @@
...
+
+        logger.warning(f"Unknown file extension: {file_ext} for file: {filename}")
+        return None
+
+    except Exception as e:
+        logger.warning(f"File type detection failed for {filename}: {str(e)}")
+        return None
+
+
+DEFAULT_FILE_TYPE_TO_CONVERTER_CLASS_MAP = {
```
olbychos, line 67: How about csv/excel files?

olbychos (dismissed):
Overall, lgtm; any other important file types worth adding (like excel,csv)?

dynamiq/nodes/tools/file_tools.py
maksymbuleshnyi: in this case when nothing will be found it will fallback to pure content, which works fine for csv.

maksymbuleshnyi:
> Overall, lgtm; any other important file types worth adding (like excel,csv)?

This is generally supported because we will fallback to pure content, which works well with csv

dynamiq/nodes/tools/file_tools.py
```diff
@@ -41,86 +96,166 @@ class FileWriteInputSchema(BaseModel):
...
+
+                if result.status == RunnableStatus.SUCCESS:
+                    documents = result.output.get("documents", [])
+                    if documents:
+                        text_content = "\n\n".join([doc.content for doc in documents if hasattr(doc, "content")])
+                        logger.info(f"Successfully extracted text using {converter_name}")
+                        return text_content
+                    else:
+                        logger.warning(f"No documents extracted by {converter_name}")
+                        return ""
```
acoola: Why do we return an empty string here? Maybe the proper way is to raise some exception or return None? The same for other returns in this method

dynamiq/nodes/tools/file_tools.py
```diff
@@ -167,22 +310,51 @@ def execute(
...
-                if is_image and self.llm.is_vision_supported:
-                    logger.info(f"Tool {self.name} - {self.id}: processing image with LLM (vision supported)")
-                    return self._process_with_llm(content, input_data.instructions, config, **kwargs)
-                elif is_image and not self.llm.is_vision_supported:
-                    error_msg = (
-                        f"Cannot process image file '{input_data.file_path}' with current LLM. "
-                        f"The model '{self.llm.model}' does not support vision/image processing."
+            try:
+                file_io = BytesIO(content)
+                filename = input_data.file_path.split("/")[-1]  # Extract filename from path
```
acoola: Please use os.path for this to support not just UNIX like path

dynamiq/nodes/tools/file_tools.py
```diff
@@ -242,12 +414,51 @@ def _create_chunked_content(self, content: bytes, chunk_size: int, file_path: st
 
         return chunked_bytes
 
+    def _create_chunked_text_content(self, content: str, chunk_size: int, file_path: str) -> str:
```
acoola, line 411: Chunked content could looks weird in case if small value of `self.max_size`. Not sure if we need to add some checks for this, and if this affect some further Agent processing. If affect - make sense to add, if not - just keep in mind

maksymbuleshnyi: It solves a problem when we are reading big txt/csv file.

PR #452, opened 2025-10-28
feat: add tool auth

olbychos:


examples/components/auth/agent_with_exa_auth.py
```diff
@@ -0,0 +1,110 @@
...
+
+def run_with_credentials(workflow: Workflow, agent: Agent, api_key: str):
+    """
+    Re-run the agent with `tool_auth` populated by an AuthConfig object.
+    The tool now receives the key via headers and can complete the search.
+    """
+    auth_config = AuthConfig(
+        scheme=AuthScheme(
+            type=AuthSchemeType.API_KEY,
+            name="x-api-key",
```
vitalii-dynamiq, line 76: @olbychos would this work with our tools natively?

olbychos: We''ll have to update each of our tool in same way I've done with exa

olbychos:
@delamainer  / @acoola  / @maksymbuleshnyi please review the conceptual thing

dynamiq/auth/README.md
```diff
@@ -0,0 +1,91 @@
+# Tool Authentication
+
+This document summarizes the current authentication capabilities in Dynamiq, how agents surface missing credentials, and how to resume execution once credentials are supplied.
+
+## Architecture Overview
+
+```
+User / Client  <–– dynamiq_request_credential event ––  Agent  –– ToolExecutionContext –– Tool
```
acoola, line 8: This is not clear in context of start and end in this flow (or what is it)

dynamiq/auth/README.md
```diff
@@ -0,0 +1,91 @@
...
+## Runtime Flow
+
+1. Agent runs normally until a tool needs credentials.
+2. Tool calls `tool_context.request_auth(auth_config)`.
+   - Agent raises `ToolAuthRequiredException`.
+   - Streaming clients receive an event with `event: "dynamiq_request_credential"`.
+   - Callbacks receive the `AuthRequest` via `on_node_auth_request`.
+   - Agent output contains `status: "auth_required"` and serialized `auth_requests`.
+3. Client prompts the user and collects credentials (API key, OAuth redirect URL, etc.).
+4. Client re-runs the workflow/agent, injecting credentials via `tool_auth` or a stored credential cache.
```
acoola, line 43: Should we rerun, or we can resume and make retry of last action?

dynamiq/auth/models.py
```diff
@@ -0,0 +1,139 @@
+from datetime import datetime
```
acoola, line 1: we don't use models.py in lib, make sense to use another naming (types, etc)

dynamiq/auth/manager.py
```diff
@@ -0,0 +1,51 @@
+from typing import Any
```
acoola, line 1: `managers.py` instead of `manager.py`

dynamiq/auth/models.py
```diff
@@ -0,0 +1,139 @@
...
+    description: str | None = None
+    location: AuthSchemeLocation | None = Field(default=None, alias="in")
+    bearer_format: str | None = None
+    token_url: str | None = None
+    scopes: dict[str, str] = Field(default_factory=dict)
+
+    model_config = {"populate_by_name": True}
+
+
+class AuthCredential(BaseModel):
```
acoola, line 48: I think each type should have a properly defined structure, not smth general with all potencial filelds.
After there should be union of all of them with discriminator usage to choose proper model by `type`
https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions

dynamiq/auth/models.py
```diff
@@ -0,0 +1,139 @@
...
+            "headers": {},
+            "query": {},
+            "cookies": {},
+            "metadata": self.metadata.copy(),
+        }
+
+        scheme = self.scheme
+        credential = self.credential
+
+        if scheme.type == AuthSchemeType.API_KEY:
```
acoola, line 92: Each scheme should by self be able to create own partial payload

dynamiq/auth/models.py
```diff
@@ -0,0 +1,139 @@
...
+        return payload
+
+
+class AuthRequest(BaseModel):
+    """
+    Represents an authentication request emitted by a tool.
+    """
+
+    id: str = Field(default_factory=lambda: str(uuid4()))
+    tool_id: str | None = None
```
acoola, line 134: Let's try to avoid using `tool` and be more general, using something like `node`.

dynamiq/nodes/agents/base.py
```diff
@@ -289,6 +296,74 @@ class ToolParams(BaseModel):
     by_id_params: dict[str, Union[dict[str, Any], "ToolParams"]] = Field(default_factory=dict, alias="by_id")
 
 
+class ToolAuthConfig(BaseModel):
+    global_auth: AuthConfig | dict[str, Any] = Field(default_factory=dict, alias="global")
```
acoola, line 300: What is mean `global`?

dynamiq/nodes/agents/base.py
```diff
@@ -289,6 +296,74 @@ class ToolParams(BaseModel):
...
+    by_name_auth: dict[str, Union[dict[str, Any], "ToolAuthConfig", AuthConfig]] = Field(
+        default_factory=dict, alias="by_name"
+    )
+    by_id_auth: dict[str, Union[dict[str, Any], "ToolAuthConfig", AuthConfig]] = Field(
+        default_factory=dict, alias="by_id"
+    )
```
acoola, line 306: - Let's call `auth_by_id`, `auth_by_name`, etc - as it seems more clear and readable
- such options not clear for me. What they mean?

dynamiq/nodes/tools/context.py
```diff
@@ -0,0 +1,62 @@
+from typing import Any, Callable
+
+from dynamiq.auth import AuthConfig, AuthRequest
+
+
+class ToolExecutionContext:
```
acoola, line 6: Let's use Pydantic model here too.

dynamiq/nodes/tools/exa_search.py
```diff
@@ -214,6 +216,39 @@ def execute(self, input_data: ExaInputSchema, config: RunnableConfig | None = No
         config = ensure_config(config)
         self.run_on_node_execute_run(config.callbacks, **kwargs)
 
+        tool_context = kwargs.get("tool_context")
```
acoola, line 219: This should be extracted to a separate method. And I think this method should have the same name. Maybe make sense to define in base class for tools/nodes and minimize duplication and rewriting the same things in each tool

PR #458, opened 2025-10-31
feat: add logic to return files instance as Elevenlabs output

TrachukT:
> [!IMPORTANT]
> Add `BytesIO` file object to ElevenLabs audio generation output and update tests accordingly.
> 
>   - **Behavior**:
>     - `execute()` in `ElevenLabsTTS` and `ElevenLabsSTS` now returns a dictionary with `content` (bytes) and `files` (list of `BytesIO`).
>     - `files` contains a single `BytesIO` object named `audio.mp3` with content type `audio/mpeg`.
>   - **Tests**:
>     - Updated `test_workflow_with_elevenlabstts` and `test_workflow_with_elevenlabssts` to check for `files` key in output.
>     - Added assertions for `BytesIO` object properties in both tests.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for 3d4109c67df13b10dca0230a6c2bbff7d1d38030. You can [customize](https://app.ellipsis.dev/dynamiq-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dynamiq/nodes/audio/elevenlabs.py
+++ b/dynamiq/nodes/audio/elevenlabs.py
@@ -99,7 +99,7 @@ class ElevenLabsTTS(ConnectionNode):
 
     def execute(
         self, input_data: ElevenLabsTTSInputSchema, config: RunnableConfig = None, **kwargs
-    ) -> dict[str, bytes | io.BytesIO]:
+    ) -> dict[str, bytes | list[io.BytesIO]]:
         """Execute the audio generation process.
 
         This method takes input data and returns the result.
@@ -196,7 +196,7 @@ class ElevenLabsSTS(ConnectionNode):
 
     def execute(
         self, input_data: ElevenLabsSTSInputSchema, config: RunnableConfig = None, **kwargs
-    ) -> dict[str, bytes | io.BytesIO]:
+    ) -> dict[str, bytes | list[io.BytesIO]]:
         """Execute the audio generation process.
 
         This method takes input data and returns the result.


dynamiq/nodes/audio/elevenlabs.py
```diff
@@ -133,8 +134,15 @@ def execute(
...
+
+        audio_bytes = response.content
+        audio_file = io.BytesIO(audio_bytes)
+        audio_file.name = "audio.mp3"
+        audio_file.content_type = "audio/mpeg"
+
         return {
-            "content": response.content,
+            "content": audio_bytes,
+            "files": [audio_file],
```
delamainer, line 145: Why do we return a list of files if we only have one?

delamainer: I don't even think we should add one more key (`files`). The content should be `io.BytesIO`. There is no point in having both `content` and `files`.

dynamiq/nodes/audio/elevenlabs.py
```diff
@@ -226,6 +235,13 @@ def execute(
...
```
delamainer, line 246: Same here.

delamainer (changes requested):

vitalii-dynamiq:
@TrachukT what's the main reason for this PR?

acoola (approved):

TrachukT:
> @TrachukT what's the main reason for this PR?

Unification of all tools that`s working with files. We already have file output logic, make sense to return already configured file as response for all such nodes

PR #522, opened 2026-01-14
fix: dry run for VectorStoreWriter node

tyaroshko:
> [!NOTE]
> Improves dry-run cleanup behavior and visibility.
> 
> - VectorStoreWriter now exposes `dry_run_cleanup` that delegates to its `document_writer`
> - Flow `_cleanup_dry_run` and `_cleanup_dry_run_async` now:
>   - Use debug-level start logs and list nodes that implement a cleanup override
>   - Filter to nodes that override `dry_run_cleanup` (skip base `Node.dry_run_cleanup`)
>   - Run async cleanup via `asyncio.to_thread` and report per-node errors
> - DryRunMixin elevates successful cleanup logs from debug to info for documents and collections
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 55f62914c1071a87876a670c1220b42fe6cf34f9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/dynamiq/storages/vector/pgvector/pgvector.py
+++ b/dynamiq/storages/vector/pgvector/pgvector.py
@@ -16,6 +16,7 @@ from dynamiq.nodes.dry_run import DryRunMixin
 from dynamiq.storages.vector.base import BaseVectorStore, BaseVectorStoreParams, BaseWriterVectorStoreParams
 from dynamiq.storages.vector.exceptions import VectorStoreException
 from dynamiq.storages.vector.pgvector.filters import _convert_filters_to_query
+from dynamiq.storages.vector.utils import create_pgvector_file_id_filter, create_pgvector_file_ids_filter
 from dynamiq.types import Document
 from dynamiq.types.dry_run import DryRunConfig
 from dynamiq.utils.logger import logger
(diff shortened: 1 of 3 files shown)


dynamiq/flows/flow.py
```diff
@@ -232,11 +232,20 @@ def _cleanup_dry_run(self, config: RunnableConfig = None):
         if not config or not getattr(config.dry_run, "enabled", False):
             return
 
-        logger.info("Starting dry run cleanup...")
-        for node in self.nodes:
+        logger.debug("Starting dry run cleanup...")
+
+        # Filter nodes that have dry_run_cleanup implemented
+        nodes_with_cleanup = [
```
acoola, line 238: Why do we need this if DryRunMixin with `dry_run_cleanup` is present in each node?

tyaroshko: Yes, we do have the `dry_run_cleanup` in each node, that's why I updated the logs to make sure we only include the nodes that override this method in their implementation (currently these are vector store, vector store writer, writer and map nodes). Previously, you would see all nodes of the worfklow there which I thought doesn't really make sense.

acoola (approved):

maksymbuleshnyi (approved):

PR #533, opened 2026-01-23
fix: agent node creation

mykhailobuleshnyi:
Fixed node generation to support multiple edges in Agent node.

### DIFF
--- a/dynamiq/utils/workflow_generation/node_generation.py
+++ b/dynamiq/utils/workflow_generation/node_generation.py
@@ -209,7 +209,7 @@ def generate_node(node_cls: type[Node], node_info: dict[str, Any], taken_names: 
             validate_input_transformer(node.prompt.messages, node)
 
         elif isinstance(node, Agent):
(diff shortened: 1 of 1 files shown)


olbychos (dismissed):

dynamiq/utils/workflow_generation/node_generation.py
```diff
@@ -209,7 +209,7 @@ def generate_node(node_cls: type[Node], node_info: dict[str, Any], taken_names:
             validate_input_transformer(node.prompt.messages, node)
 
         elif isinstance(node, Agent):
-            validate_input_transformer([Message(role=MessageRole.USER, content="{{input}}")], node)
+            validate_input_transformer([Message(role=MessageRole.USER, content="{{input}} " + node.role)], node)
```
acoola: Please use here f-string formatting as this is a preferable way and faster

vitalii-dynamiq: @mykhailobuleshnyi please fix it

PR #537, opened 2026-01-29
fix: Rework parallel tool calling

mykhailobuleshnyi:


dynamiq/nodes/tools/parallel_tool_calls.py
```diff
@@ -0,0 +1,242 @@
...
+    Example:
+        {
+            "tools": [
+                {"name": "search_tool", "input": {"query": "AI news"}},
+                {"name": "weather_tool", "input": {"city": "NYC"}}
+            ]
+        }
+    """
+
+    tools: list[dict[str, Any]] = Field(
```
acoola: Can we also define the tool schema (name, input, etc) instead of docstring with example? dict is so general.

dynamiq/nodes/tools/parallel_tool_calls.py
```diff
@@ -0,0 +1,242 @@
...
+Example - search two topics at once:
+{
+    "tools": [
+        {"name": "search", "input": {"query": "Python best practices"}},
+        {"name": "search", "input": {"query": "Python performance tips"}}
+    ]
+}"""
+    )
+
+    error_handling: ErrorHandling = Field(
```
acoola: This is redundant for this tool as it doesn't make execution (inherit from base Node is enought)

dynamiq/nodes/agents/agent.py
```diff
@@ -758,9 +710,8 @@ def _execute_tools_and_update_prompt(
                 )
 
                 # Handle XML parallel mode (but not for ContextManagerTool)
-                if self.inference_mode == InferenceMode.XML and self.parallel_tool_calls_enabled and not skip_parallel:
-                    tools_data = action_input if isinstance(action_input, list) else [action_input]
-                    execution_output = self._execute_tools(tools_data, thought, loop_num, config, **kwargs)
+                if action == "<Parallel Tool Calls>" and self.parallel_tool_calls_enabled and not skip_parallel:
```
acoola: - What is this magic `<Parallel Tool Calls>` string? I this is name if Tool, please define it as constant and reuse it here and in tool naming.
- Make sure that the user can't specify their own naming for the Parallel Tool (but it seems we always add our own parallel tool)
- Please check the case if user add this tool by self for Agent. Maybe not urgent for now, and we should check similar cases later.

dynamiq/nodes/tools/parallel_tool_calls.py
```diff
@@ -0,0 +1,242 @@
...
+
+from typing import Any, ClassVar, Literal
+
+from pydantic import BaseModel, ConfigDict, Field
+
+from dynamiq.nodes import ErrorHandling, Node, NodeGroup
+
+
+REACT_BLOCK_MULTI_TOOL_PLANNING = """
+MULTI-TOOL PLANNING AND STRATEGY:
```
olbychos: We have to update prompt for multi-tools

dynamiq/nodes/tools/parallel_tool_calls.py
mykhailobuleshnyi: Fixed

dynamiq/nodes/tools/parallel_tool_calls.py
mykhailobuleshnyi: Did I miss something during refactoring? Or do you mean to update prompt itself?

dynamiq/nodes/agents/agent.py
```diff
@@ -645,10 +597,10 @@ def _add_observation_and_stream(
         if self.streaming.enabled and self.streaming.mode == StreamingMode.ALL:
             if tool is not None:
                 source_name = tool_name = tool.name
-            elif isinstance(action, list):
+            elif self.sanitize_tool_name(action) == PARALLEL_TOOL_NAME:
                 tool_names = [
                     tool_data["name"] if isinstance(tool_data, dict) and "name" in tool_data else UNKNOWN_TOOL_NAME
-                    for tool_data in action
+                    for tool_data in action_input["tools"]
```
maksymbuleshnyi, line 598: Before doing this validate it with InputTranformer of this node

dynamiq/nodes/agents/base.py
```diff
@@ -297,6 +297,11 @@ def __init__(self, **kwargs):
             self.tools.append(FileSearchTool(file_store=self.file_store_backend))
             self.tools.append(FileListTool(file_store=self.file_store_backend))
 
+        if getattr(self, "parallel_tool_calls_enabled", False):
```
maksymbuleshnyi: Put parallel_tool_calls_enabled in BaseAgent and call it with just self..... similarly as other parameters

dynamiq/nodes/tools/parallel_tool_calls.py
olbychos: Prompts look outdated and somewhat redundant in their original form, so if we’re refactoring them, it makes sense to update their behavior as well

dynamiq/nodes/agents/agent.py
mykhailobuleshnyi: Done

dynamiq/nodes/agents/base.py
mykhailobuleshnyi: Done

olbychos (approved):
Let’s move the prompt refactor to a new task, @mykhailobuleshnyi

dynamiq/nodes/agents/agent.py
mykhailobuleshnyi: 1. Fixed 
2; 3 to be done in future

acoola (approved):

PR #594, opened 2026-02-27
fix: handle E2B sandbox timeout issues

tyaroshko:
> [!NOTE]
> **Low Risk**
> Small, localized change to sandbox connection parameters; main risk is behavior change if the upstream SDK treats `timeout` differently on connect/reconnect.
> 
> **Overview**
> Fixes E2B sandbox reconnection to *preserve the configured session timeout*.
> 
> `E2BSandbox._reconnect_with_retry()` now always passes `timeout=self.timeout` to `Sandbox.connect()`, and `_ensure_sandbox()` explicitly calls `set_timeout(self.timeout)` after reconnect (with a debug-only fallback if it fails) to avoid reverting to the SDK’s default timeout.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit adfd686d5b0dc98030abf113b3f5133e0c250f82. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/dynamiq/nodes/tools/e2b_sandbox.py
+++ b/dynamiq/nodes/tools/e2b_sandbox.py
@@ -280,7 +280,7 @@ class E2BInterpreterInputSchema(BaseModel):
 
 class E2BInterpreterTool(ConnectionNode):
     """
-    A tool for executing code and managing files in an E2B sandbox environment.
+    A tool for executing code and managing files in a sandbox environment.
 
     This tool provides a secure execution environment for running Python code,
     shell commands, and managing file operations.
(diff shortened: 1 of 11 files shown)


dynamiq/sandboxes/e2b.py
```diff
@@ -115,6 +115,12 @@ def _ensure_sandbox(self) -> E2BDesktopSandbox:
                 try:
                     self._sandbox = self._reconnect_with_retry()
                     logger.debug(f"E2B sandbox reconnected: {self.sandbox_id}")
+                    # Call set_timeout explicitly to ensure the sandbox timeout
```
olbychos, line 122: lets clean comments

tyaroshko: The comments are actually intentional here as the bug fix is not as straightforward, so I'm not sure we want to remove them

vitalii-dynamiq:
@acoola please check

PR #607, opened 2026-03-04
fix: fix context overflow

mykhailobuleshnyi:
> [!NOTE]
> **Medium Risk**
> Changes core agent history compaction behavior and summarization triggering, which can alter what context is preserved/removed during runs. Added unit/integration coverage reduces risk but token counting and early-return paths could affect edge-case conversations.
> 
> **Overview**
> **Context compaction is now token-budgeted instead of message-count based.** `SummarizationConfig` replaces `preserve_last_messages` with `max_preserved_tokens`, and `_split_history` now preserves as many recent messages as fit within that token budget (falling back to summarizing everything if the overall prompt is still over limit).
> 
> **No-op context-manager calls are avoided.** When `_split_history` yields nothing to summarize, the agent returns early (while still streaming a matching tool-result event) and does not modify the prompt; when compaction does run, `_compact_history` can reuse the preserved set computed during splitting.
> 
> Tests were updated for the new config and expected message counts, and new unit/integration regression tests were added to cover token-bounded splitting/compaction and the early-return behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 955a67291411437fb30fdc0b55474c4d61267575. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/dynamiq/nodes/agents/agent.py
+++ b/dynamiq/nodes/agents/agent.py
@@ -673,7 +673,7 @@ class Agent(HistoryManagerMixin, BaseAgent):
                         config,
                         **kwargs,
                     )
-                    return skip_message, [], False, False, None
+                    return skip_message, [], False, True, None
                 tool_input = {**(action_input if isinstance(action_input, dict) else {}), "messages": to_summarize}
             else:
                 tool_cache_entry = ToolCacheEntry(action=action, action_input=action_input)
--- a/tests/integration/nodes/agents/test_agent_methods.py
+++ b/tests/integration/nodes/agents/test_agent_methods.py
@@ -913,7 +913,7 @@ class TestContextManagerEarlyReturn:
         assert "Nothing was summarized" in tool_result
         assert tool_files == []
         assert is_delegated is False
-        assert success is False
+        assert success is True
         assert dependency is None
 
     def test_prompt_unchanged_after_early_return(self, context_agent):


dynamiq/nodes/agents/utils.py
```diff
@@ -967,16 +967,18 @@ class SummarizationConfig(BaseModel):
...
           prompt and generated output. Defaults to 0.75.
     """
 
+    model_config = ConfigDict(extra="ignore")
+
     enabled: bool = False
     max_token_context_length: int | None = None
     context_usage_ratio: float = Field(default=0.8, gt=0, le=1)
-    preserve_last_messages: int = Field(default=4, ge=0)
+    max_preserved_tokens: int = Field(default=10000, ge=0)
```
olbychos, line 983: maybe lets have it and other params as constants

mykhailobuleshnyi: Not sure I understad how you want to make them constants.

tyaroshko: @mykhailobuleshnyi just move these default values into constants at the top of the module (to avoid magic numbers)

mykhailobuleshnyi: Actually I don't see why we need to put this in constants. It is just default values for pydantic model.

PR #613, opened 2026-03-10
feat: add cache control for Anthropic models

tyaroshko:
> [!NOTE]
> **Medium Risk**
> Touches provider request parameter construction and usage accounting; misconfigured caching params could change Anthropic API behavior or break calls, though the change is gated behind an optional `cache_control` setting and covered by tests.
> 
> **Overview**
> Adds first-class Anthropic prompt caching support via a new `AnthropicCacheControl` model, exported from `dynamiq.nodes.llms`, and wires it into `Anthropic.update_completion_params()` to send `cache_control_injection_points` with completions.
> 
> Extends `BaseLLMUsageData` and usage extraction to include `cache_read_input_tokens` and `cache_creation_input_tokens`, and updates integration tests to validate the new request params and assert cache usage behavior (including switching an agent memory integration test to Anthropic with caching enabled).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit be68dc2d38f92fd26c897837c4c576fe5f419ac6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/dynamiq/callbacks/streaming.py
+++ b/dynamiq/callbacks/streaming.py
@@ -330,7 +330,6 @@ class AgentStreamingParserCallback(BaseStreamingCallbackHandler):
         self.loop_num = loop_num
         self.kwargs = kwargs
 
-        # Aggregate streamed text from the LLM in the current loop for proper tracing inside the agent
         self.accumulated_content: str = ""
 
         self._buffer: str = ""
--- a/dynamiq/memory/memory.py
+++ b/dynamiq/memory/memory.py
@@ -104,7 +104,7 @@ class Memory(BaseModel):
             effective_limit = limit if limit is not None else self.message_limit
 
             messages = self.backend.get_all(limit=effective_limit)
-            retrieved_messages = [Message(**msg.model_dump()) for msg in messages]
+            retrieved_messages = [Message(**msg.model_dump(), static=True) for msg in messages]
             logger.debug(f"Memory {self.backend.name}: Retrieved {len(retrieved_messages)} messages")
             return retrieved_messages
         except Exception as e:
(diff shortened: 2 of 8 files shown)


dynamiq/nodes/llms/anthropic.py
```diff
@@ -1,19 +1,32 @@
...
 from dynamiq.nodes.llms.base import BaseLLM
 from dynamiq.prompts.prompts import VisionMessageType
 from dynamiq.utils.logger import logger
 
 
+class AnthropicCacheControl(BaseModel):
+    """Anthropic prompt caching configuration."""
+
+    type: Literal["ephemeral"] = "ephemeral"
+    ttl: Literal["1h"] | None = None
```
maksymbuleshnyi: Maybe also allow 5m ttl, it can be beneficial for some cases.

tyaroshko: It's 5 minutes by default, the `1hr` is charged 2x.

maksymbuleshnyi: OKay, but also allow 5m in Literal

delamainer: We should support many values for the TTL.

```python
ttl: Literal["1h"] | None = None
```
That means the only allowed values are:
- `"1h"`
- `None`

tyaroshko: It was fine because Anthropic's API used `5m` as a default even if `None` was provided. However, now I added explicit `5m` default, so should be fixed.

PR #739, opened 2026-05-14
fix: update saving of function calling messages to memory

tyaroshko:
> [!NOTE]
> **Medium Risk**
> Changes how function-calling fields are persisted/hydrated across memory backends and how agents derive fallback assistant output, which can affect stored conversation history and replay fidelity.
> 
> **Overview**
> Ensures OpenAI function-calling fields (`tool_calls`, `tool_call_id`, `name`) survive memory persistence and can be rehydrated correctly on retrieval/search/replace.
> 
> `Memory.add()` now accepts these fields, stashes them under reserved metadata keys (JSON-encoding `tool_calls` for flat-metadata stores), and `Memory.get_all()`/`search()` restore them while removing the stash keys from user-visible metadata.
> 
> Updates the Dynamiq backend wire format to send/read function-calling fields as first-class members of `data` (and keep metadata at the top level), and updates agent snapshot saving/fallback logic to carry FC fields and to extract a final answer from a `provide_final_answer` tool call when `final_output` is missing.
> 
> Adds unit tests covering Dynamiq API payload shape and FC round-trip behavior, plus a `SQLite.get_all(limit=...)` implementation to match other backends’ limiting semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 200a2e592d4aa0e62bde518c719fdef67e596824. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dynamiq/memory/backends/dynamiq.py
+++ b/dynamiq/memory/backends/dynamiq.py
@@ -62,10 +62,23 @@ class Dynamiq(MemoryBackend):
         if not effective_session_id:
             raise DynamiqMemoryError("Session identifier is required to create a memory item.")
 
-        data_payload = {
+        # Send function-calling fields as first-class 
+        # members of `data` instead of metadata
+        for stash_key in ("_tool_calls", "_tool_call_id", "_name"):
+            metadata.pop(stash_key, None)
+
+        data_payload: dict[str, Any] = {
             "role": message.role.value if isinstance(message.role, MessageRole) else message.role,
             "content": self._format_content_payload(message.content),
         }
+        if message.tool_calls is not None:
+            data_payload["tool_calls"] = message.tool_calls
+        if message.tool_call_id is not None:
+            data_payload["tool_call_id"] = message.tool_call_id
+        if message.name is not None:
+            data_payload["name"] = message.name
+        if metadata:
+            data_payload["metadata"] = metadata
 
         payload: dict[str, Any] = {
             "user_id": effective_user_id,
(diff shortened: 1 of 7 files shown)


olbychos (dismissed):
Resolve bugbot comments, please

dynamiq/memory/backends/dynamiq.py
```diff
@@ -62,10 +62,23 @@ def add(self, message: Message) -> None:
...
             "role": message.role.value if isinstance(message.role, MessageRole) else message.role,
             "content": self._format_content_payload(message.content),
         }
+        if message.tool_calls is not None:
+            data_payload["tool_calls"] = message.tool_calls
+        if message.tool_call_id is not None:
+            data_payload["tool_call_id"] = message.tool_call_id
+        if message.name is not None:
+            data_payload["name"] = message.name
+        if metadata:
```
delamainer: What is metadata?

mykhailobuleshnyi: It is just a dict where you can save info additional for message

delamainer: The issue is that if you send the field which are not expected by some model providers, they might return an error.

mykhailobuleshnyi: it is not for llm it is for memory. it is just way to save more info, then string content.

delamainer: From the backend side there is an option to store metadata outside of the message content.
```json
{
  "user_id": "...",
  "session_id": "...",
  "type": "...",
  "data": {...},
  "metadata": {...} <- here (not inside the data)
}
```

tyaroshko: Updated

delamainer (approved):

mykhailobuleshnyi (approved):

PR #756, opened 2026-05-28
feat: flattened tool call schema

mykhailobuleshnyi:
> [!NOTE]
> **High Risk**
> Changes the FC contract (schemas, parsing, and streaming deltas) across the agent pipeline; consumers expecting nested `action_input` in streamed or parsed payloads need to adapt, though runtime unwraps the old shape.
> 
> **Overview**
> This PR **flattens function-calling tool arguments**: `thought` and each tool parameter are **top-level** fields in the LLM schema and wire JSON, replacing the nested `action_input` wrapper. **`ToolCallArguments.to_action_input()`** still unwraps the legacy nested shape when the model emits only `thought` + `action_input`.
> 
> **Streaming** adds **`JSONInnerThoughtsExtractor`** and rewires FC handling so **`thought`** streams as **REASONING** and the remaining JSON (without `thought`) as **TOOL_INPUT**, with optional **`fc_wait_for_first_key`** ordering and an end-of-stream fallback when `thought` is missing. **`provide_final_answer`** still uses the existing JSON-field path.
> 
> **Schema generation** puts **`thought` first**, merges tool properties at the root, and relaxes strict/open rules for `extra="allow"` tools. **ThinkingTool** uses **`reflection`** (with `thought` as an alias) so it does not collide with agent-level `thought`. Prompts, tests, and a live OpenAI callback test cover the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd37974e3ebfc70040b9d974b8d44f7a582dc013. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dynamiq/callbacks/inner_thoughts_extractor.py
```diff
@@ -0,0 +1,253 @@
...
+        self.main_buffer += delta
+        self.main_json_held_buffer = ""
+        self.hold_main_json = False
+        return delta
+
+    def _emit_thought(self, s: str) -> str:
+        self.inner_thoughts_buffer += s
+        return s
+
+    def _process_char(self, c: str) -> tuple[str, str]:
```
olbychos, line 86: Let's rework this, remove 'matreshka' style

olbychos (commented):
Overall looks solid. Let’s consider better patterns and rely on tests.

dynamiq/callbacks/inner_thoughts_extractor.py
mykhailobuleshnyi: What do you mean by that?

maksymbuleshnyi (approved):

tyaroshko (approved):

PR #781, opened 2026-06-12
fix: disable sampling params for Claude 4.7+ models

tyaroshko:
> [!NOTE]
> **Medium Risk**
> Touches core LLM request paths for all providers and changes error/retry semantics; behavior is well-tested but misclassification of API errors could hide real misconfiguration.
> 
> **Overview**
> **LLM completion calls** now avoid sending `temperature`, `top_p`, and `top_k` to models that reject them (notably newer Claude markers such as `claude-opus-4-7` / `4-8`), both **up front** via `update_completion_params` and as a **reactive retry** when the API returns an unsupported-parameter error—while **out-of-range validation errors** still fail normally.
> 
> **Sync and async `execute`** use a **multi-step recovery loop** (capped at 3 attempts) instead of a single retry: each failure can strip more bad params, and **instance fields are only cleared after a successful retry** via `_persist_completion_recovery`, so reusable nodes are not mutated on failed retries.
> 
> **Bedrock** keeps its `stopSequences` recovery but **defers persistence** to the same pattern and **falls through** to the base sampling-param backstop; tests cover detection, stripping, backstop behavior, and chained stop-then-sampling recovery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f765af87364948524aa040cf28c2a59082c8266. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/unit/nodes/llms/test_sampling_params_unsupported.py
+++ b/tests/unit/nodes/llms/test_sampling_params_unsupported.py
@@ -88,8 +88,25 @@ class TestReactiveBackstop:
         assert recovered is not None
         assert "temperature" not in recovered
         assert recovered["max_tokens"] == 100
+        # The recovery hook is pure: it must not mutate the instance, so a failed retry
+        # leaves the reusable node untouched.
+        assert anthropic_supported.temperature == 0.5
+
+    def test_persist_recovery_clears_dropped_params(self, anthropic_supported):
+        # Persistence runs only after a successful retry, mirroring execute().
+        anthropic_supported.temperature = 0.5
+        common = {"model": "anthropic/claude-opus-4-6", "temperature": 0.5, "max_tokens": 100}
+        recovered = {"model": "anthropic/claude-opus-4-6", "max_tokens": 100}
+        anthropic_supported._persist_completion_recovery(common, recovered)
         # Persisted so the next call skips the failing first attempt.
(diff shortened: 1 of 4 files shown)


maksymbuleshnyi (dismissed):

tests/unit/nodes/llms/test_sampling_params_unsupported.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@pytest.fixture
+def anthropic_supported():
+    return Anthropic(name="a", model="claude-opus-4-6", connection=AnthropicConnection(api_key="x"))
+
+
+class TestDetection:
+    @pytest.mark.parametrize(
+        "model",
```
maksymbuleshnyi: Here use actual list you defined in base.py

mykhailobuleshnyi (dismissed):

tests/unit/nodes/llms/test_sampling_params_unsupported.py
tyaroshko: Fixed

mykhailobuleshnyi (approved):

maksymbuleshnyi (approved):

PR #876, opened 2026-08-06
build(deps): batch security updates from open dependabot PRs

delamainer:
## Summary

Consolidates all 11 open dependabot PRs into a single resolution, plus 6 packages that have open security advisories but no dependabot PR. Everything is re-resolved in one `uv lock` pass so the upgrades are verified together rather than as 11 independent locks against a stale base.

Closes #839, #840, #846, #851, #853, #855, #865, #866, #870, #871, #872.

### From the open dependabot PRs

| PR | Package | From → To |
|---|---|---|
| #872 | cryptography | 49.0.0 → 50.0.0 |
| #871 | aiohttp | 3.14.1 → 3.14.3 |
| #870 | gitpython | 3.1.50 → 3.1.57 |
| #866 | nltk | 3.9.4 → 3.10.0 |
| #865 | datamodel-code-generator | 0.30.0 → 0.64.0 |
| #855 | pypdf | 6.14.1 → 6.14.2 |
| #853 | setuptools | 78.1.1 → 83.0.0 |
| #851 | pyasn1 | 0.6.3 → 0.6.4 |
| #846 | pillow | 12.2.0 → 12.3.0 |
| #840 | torch | 2.11.0 → 2.13.0 |
| #839 | mcp | 1.27.0 → 1.28.1 |

### Advisories with no dependabot PR

`starlette` 0.48.0 → 1.3.1, `transformers` 5.3.0 → 5.5.0, `soupsieve` 2.8.3 → 2.8.4, `python-socketio` 5.16.1 → 5.16.2, `python-engineio` 4.13.1 → 4.13.2, `pymdown-extensions` 10.21.3 → 11.0.

### Constraint changes

Two of those were unreachable behind our own version ceilings, so the floors were raised to the **minimum** release that permits the patched version rather than jumping to latest:

- `fastapi` `>=0.116.1,<0.117.0` → `>=0.133.0,<0.134.0`. 0.133.0 is the first release that drops the `starlette<0.49.0` cap. Knock-on effect: chainlit moves to 2.11.1. fastapi is only used under `examples/`, never in `dynamiq/`.
- `mkdocs-material` `>=9.5.27,<9.6.0` → `>=9.7.0,<9.8.0`, the first release accepting pymdown-extensions 11.
- `datamodel-code-generator` is pinned to `==0.64.0` as dependabot proposed in #865.
- `setuptools` is `>=83.0.0,<83.1.0`. Dependabot proposed only raising the ceiling (`>=78.1.1,<83.1.0`) in #853, which would have left the vulnerable 78.1.1 installable; the floor was lifted to 83.0.0 in 8f75be4 so the advisory is actually closed.

### Known gap

`weasyprint` stays at 68.1. Its advisory has no patched release published yet, so that alert stays open until upstream ships a fix.

## Test plan

- [x] `uv lock --check` clean; all 17 target versions confirmed present in the lock
- [x] Full CI test set (`pytest tests --ignore=tests/integration_with_creds -m "not smoke"`): 3252 passed, 2 skipped
- [x] `make build-mkdocs` succeeds under mkdocs-material 9.7
- [x] `pre-commit run --all-files` passes
- [x] `uv sync --group examples` resolves and installs
- [x] CI green on this PR

### Note on the torch check

#840 currently shows a failing `test` check: `test_react_agent_map_streaming_all_mode_isolated` dying on a pydantic `'MockValSer' object cannot be converted to 'SchemaSerializer'` error. That PR's lock only moved torch, Triton and the NVIDIA CUDA libraries and never touched pydantic, and the test passes here with torch 2.13.0 installed. Treating it as flakiness on a month-old base rather than a real regression.


delamainer:
## Why 6ca0dd2d was added to this PR

The `setuptools` bump in this branch has a side effect that CI cannot see, so I pushed a fix for it alongside.

### The problem

`pkg_resources` was removed in **setuptools 82.0.0** (81 still ships it with a "slated for removal" warning). This branch moves setuptools to 83, so `pkg_resources` disappears from the environment.

`milvus-lite` **2.5.1** — the last 2.x release, and what the lockfile pinned — imports it unguarded at module load:

```python
from pkg_resources import DistributionNotFound, get_distribution
```

An AST scan of the whole synced `site-packages` confirms it is the *only* remaining unguarded importer; litellm, pytz, wrapt and babel all fall back to `importlib`.

pymilvus imports `milvus_lite` lazily for local `.db` URIs, so with this branch as originally submitted:

```
Milvus(deployment_type=MilvusDeploymentType.FILE, uri="x.db").connect()
→ ModuleNotFoundError: No module named 'pkg_resources'
```

Every Milvus Lite / FILE deployment breaks. CI stays green because `tests/integration/nodes/{writers,retrievers}/test_milvus_*_flow.py` patch `pymilvus.MilvusClient`, so `milvus_lite` is never really loaded.

The Dockerfile installs the dev group into the app environment (`UV_PROJECT_ENVIRONMENT=/usr/local`, `uv sync --frozen`), so the shipped image is affected too, not just local dev.

### The fix

**1. `uv lock --upgrade-package milvus-lite`** → `2.5.1 → 3.2.0`, adds `faiss-cpu` (milvus-lite 3.x is now a pure-Python wheel and delegates to faiss/pyarrow). milvus-lite 3.x dropped the `pkg_resources` import entirely, and pymilvus 2.5.18 declares `milvus-lite>=2.4.0` with no ceiling, so no pymilvus change is needed. Worth noting: a plain `pip install pymilvus==2.5.18` already resolves setuptools 83 + milvus-lite 3.2.0 today — the lockfile was the only thing keeping us on 2.5.1.

**2. `index_type="INVERTED"` on the `id` index** in `MilvusVectorStore.__init__`. milvus-lite 3.x validates index params more strictly and rejects an entry with no `index_type`:

```
MilvusException: (code=6, message=create_index missing required 'index_type' parameter)
```

`INVERTED` is the scalar index the server already picks for a VARCHAR field. `AUTOINDEX` is not a substitute — Milvus rejects it on non-vector fields (`field 'id' has type VARCHAR; create_index with index_type 'AUTOINDEX' requires a vector field`).

This second issue is not caused by the dependency bump — it already affects anyone installing dynamiq from PyPI, since their resolver picks milvus-lite 3.x regardless of our lock.

### Verification

- **No behavioural change.** Per-test outcomes (not just counts) diffed across three full suite runs — main's dependency stack, this branch as submitted, and this branch with the fix. All three are byte-identical across 3250 tests: `3249 passed, 2 skipped, 88 deselected`.
- **Milvus Lite works again** with `pkg_resources` genuinely absent: create collection → write → count → embedding search → metadata filter → list → delete → drop, all pass.
- **Real-server no-op.** Ran the same CRUD path against a Milvus 2.5.14 standalone container with the original and the patched code — identical output line for line, including `get_documents_by_id`, `filter_documents` and `replace_document_metadata`. The only observable difference is `describe_index("id")` now reporting `index_type: INVERTED` where it previously reported none. Index params apply only at `create_collection`, so **existing collections are untouched** (verified by reopening one with `create_if_not_exist=False`).
- **Backward compatible.** `INVERTED` also works on the old milvus-lite 2.5.1, so stale environments aren't broken. Search ranking matches on both.
- **Build safe.** faiss-cpu has manylinux x86_64 wheels and milvus-lite 3.2.0 is `py3-none-any`, so the linux/py3.13 Docker build has wheels for everything. `uv lock --check` and pre-commit pass.

### Known cosmetic issue

milvus-lite 3.x logs `NotImplementedError: AllocTimestamp Method not implemented!` from its gRPC server thread during FILE-mode operations. It does not propagate and results are correct. It also occurs with pymilvus 3.0.1, so it is inherent to milvus-lite 3.x rather than a version-mismatch artifact.

### Follow-up worth doing separately

The Milvus tests mock `pymilvus.MilvusClient`, so none of the above is covered. A single FILE-mode test that actually touches milvus-lite would have caught both issues.

---

The rest of the branch checks out: starlette `0.48 → 1.3.1` is a major bump but fastapi, sse-starlette, mcp, chainlit and streamlit all work against it (SSE streaming and WebSocket paths from the examples verified end to end); the mcp 1.28 client signatures still match our call sites; and the docs build produces an identical warning count before and after the mkdocs-material/pymdown-extensions bumps.

tyaroshko (dismissed):

tyaroshko (approved):

PR #887, opened 2026-08-12
feat: per-node mock (dry run) execution for tools and agents

vitalii-dynamiq:
## What

Adds a **dry run mode**: a node can be marked so it never executes. The framework short-circuits it and returns a synthetic result in its place.

Two use cases drive this:

- **Sensitive tools** — a tool that writes to production sits on a canvas and takes part in a run without ever firing.
- **A/B testing** — the same workflow runs twice, once for real and once side-effect-free, to compare agent behaviour.

UI companion: dynamiq-ai/ui#1521.

## Design

The seam is in `Node.run_sync` / `Node._run_async_native` and replaces **only** the `execute` call:

```
validate_depends → [approval, if not mocked] → transform_input → validate_input_schema
→ on_node_start → [ mock ?  synthetic result  :  cache(execute_with_retry) ]
→ transform_output → on_node_end
```

Everything except the side effect still happens, so **a dry run still fails when an agent calls a tool with malformed arguments**. Mocking sits below validation and above execution, deliberately.

This matches the dominant pattern across the ecosystem — ADK's `before_tool_callback` returning a dict to skip the tool, LangChain's `wrap_tool_call` not calling `handler`, Semantic Kernel's filters setting `context.Result` without calling `next`, Pydantic AI's `WrapperToolset.call_tool`.

### Per-node: `Node.mock`

| field | effect |
|---|---|
| `enabled` | short-circuit this node |
| `locked` | no run-level policy can turn it back on — for nodes that must never fire |
| `output` | `dict` returned **verbatim** (author owns the whole shape); `str` Jinja-rendered against the node input and wrapped as `{"content": ...}`; `None` synthesises a bounded description of the suppressed call |
| `error` | inject a recoverable failure instead, to exercise error-handling paths |
| `latency_seconds` | simulated latency, cancellable on both the sync and async seams |
| `marker` | prefix for string output (default `[MOCKED]`); `None` disables it |

### Per-run: `RunnableConfig.mock`

`policy` is `NODE` (honour node configs, default), `ALL` (whole-run dry run) or `NONE` (real run, ignoring configured mocks). `ALL` is scoped by node group — **tools only by default, so agents keep reasoning with real LLMs** — with unknown group names rejected rather than silently matching nothing. `exclude_ids` keeps chosen nodes live unambiguously; `exclude_names` exists too but carries an explicit warning that names are class defaults and not unique. Framework machinery (`ContextManagerTool`, `TodoWriteTool`, `SkillsTool`) is never swept in by `ALL` (`is_mockable = False`) — mocking an agent's own context manager corrupts its history rather than protecting the outside world; explicit node-level mocks still apply.

This means a canvas needs no edits to be dry-run: the policy lives on the run.

### Mocked runs are marked

Output carries `is_mocked` (tool-shaped mocks; authored dicts stay verbatim), and traces carry `metadata.is_mocked` on **both** success and failure. Of twelve frameworks surveyed, none marks a mocked result in the message history or trace — ADK emits an identical `execute_tool` span whether the tool ran or a plugin faked it. Without this, an eval run and a production run are indistinguishable in observability.

The synthesised description is bounded (`reprlib` + hard cut) and **strips fields marked `is_accessible_to_agent=False`**, so a mocked run can never reveal more than a real one.

## Interaction with human-in-the-loop

**A mocked node does not request approval**, even with `approval.enabled` set. When both are configured the skip is logged, so it is never silent. Approval authorizes a real side effect; a mocked node has none — and prompting would deadlock unattended `MockPolicy.ALL` runs on every approval-gated tool (Mastra ships exactly this deadlock: its `toolMocks` experiments suspend mid-run on `requireApproval` tools).

| framework | mocked call still prompts a human? |
|---|---|
| **Google ADK** | No — `require_confirmation` checked inside `FunctionTool.run_async`; simulation short-circuits upstream |
| **Agno** | No, by exclusion — the flags are mutually exclusive |
| **Letta** | Unified — the supplied result *is* the approval response |
| **LangChain / LangChain.js** | Yes, always — HITL is `afterModel`, emulator is `wrapToolCall`; ordering can't fix it |
| **Mastra** | Yes — and its own experiments deadlock on it |
| **Strands** | Yes — intervention hooks precede mock middleware |
| **OpenAI Agents** | Yes by default (`pre_approval_tool_input_guardrails=False`) |
| **LlamaIndex** | No, but worse — approval lives inside the tool body, so mocking deletes it |

`HumanFeedbackTool` is an ordinary tool the agent chooses to call, so **mocking it is the supported way to supply a canned human answer** and evaluate a HITL workflow unattended.

## MCP

MCP servers are a primary use case, so they get first-class handling:

- An `MCPServer` is a discovery wrapper whose `execute()` is disabled, so each discovered `MCPTool` **reads its owning server's mock at resolve time** (not snapshotted at discovery — a mock configured later still applies), and answers to the server's id/name in `exclude_ids`/`exclude_names`.
- **An unreachable mocked server degrades instead of dying**: discovery failure on a server whose mock is enabled logs a warning and contributes no tools, so a canvas with a pinned-off MCP server loads and runs without connectivity to it. An unreachable *unmocked* server still fails loudly, as a real run must.
- **Validated against real MCP servers two ways**: a live public server (DeepWiki), and a purpose-built instrumented local server (stdio + streamable-http) whose own invocation log is the ground truth — 21 scenarios covering discovery, per-tool/server-level/run-level mocking, locked-vs-NONE, exclusion, injected vs real server errors, verbatim dict output, a full stubbed-LLM agent A/B loop (read-only tool ran for real, charge suppressed, `[MOCKED]` observation reached the model), simulated latency, unreachable-at-discovery, and mid-session server death (mocked call keeps working, unmocked call fails cleanly; zero server-side invocations recorded).

## Coverage of every tool shape

Audited every node in `dynamiq/nodes/tools`. Exactly two are wrappers whose `execute()` is disabled and which hand work to a node that does run — `MCPServer` (above) and `SubAgentTool`: a mock on it suppresses the whole delegation, resolved **before** the sub-agent is built (both initialized and factory modes; factory mode is the parallel-delegation path). Everything else routes through `Node.run` and is covered directly: Pipedream, MCP tools, HTTP, Python/E2B/Daytona, SQL/Cypher, search/scrape tools, file tools, `HumanFeedbackTool`. `ParallelToolCallsTool` is schema-only — each real tool dispatches through `_run_tool` individually, and id-based exclusions survive the cloning that path performs (shared `dynamiq/nodes/cloning.py` used by both the agent and the `Map` operator, replacing two duplicate id-regeneration walkers).

## Checkpoint resume keeps the run config

`Flow` always wrote `original_config` into the checkpoint but never read it back, so a dry run resumed from a mid-flow snapshot without `mock=` executed its remaining mocked nodes for real — the same gap applied to `dry_run` and `max_node_workers`. The restore mirrors `original_input`: explicit caller values win, unset fields fall back to the checkpoint. Runtime carriers (callbacks, cache backend, streaming queues, cancellation) cannot round-trip and stay caller-owned.

## Other guards

Each with a regression test that fails without its fix:

- An **injected LLM failure does not trigger the configured fallback provider** — otherwise a dry run makes a real, billed call.
- **`is_mocked` is framework-reserved**, dropped before validation can fail, so a caller kwarg can never mark a real run as mocked.
- Async simulated latency is **cancellable**.
- Id-based exclusions are **carried onto clones**, nested sub-agent tools included, with a set-valued id map so one original cloned twice keeps both exclusions.

## Known limits (documented in the module docstring)

- **No subtree scope**: two agents sharing one tool instance cannot mock it differently — `resolve_mock` sees the node, not the caller. The working A/B pattern is per-agent tool instances selected with `exclude_ids`. A caller-scoped policy is follow-up work.
- **MCP tool schemas come from discovery**: for full fidelity (the agent sees and "calls" the real tool names) the server should be reachable at startup; only execution is suppressed. Unreachable mocked servers degrade as described above.

## Testing

`3482 passed, 1 skipped` across `tests/unit` and `tests/integration`. ~125 mock-specific tests across four files (node seams, policy matrix, agent loop, MCP/Pipedream/HumanFeedback, approval interaction, cloning, checkpoint-resume restore, observability, serialization round-trip). Every guard above was verified to fail without its fix. `flake8`, `darker` (black + isort) and `bandit` clean. Example: `examples/mocking/mock_tools.py`.


dynamiq/types/mocking.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+class MockConfig(BaseModel):
+    """Per-node mock configuration — suppress this node and return a synthetic result.
+
+    Attributes:
+        enabled (bool): Replace this node's execution with a synthetic result.
+        locked (bool): Keep mocking even under ``MockPolicy.NONE``. Use for
+            nodes that must never fire — production writes, payments, emails.
+        response (str | dict | None): The synthetic result. A ``dict`` is
```
delamainer: I would rename this to `output`.

dynamiq/types/mocking.py
```diff
@@ -0,0 +1,307 @@
...
+            returned as the node output verbatim. A ``str`` is Jinja-rendered
+            and wrapped as ``{"content": ...}``. ``None`` synthesises a
+            description of the suppressed call.
+        latency_seconds (float): Artificial delay, to keep timing-sensitive
+            behaviour comparable to a real run.
+        annotate (bool): Prefix string output with ``marker`` so a mocked
+            result is visually unmistakable in transcripts and traces.
+        marker (str): The prefix used when ``annotate`` is set.
```
delamainer: I think we can merge annotate and marker into a single parameter. Like `marker: str | None = ...`. If the marker is None, we ignore it. If it's not, we apply it.

acoola:
The design is good and the seam is in the right place, but one thing blocks the merge: a mocked `SubAgentTool` in factory mode crashes the run. Eight important items below.

## Blocker: a mocked `SubAgentTool` in factory mode crashes the run

When the delegation is mocked, `resolved_agent` is set to `None` at `dynamiq/nodes/agents/base.py:1613`. But the factory branch at `:1713` does not check for that. It passes `None` into `_clone_tool_for_execution`, which reads `tool.id` at `:1459`. That line sits outside the try/except, so it raises. The fix at `:1722` runs after the crash, so it never helps.

I ran the same tool built both ways:

```
initialized mode (agent=):     -> ('[MOCKED] simulated delegation', [], {'is_mocked': True})
factory mode (agent_factory=): -> AttributeError: 'NoneType' object has no attribute 'id'
```

This is not a rare case. `validate_agent_config` sets `is_parallel_execution_allowed = self.is_factory_mode` (`agent_tool.py:155`). So factory mode is the only mode that allows parallel sub-agent delegation. Anyone who runs several A/B agents at the same time and mocks some of them will hit this.

Fix: change `:1713` to `if is_child_agent and tool.is_factory_mode and not is_delegation_mocked:`. Then delete the line at `:1722`, because `:1709` already returns `tool` when `resolved_agent` is `None`. Also move `original_id = tool.id` inside the try at `:1459`.

## Important

1. **The echo sends hidden fields, including secrets, to the LLM and the trace.** `_build_mocked_output` echoes the whole validated input (`node.py:1053`). It ignores `is_accessible_to_agent=False`, which is how this codebase marks fields the agent must never see (`schema_utils.py:19-27`). I ran a node with a hidden `access_token` and got back: `[MOCKED] Execution of 'crm_update' was skipped: ... It was called with: {'access_token': 'ghp_SUPERSECRET_TOKEN_123', 'query': 'acme'}`. `CodeInterpreter` marks `env` hidden for exactly this reason (`code_interpreter.py:254-263`). In a real run these values go to the tool and never reach the transcript. So mocking makes the exposure worse, by default, with no node config. Fix: strip hidden fields before the echo with `strip_inaccessible_fields`, or make the echo opt-in.

2. **`MockPolicy.ALL` mocks the agent's own tools and destroys its history.** `ContextManagerTool` has `group=tools` and name `context-manager`, so the default `groups={"tools"}` sweeps it in. I confirmed this by running `resolve`. The agent calls this tool by itself when the token budget is passed (`agent.py:1939`), and `agent.py:1386` reads `tool_output_meta.get("summary", tool_result)`. The mocked output has only `['content', 'is_mocked']` — no `summary`. So the generic echo becomes the summary, and `_compact_history` replaces the whole history with it. The agent then continues with no memory. The echo also repeats the messages it was asked to shorten, so the prompt grows instead of shrinking. Fix: add `Agent._excluded_tool_ids` to `config.mock.exclude`, and skip `_compact_history` when the result is mocked.

3. **`ECHO_INPUT_MAX_CHARS` does not limit the echo.** The comment at `mocking.py:83-94` says it uses the "same idiom as the payload logger in `dynamiq.nodes.node`". It does not. That logger cuts the string after rendering (`node.py:1145-1146`). `describe_skipped_call` has no such cut. `reprlib` limits each element, not the total, so `maxdict=20` × `maxlist=20` × `maxstring=1000` multiply. Measured: a normal 20-document payload gives 20,442 characters against a documented 1,000. A 50-row bulk input gives 403,834. A valid nested shape gives 8,069,097. All of it goes into every later LLM call and into the trace. Fix: copy the cut from `node.py:1145-1146`.

4. **"A dict response is returned verbatim" is wrong, and it is written in three places.** `render` adds `content` (`:204`), the annotate block turns that empty string into the bare marker (`:218-221`), and `:223` adds `is_mocked`. So `response={"documents": [...], "score": 0.9}` comes back as `{'documents': [...], 'score': 0.9, 'content': '[MOCKED]', 'is_mocked': True}`. The module docstring at `:57-64` tells authors to give non-tool nodes "a dict response matching the real shape" — but this code makes that impossible. The test at `test_node_mocking.py:72` is called `test_dict_response_is_returned_verbatim_...` and asserts the opposite. Fix the code or the docs, and rename that test.

5. **Mocking has node scope and run scope, but no subtree scope — which is what A/B testing needs.** `NodeRunnableConfig` only has `streaming`, so there is no `nodes_override[...].mock`. If three agents share tool instances, which is the natural way to build it, `Node.resolve_mock` sees only `self.id`, `self.name` and `self.group`. All three callers look the same, so all three fire the real tool. The only working setup is to give each agent its own tool instances and list the live agent's ids in `exclude`. That fails in the dangerous direction: add a tool to the live agent later, forget to update `exclude`, and it quietly stops doing the real work while still reporting SUCCESS. Fix: add `mock` to `NodeRunnableConfig` and read `nodes_override[self.id].mock` first, the same way streaming overrides work at `node.py:1735`.

6. **A mocked and `locked` MCP server is still contacted over the network.** `Agent.__init__` always expands MCP servers with `get_mcp_tools()`, and `get_mcp_tools_async` calls `initialize_tools()` — a real `connect()` and `list_tools()`. Nothing checks the mock there. So an operator who sets `locked=True` to make sure a run never touches production gets `ToolExecutionException: failed to initialize session`, and the run dies before any node runs. Fix: skip discovery when the server resolves to a mock, or at least document that a mocked MCP server must still be reachable.

7. **You cannot exclude an MCP server by its own id or name.** `resolve_mock` passes `self.mock_config`, which can be the server's config, together with the tool's id and name (`node.py:1017`). `resolve` only matches those against `exclude` (`mocking.py:290`). So there is no way to name the node that actually holds the mock. The operator has to know every tool name found at runtime. `test_mcp_pipedream_mocking.py` has eight propagation tests and none for exclusion. Fix: also offer the server's id and name to `resolve` when the mock came from the server.

8. **Name-based `exclude` can match the wrong node and let a production write fire.** `exclude` matches id or name, and a match beats the node's own `mock.enabled` unless it is `locked`. Names are class defaults and are not unique — `HttpApiCall` is `"api-call"`, `function_tool` uses `func.__name__`. `Flow._validate_nodes` only checks ids (`flow.py:137`), and tools inside agents are not flow nodes at all. Take two `HttpApiCall` tools: a read-only GET, and a POST to `/charge` that was shipped with `mock.enabled=True`. Writing `exclude={"api-call"}` matches both, and the charge fires. This happens under the default `MockPolicy.NODE`. The example even promotes the name form (`examples/mocking/mock_tools.py:78`). Fix: split id and name into separate fields, or match a name only when it is unique.

## Requested changes

1. Add `not is_delegation_mocked` to the factory branch at `agents/base.py:1713`, and add a factory-mode test.
2. Strip `is_accessible_to_agent=False` fields from the echo.
3. Keep the agent's own internal tools out of `MockPolicy.ALL`, and skip `_compact_history` when the result is mocked.
4. Add the missing length cut that `node.py:1145` has.
5. Make "dict returned verbatim" true in the code, or fix the docs and rename the test.
6. Add `mock` to `NodeRunnableConfig` so a policy can cover one agent's subtree.
7. Skip MCP discovery for a mocked server, and let `exclude` match the server's id or name.
8. Split id matching and name matching in `exclude`.

vitalii-dynamiq:
Thanks — this was a genuinely good review. I reproduced every item before changing anything; all nine were real. Seven are fixed, two are documented as known limits with reasoning below. Rebased onto current `main` (your merge is included as a rebase, so the branch is one commit again).

`3416 passed, 1 skipped`. Every fix has a test that fails without it.

## Blocker — fixed

Reproduced exactly as you described:

```
initialized mode (agent=):     -> ('[MOCKED] simulated delegation', [], {'is_mocked': True})
factory mode (agent_factory=): -> AttributeError: 'NoneType' object has no attribute 'id'
```

Took your fix: `not is_delegation_mocked` on the factory branch, and deleted the now-dead override — `resolved_agent is None` already yields the wrapper. Also added a `tool is None` guard inside `_clone_tool_for_execution` so that path can never explode again. `TestMockedSubAgentToolInFactoryMode` covers both the mocked and unmocked factory paths; it fails without the branch fix.

Your point about severity was the decisive one — `is_parallel_execution_allowed = self.is_factory_mode` means factory mode is *the* parallel delegation path, so any parallel A/B run hits this.

## 1, 3, 4, 7, 8 — fixed as suggested

**1. Secret leak.** Confirmed a hidden `access_token` reaching the transcript. The description now runs through `strip_inaccessible_fields` against `resolved_input_schema`. Visible arguments still appear; hidden ones are dropped and logged at debug. You were right that this made mocking *worse* than a real run by default.

**3. `ECHO_INPUT_MAX_CHARS` didn't bound anything.** My regression, and my comment claiming parity with the payload logger was simply wrong — that logger cuts after rendering and I'd omitted the cut. Measured 30,771 chars on your 20-document case. Added the cut from `node.py:1145`; `reprlib` now bounds breadth and the slice bounds total.

**4. "Verbatim" was false.** Fixed the code rather than the docs, since the docstring's advice to non-tool nodes only makes sense if it's true. A dict response is now returned exactly as authored — no `content`, no `is_mocked`, no annotation. `metadata.is_mocked` in the trace is unaffected, so mocked runs stay identifiable. Test renamed and a second one added for the retriever shape.

**7. MCP server not addressable.** Added `Node.mock_identity`, which `MCPTool` extends with its owning server's id and name, so `exclude_ids={server.id}` reaches the discovered tools. Two tests added.

**8. Name matching.** Split into `exclude_ids` and `exclude_names`. `exclude_names` carries an explicit warning that names are class defaults and not unique. Your `api-call` scenario is now a test.

## 2 — fixed, but not the suggested way

Confirmed `ContextManagerTool` and `TodoWriteTool` are swept in, and that `agent.py` reads `tool_output_meta.get("summary", ...)`, so the history destruction is real.

I didn't add `_excluded_tool_ids` to `mock.exclude` though: that set also holds MCP subtools, which are the highest-value mocking targets, so it would quietly un-mock the tools most likely to touch production. Instead there's `Node.is_mockable` (ClassVar, default `True`), set `False` on `ContextManagerTool`, `TodoWriteTool` and `SkillsTool` — framework machinery that "mock all the tools" never means. `MockPolicy.ALL` skips them; an explicit node-level mock still applies, so nothing is taken away. That also means `_compact_history` can't receive a mocked result, so it needs no change.

## 5 and 6 — documented as known limits

**6. MCP discovery still connects.** Correct. But skipping discovery would leave the agent unaware the tools exist, so it takes a different path than the real run — which defeats the comparison a dry run is for. Fixing it properly means caching schemas, a separate feature. Documented: a mocked MCP server must still be reachable.

**5. No subtree scope.** Agreed this is the real gap for A/B. I don't think `nodes_override[...].mock` closes it, though: `nodes_override` is keyed by node id, and shared instances have the *same* id across all three agents, so the override can't tell the callers apart either. Genuinely solving it needs the agent to push scoped config down its subtree — bigger than this PR. Documented, with the working pattern (per-agent tool instances selected by `exclude_ids`) spelled out.

Happy to file both as issues if you'd prefer them tracked rather than only in the docstring.

---

dynamiq/types/mocking.py
vitalii-dynamiq: Renamed to `output` in 32a3255 — agreed it matches the codebase vocabulary (`RunnableResult.output`, `output_transformer`). Propagated through tests, the example, and the UI companion patch.

---

dynamiq/types/mocking.py
vitalii-dynamiq: Merged in 32a3255: `marker: str | None = "[MOCKED]"` — `None` disables the prefix, and the meaningless `annotate=False + custom marker` state is gone. Applies to the injected-error message too, and a `None` marker renders as an empty string inside authored Jinja templates rather than the literal "None".

---

delamainer (dismissed):

dynamiq/nodes/tools/mcp.py
```diff
@@ -414,6 +417,33 @@ def get_input_schema(schema_dict) -> type[BaseModel]:
...
+    @property
+    def mock_config(self) -> MockConfig:
+        """Fall back to the owning server's mock.
+
+        ``MCPServer`` is a discovery wrapper whose ``execute`` is disabled, so a mock set on it
+        can only take effect through the tools it discovered. Reading it here rather than
+        copying it at discovery time means a mock configured after discovery still applies.
+        """
+        owner_mock = getattr(self._owner_server, "mock", None)
+        if not self.mock.enabled and owner_mock is not None and owner_mock.enabled:
```
vitalii-dynamiq: ### A tool-level mock discards the owning server's `locked` pin

**Medium severity**

When a tool has its own enabled mock, this returns `self.mock` and the server's config is never consulted — including its `locked` flag. A server pinned off with `MockConfig(enabled=True, locked=True)` therefore loses its pin for any tool that carries its own mock.

Chain: server `mock=MockConfig(enabled=True, locked=True)`, and one discovered tool given `tool.mock = MockConfig(enabled=True, output="…")` (`locked` defaults to `False`). A run with `MockPolicy.NONE` calls `resolve_mock` → `mock_config` returns the tool's unlocked config → `RunMockConfig.resolve` hits the `MockPolicy.NONE` branch → `node_mock.is_pinned` is `False` → returns `None` → `MCPTool.execute_async` opens a real session and calls the tool on the server the operator declared must never fire.

Tool-level precedence itself is intended and tested (`test_a_tools_own_mock_wins_over_the_servers`); what is lost is the safety pin, which the tool author never opted out of. The owner's `locked` needs to survive the override — e.g. return `self.mock.model_copy(update={"locked": True})` when the owner's mock `is_pinned`.

---

dynamiq/nodes/agents/base.py
```diff
@@ -1634,9 +1622,13 @@ def _run_tool(
...
 
         is_child_agent = isinstance(tool, SubAgentTool)
+        # Resolved before the sub-agent is built: a mocked delegation suppresses the whole
+        # sub-agent, so constructing it (LLM clients, tools, factory blueprints) just to throw
+        # it away would be wasted work on every suppressed call.
+        is_delegation_mocked = is_child_agent and tool.resolve_mock(ensure_config(config)) is not None
         resolved_agent = None
         try:
-            resolved_agent = tool.get_or_create_agent() if is_child_agent else None
+            resolved_agent = tool.get_or_create_agent() if is_child_agent and not is_delegation_mocked else None
```
vitalii-dynamiq, line 1631: ### A mocked delegation echoes injected `user_id`/`session_id`/`metadata` into the parent transcript

**Medium severity**

Skipping `get_or_create_agent()` leaves `resolved_agent` as `None`, so the mocked delegation runs on the `SubAgentTool` wrapper instead of the sub-agent. The context injection below (lines 1696-1702) is not gated on mocking, so it still writes `user_id`, `session_id` and `metadata` into `merged_input` before `tool.run(...)`.

`SubAgentToolInputSchema` is `extra="allow"` and declares no `is_accessible_to_agent=False` fields, so those keys survive `validate_input_schema` and the `strip_inaccessible_fields` guard in `Node._build_mocked_output` removes nothing. `describe_skipped_call` then renders the whole input dict into `{"content": ...}`, which is returned to the parent agent and appended to its transcript and trace.

A real run never does this: the sub-agent consumes those values and only its own output reaches the parent. That breaks the invariant the guard in `_build_mocked_output` states — "In a real run these values go to the tool and never reach the transcript; mocking must not change that" — and puts caller-supplied run `metadata` into the parent model's prompt. Stripping the injected context keys from the description on the mocked path, or skipping the injection when `is_delegation_mocked`, closes it.

Separately on the same path, `_build_child_agent_context(resolved_agent)` receives `None` here, so the injected ids are suffixed `:subagent` rather than with the sub-agent's name — the mocked values differ from what a real delegation would carry.

---

vitalii-dynamiq (commented):
Reviewed 4595717: 2 medium findings.

---

dynamiq/flows/flow.py
```diff
@@ -406,6 +434,7 @@ def run_sync(
             # instead of overwriting the loaded one in-place.
             self._checkpoint_persisted = True
             self._restore_from_checkpoint(self._checkpoint)
+            config = self._restore_config_from_checkpoint(config, self._checkpoint)
```
vitalii-dynamiq, line 437: ### The restore only survives one resume: a checkpoint written during a resumed run keeps the *first* run's config

**Medium severity**

`original_config` is written only on the fresh-run branch (line 474, and 678 in `run_async`). The resume branch above it updates `run_id`, `wf_run_id` and `status` and leaves `original_config` untouched, so every snapshot appended during a resumed run still carries the config of the run that first created the checkpoint. `_restore_config_from_checkpoint` reads that stale dict, so the second resume restores the wrong config.

Chain: real run interrupted mid-flow → resume with `mock=RunMockConfig(policy=MockPolicy.ALL)`, which correctly mocks the pending nodes → that resumed run is itself interrupted → resume its snapshot with no `mock=` → `stored.get("mock")` is `None` (inherited from run 1) → nothing is restored → the remaining nodes execute for real.

Reproduced against HEAD with the fixtures from `tests/integration/checkpoints/test_resume_restores_run_config.py`:

```
run2 (dry run) executed: []
cp2 original_config.mock = None          # snapshot taken *during* the dry run
run3 (resume of the dry run, no config) executed: ['third']
```

`third` really ran. This is the same failure the module docstring describes as fixed — "a dry run resumed from a mid-flow snapshot without `mock=` executed its remaining mocked nodes for real" — just one resume deeper, and it is symmetric: a real run resumed from a snapshot appended after a `MockPolicy.ALL` resume silently becomes mocked. The four tests in that file all resume exactly once, so none of them reach it.

Refreshing `original_config` on the resume branch (alongside `run_id`/`wf_run_id`/`status`) makes the checkpoint describe the run that is actually writing it, in both `run_sync` and `run_async`.

---

vitalii-dynamiq (commented):
Reviewed 5d950984720cd93506f65fc228a513a9816821f6: 1 medium finding.

---

dynamiq/nodes/tools/mcp.py
```diff
@@ -534,18 +571,29 @@ async def initialize_tools(self):
...
                             server_metadata=ServerMetadata(id=self.id, name=self.name, description=self.description),
                             is_optimized_for_agents=self.is_optimized_for_agents,
                         )
+                        mcp_tool._owner_server = self
+                        self._mcp_tools[tool.name] = mcp_tool
 
             logger.info(f"Tool {self.name}: {len(self._mcp_tools)} MCP tools initialized from a server.")
         except Exception as e:
             error = format_exception(e)
+            if self.mock.enabled:
```
vitalii-dynamiq: ### An unreachable server degrades silently on runs that will execute its tools for real

**Medium severity**

The degradation branch keys off `self.mock.enabled`, but `enabled` alone does not mean this server will be mocked at run time — only `is_pinned` (`enabled and locked`) survives a run-level override. Discovery happens in `Agent.__init__` (`agents/base.py:416`), before any `RunnableConfig` exists, so an enabled-but-unlocked mock swallows the transport failure for runs that were explicitly asked to execute for real.

Chain: server configured `mock=MockConfig(enabled=True)`, `locked` left at its `False` default. The server is unreachable, so `initialize_tools` logs a warning and returns; `_mcp_tools` stays empty and `expanded_tools` receives nothing. The agent is then run with `RunnableConfig(mock=RunMockConfig(policy=MockPolicy.NONE))` — under that policy `RunMockConfig.resolve` takes the `MockPolicy.NONE` branch, `node_mock.is_pinned` is `False`, and it returns `None`, i.e. these tools were meant to execute for real this run. They are absent instead, with no error anywhere in the run: the agent proceeds tool-less and answers from what remains. The same holds for `exclude_ids={server.id}` under the default `MockPolicy.NODE` — the exclusion means "this one must really run", and it silently does not.

The identical server with mocking off fails loudly at construction with `ToolExecutionException`, which is the behaviour the module docstring promises for a real run.

Gating on `self.mock.is_pinned` rather than `self.mock.enabled` closes it and matches the intent already stated in the PR description — "a canvas with a *pinned-off* MCP server loads and runs without connectivity to it". A pinned server can never be turned back on by a policy, so its tools genuinely cannot be needed; an unlocked one can.

---

vitalii-dynamiq (commented):
Reviewed 04eb523: 1 medium finding.

---

dynamiq/nodes/tools/mcp.py
tyaroshko: Fixed in a8104c51 — gated on `self.mock.is_pinned` rather than `self.mock.enabled`, exactly as suggested.

The reasoning is now recorded at the branch: discovery runs at agent construction, before any `RunnableConfig` exists, so this cannot ask whether *this run* will mock the server. Only a pin survives every run-level policy, so only a pinned server's tools can never be needed for real — an unlocked mock is still turned off by `MockPolicy.NONE` or an exclusion, and swallowing the transport failure there leaves that run silently tool-less.

Regression test added: `TestUnreachableMCPServer::test_an_enabled_but_unlocked_unreachable_server_still_fails_loudly`, verified failing against the previous gate. The existing `test_a_mocked_unreachable_server_degrades_instead_of_raising` already used `locked=True`, so it still passes unchanged — the fix narrows the degradation to exactly the pinned case the PR description promised.

The "Known limits" section of the module docstring was promising the old behaviour and has been updated to say *pinned* rather than *mocked*.

mykhailobuleshnyi (approved):

TrachukT (approved):

PR #234, opened 2025-04-11
feat: process failed/skipped node dependencies

acoola:
> [!IMPORTANT]
> Enhance error handling and dependency management in `dynamiq` by introducing `RunnableResultError` and supporting conditional dependencies.
> 
>   - **Error Handling**:
>     - Introduce `RunnableResultError` in `runnables/base.py` to encapsulate error details, including type, message, and recoverability.
>     - Update `complete_nodes()` in `executors/pool.py` to use `RunnableResultError` for logging execution errors.
>     - Modify `run_sync()` and `run_async()` in `nodes/node.py` to handle `RunnableResultError` for failed executions.
>   - **Dependency Management**:
>     - Add `condition` attribute to `NodeDependency` in `nodes/node.py` to support conditional dependencies.
>     - Implement `_validate_dependency_condition()` in `nodes/node.py` to evaluate conditions using `ChoiceCondition`.
>     - Update `validate_depends()` in `nodes/node.py` to check conditions and options for dependencies.
>   - **Node Execution**:
>     - Modify `execute()` methods in various orchestrators (e.g., `graph.py`, `linear.py`) to handle errors using `RunnableResultError`.
>     - Update `call_llm()` in `nodes/converters/llm_text_extractor.py` and `nodes/rankers/llm.py` to log errors using `RunnableResultError`.
>   - **Testing**:
>     - Add tests in `test_flow.py` to verify behavior with conditional dependencies and error handling.
>     - Update `test_exa.py` to check for error handling in invalid input scenarios.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=dynamiq-ai%2Fdynamiq&utm_source=github&utm_medium=referral)<sup> for d852fd9f3f65422a2d2cb73aec3e9b3175ff168e. It will automatically update as commits are pushed.</sup>

PR #890, opened 2026-08-12
feat(nodes): let the workflow Output node survive untaken branches

vitalii-dynamiq:
## The problem

A workflow that splits on a `Choice` and rejoins on one Output node returns **nothing at all — on both branches** — while still reporting success.

```
input → prompt-injection-detector → choice ─┬─ option-1 → error   ─┐
                                            └─ default  → agent   ─┴→ output
```

Measured on `main`:

```
detected=False   agent=success  error=skip     output=SKIP  value=None
detected=True    agent=skip     error=success  output=SKIP  value=None
```

Two independent causes:

1. **Reachability.** Exactly one branch runs, so the other is `SKIP`. `_validate_dependency_status` (`dynamiq/nodes/node.py:482`) propagates that skip into the Output node, which depends on both. The Output node is skipped no matter which branch ran.
2. **Binding.** The field is bound to one source (`$.agent.output.content`), which resolves to `None` on the error branch.

Fixing either alone still leaves the workflow broken.

It also fails silently: `_get_failed_nodes_with_raise_behavior` (`flow.py:216-218`) counts only `FAILURE`, never `SKIP`, so the flow returns `RunnableStatus.SUCCESS` with `error: None`.

## The change

**`Output.validate_depends`** ignores a `SKIP` dependency as long as at least one dependency succeeded — Airflow's `none_failed_min_one_success`, scoped to the node class that means "this is the workflow's result".

Only the dependency's own `SKIP` status is tolerated. Per-option and per-condition gates are still evaluated on every dependency, so conditional branching stays strict. With a single dependency there is no successful sibling, so skip propagation is unchanged.

This applies the principle already established in #234 — that a skip should not always propagate — at the join where it is always wanted, instead of requiring `behavior: return` to be set on the upstream branch nodes (which is both the wrong node and unreachable from the Output panel).

**`jsonpath.mapper`** accepts `||` as an ordered coalesce, so one field can bind to whichever branch ran:

```yaml
input_transformer:
  selector:
    output: "$.agent.output.content || $.error.output.output || no result"
```

First alternative that resolves wins; a non-JSONPath tail is a literal default. Selectors without `||` take the identical code path as before. The value stays a single string, so the YAML and the API wire format are unchanged.

## Why not the built-in JSONPath union

`jsonpath-ng` does document `|`, and it works if each operand is parenthesized — `($.a.output.x) | ($.b.output.y)` — verified on the pinned `1.6.1`. It was rejected because when neither alternative resolves, `mapper`'s `use_expression_as_value` branch fires (the string starts with `(`, not `$`) and the **JSONPath expression itself is returned as the field's value**. Given the bug being fixed here is silent wrong data, that trade was not worth 32 lines. Union also cannot express ordering or a default.

No dependency bump: the precedence behaviour is identical in `1.6.1` and `1.8.0`, and no release has a coalesce or default operator.

## Guardrail regression — the reason for the narrow wording

The shipped `prompt-injection-detector` template wires the agent with a **mixed dependency**:

```
depends: [ {node: choice, option: '264f2233…'}, {node: input} ]
```

A broader rule ("if any dependency was taken, run") sees the ungated `input` edge as taken, excuses the not-taken `choice` option, and **the agent runs on the blocked input**. Measured against that exact topology:

| rule | injection detected | react-agent | output |
|---|---|---|---|
| current `main` | yes | skip | **skip · null** |
| **this PR** | yes | **skip** ✅ | `"BLOCKED…"` |
| broad variant | yes | **success** ⚠️ | `["AGENT ANSWER", "BLOCKED…"]` |

Covered by `test_choice_option_gate_is_still_enforced_with_extra_dependency`.

## Validation

- **1742 existing tests pass, no regressions.** Two failures (`test_e2b_sdk_compat`, `test_file_read_tool_limits_spreadsheet_preview`) fail identically with this patch reverted — pre-existing environment issues.
- **25 new tests.** In `tests/integration/workflows/test_branch_convergence.py`, reverting the patch makes **3 fail** (they pin the bug) and **6 pass** (they pin behaviour that must not change: untaken branch still skipped, single-dependency skip still propagates, all-skipped output still skips, upstream failure still fails the flow and still names the failed node, parallel fan-in untouched).
- **YAML round-trip 9/9.** Dump → reload → execute is stable; `||` survives `OmegaConf` verbatim and parses back as a single string, so Go's `map[string]string` still holds. A hand-authored YAML in the shape the Go layer emits, including a `|| no result` default, loads and runs.
- Run on `jsonpath-ng 1.6.1`, the version pinned in `pyproject.toml`.

## Scope

Deliberately scoped to `Output`. The same rule on the base `Node` class would also fix mid-graph convergence (branch → branch → shared node → output) and the guardrail holds there too, but a side-effecting node sitting at a join that is skipped today would begin firing. That is a product decision, best made separately.

The builder half is [dynamiq-ai/ui#claude/workflow-output-node-design-cyyoj1](https://github.com/dynamiq-ai/ui/pull/new/claude/workflow-output-node-design-cyyoj1). This PR stands alone for YAML- and API-authored workflows.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HByYssKX3amPQYdkn6bJZR

---
_Generated by [Claude Code](https://claude.ai/code/session_01HByYssKX3amPQYdkn6bJZR)_

### DIFF
--- a/dynamiq/nodes/utils/utils.py
+++ b/dynamiq/nodes/utils/utils.py
@@ -4,6 +4,7 @@ from pydantic import Field
 
 from dynamiq.nodes import NodeGroup
 from dynamiq.nodes.operators import Pass
+from dynamiq.runnables import RunnableResult, RunnableStatus
 
 
 class Input(Pass):
@@ -65,3 +66,31 @@ class Output(Pass):
     """,
     )
     _json_schema_fields: ClassVar[list[str]] = ["json_schema"]
+
+    def validate_depends(self, depends_result: dict[str, RunnableResult]) -> None:
+        """Validate dependencies, tolerating branches that were not taken.
+
+        A workflow output is a join point: when a Choice sends execution down one branch,
+        the other branch is SKIPPED, and the default rule would skip the output node too.
+        Here a SKIPPED dependency is ignored as long as at least one dependency succeeded.
+
+        Only the dependency's own SKIP status is tolerated; per-option and per-condition
+        gates are still evaluated on every dependency, so conditional branching is
+        unaffected. With a single dependency there is no successful sibling, so skip
+        propagation behaves exactly as before.
+        """
+        tolerate_skips = any(
+            depends_result[dep.node.id].status == RunnableStatus.SUCCESS
+            for dep in self.depends
+            if dep.node.id in depends_result
+        )
+
+        for dep in self.depends:
+            dep_result = depends_result.get(dep.node.id)
+            if tolerate_skips and dep_result is not None and dep_result.status == RunnableStatus.SKIP:
+                continue
+            self._validate_dependency_status(depend=dep, depends_result=depends_result)
+            if dep.condition:
+                self._validate_dependency_condition(depend=dep, depends_result=depends_result)
+            if dep.option:
+                self._validate_dependency_option(depend=dep, depends_result=depends_result)
--- a/dynamiq/nodes/utils/utils.py
+++ b/dynamiq/nodes/utils/utils.py
@@ -67,27 +67,55 @@ class Output(Pass):
     )
     _json_schema_fields: ClassVar[list[str]] = ["json_schema"]
 
+    @staticmethod
+    def _is_branch_gated(node, seen: set[str] | None = None) -> bool:
+        """Whether `node` sits downstream of a Choice option or a dependency condition.
+
+        Only such a node can legitimately be SKIPPED because a branch was not taken. A
+        node with no gate anywhere in its ancestry always runs, so its success says
+        nothing about which branch executed and must not license tolerating a skip.
+        """
+        seen = seen if seen is not None else set()
+        if node.id in seen:
+            return False
+        seen.add(node.id)
+
+        return any(
+            dep.option or dep.condition or Output._is_branch_gated(dep.node, seen)
+            for dep in getattr(node, "depends", [])
+        )
+
     def validate_depends(self, depends_result: dict[str, RunnableResult]) -> None:
         """Validate dependencies, tolerating branches that were not taken.
 
         A workflow output is a join point: when a Choice sends execution down one branch,
         the other branch is SKIPPED, and the default rule would skip the output node too.
-        Here a SKIPPED dependency is ignored as long as at least one dependency succeeded.
+        Here a SKIPPED dependency is ignored as long as a *branch-gated* dependency
+        succeeded -- i.e. some branch demonstrably ran.
+
+        The scan is restricted to branch-gated dependencies on purpose. An Output node
+        often also depends on an ungated node such as `input`, which always succeeds;
+        counting it would make skip tolerance unconditional and let an Output whose every
+        branch was skipped report success carrying nulls.
 
         Only the dependency's own SKIP status is tolerated; per-option and per-condition
         gates are still evaluated on every dependency, so conditional branching is
         unaffected. With a single dependency there is no successful sibling, so skip
         propagation behaves exactly as before.
         """
+        gated = {dep.node.id for dep in self.depends if self._is_branch_gated(dep.node)}
         tolerate_skips = any(
-            depends_result[dep.node.id].status == RunnableStatus.SUCCESS
-            for dep in self.depends
-            if dep.node.id in depends_result
+            depends_result[node_id].status == RunnableStatus.SUCCESS for node_id in gated if node_id in depends_result
         )
 
         for dep in self.depends:
             dep_result = depends_result.get(dep.node.id)
-            if tolerate_skips and dep_result is not None and dep_result.status == RunnableStatus.SKIP:
+            if (
+                tolerate_skips
+                and dep.node.id in gated
+                and dep_result is not None
+                and dep_result.status == RunnableStatus.SKIP
+            ):
                 continue
             self._validate_dependency_status(depend=dep, depends_result=depends_result)
             if dep.condition:
--- a/dynamiq/utils/jsonpath.py
+++ b/dynamiq/utils/jsonpath.py
@@ -22,27 +22,50 @@ def is_jsonpath(path: str) -> bool:
         return False
 
 
-def _coalesce(json: dict | list, path: str):
+def is_expression(path: str, expression_prefixes: tuple = JSONPATH_EXPRESSION_PREFIXES) -> bool:
+    """
+    Check whether a string is a rooted JSONPath expression rather than a literal value.
+
+    `is_jsonpath` alone is not enough here: a bare word such as "a" parses as a valid
+    JSONPath, so plain text would otherwise be mistaken for a path.
+
+    Args:
+        path (str): The string to be checked.
+        expression_prefixes (tuple): The prefixes a JSONPath expression must start with.
+
+    Returns:
+        bool: True if the string is a rooted JSONPath expression, False otherwise.
+    """
+    if not isinstance(path, str) or not path.startswith(expression_prefixes):
+        return False
+
+    try:
+        return is_jsonpath(path)
+    except Exception:
+        return False
+
+
+def _coalesce(json: dict | list, candidates: list[str], expression_prefixes: tuple):
     """
     Resolve an ordered list of alternatives, e.g. "$.a.output.x || $.b.output.y".
 
     Returns the value of the first candidate that resolves to something, or None.
-    A non-JSONPath candidate is treated as a literal default, so a trailing
+    A candidate that is not a rooted JSONPath is a literal default, so a trailing
     "|| no result" always yields a populated field. Used when branches of a Choice
     converge on a single field and only one of them produced a value.
 
     Args:
         json (dict | list): The input JSON object to resolve against.
-        path (str): Candidates separated by COALESCE_SEPARATOR.
+        candidates (list[str]): Alternatives, already split and stripped.
+        expression_prefixes (tuple): The prefixes of the JSONPath expressions.
 
     Returns:
         The first resolved value, or None if nothing resolved.
     """
-    for candidate in (c.strip() for c in path.split(COALESCE_SEPARATOR)):
-        if not candidate:
-            continue
-        if not is_jsonpath(candidate):
+    for candidate in candidates:
+        if not is_expression(candidate, expression_prefixes):
             return candidate
+
         found = [match for match in parse(candidate).find(json) if match.value is not None]
         if found:
             return found[0].value if len(found) == 1 else [match.value for match in found]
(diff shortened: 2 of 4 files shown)


dynamiq/utils/jsonpath.py
```diff
@@ -52,6 +104,14 @@ def mapper(
 
     new_json = {}
     for key, path in expression_map.items():
+        if isinstance(path, str) and COALESCE_SEPARATOR in path:
+            candidates = [c.strip() for c in path.split(COALESCE_SEPARATOR) if c.strip()]
+            # "||" only means "try these in order" when at least one alternative is a
+            # rooted JSONPath. Otherwise the value is ordinary text that happens to
+            # contain the characters, and must be preserved as-is.
+            if any(is_expression(candidate, expression_prefixes) for candidate in candidates):
```
vitalii-dynamiq: ### Literal-text guard does not cover `@`-prefixed alternatives

**Low severity**

This guard is meant to keep ordinary text that happens to contain `||` from being coalesced, but `is_expression` accepts `@` as a rooting prefix and `@channel` parses as a valid JSONPath. So a selector value like `"@channel || @here"` passes the guard, `_coalesce` runs, neither alternative matches anything in the input, and the field becomes `None`. On the base commit the same string failed `is_jsonpath` as a whole and was preserved verbatim.

The same hole applies to the literal-default form this PR documents: `"$.agent.output.content || @channel"` resolves to `None`, not `"@channel"`, when the agent branch was skipped — the documented "non-JSONPath tail is a literal default" silently fails for any default starting with `@`.

Reproduced against the venv on this HEAD: `mapper({'a': {'b': 1}}, {'k': '@channel || @here'})` returns `{'k': None}`; the same call against `dynamiq/utils/jsonpath.py` at `9fcec5f` returns `{'k': '@channel || @here'}`.

---

vitalii-dynamiq (commented):
Reviewed 30a4530db80da2b98b6115e9aa92dbf163c20b2a — 1 finding (0 critical, 0 high, 0 medium, 1 low).

---

dynamiq/utils/jsonpath.py
```diff
@@ -21,6 +22,57 @@ def is_jsonpath(path: str) -> bool:
...
+    Args:
+        json (dict | list): The input JSON object to resolve against.
+        candidates (list[str]): Alternatives, already split and stripped.
+        expression_prefixes (tuple): The prefixes of the JSONPath expressions.
+
+    Returns:
+        The first resolved value, or None if nothing resolved.
+    """
+    for candidate in candidates:
+        if not is_expression(candidate, expression_prefixes):
```
vitalii-dynamiq: ### An unparseable rooted alternative aborts the chain and leaks as the field value

**Medium severity**

`is_expression` returns `False` for two different things: a candidate that is not rooted (a literal default), and a candidate that *is* rooted but fails `jsonpath_ng.parse`. This branch treats both as a literal default, so an unparseable `$…` alternative is returned verbatim as the field's value and every remaining alternative — including an explicit `|| no result` tail — is never tried.

The dominant reachable case is a node id that `jsonpath-ng` cannot lex. Node ids default to `generate_uuid()` (`dynamiq/nodes/node.py:314`), and a UUID beginning with a decimal digit does not parse:

```python
is_jsonpath("$.264f2233-1f0a-4b21-8d1e-000000000000.output.content")  # False
is_jsonpath("$.ab4f2233-1f0a-4b21-8d1e-000000000000.output.content")  # True
```

Reproduced on this HEAD, with both branch nodes present in the input and the error branch holding the real value:

```python
mapper(data, {"output": "$.264f2233-….output.content"
                        " || $.ab4f2233-….output.output"
                        " || no result"})
# {'output': '$.264f2233-1f0a-4b21-8d1e-000000000000.output.content'}
```

The field is populated with the raw selector text rather than `BLOCKED`, so the Output node reports success carrying an expression string as its payload — the same "expression leaks as the value" failure the PR rejected the JSONPath union operator to avoid. A rooted-but-unparseable candidate should be skipped so the next alternative runs, not returned as a default.

---

dynamiq/utils/jsonpath.py
```diff
@@ -21,6 +22,57 @@ def is_jsonpath(path: str) -> bool:
...
+            return candidate
+
+        found = [match for match in parse(candidate).find(json) if match.value is not None]
```
vitalii-dynamiq: ### Coalesced multi-match selectors drop `None` elements and flip list to scalar

**Medium severity**

The `None` filter here is applied before the result is shaped, so for a candidate matching more than one location the surviving matches decide both the contents *and* the list-vs-scalar form. The plain path at lines 119-128 does no such filtering, so the same selector yields a different type and a different number of elements depending only on whether a `||` appears elsewhere in the string.

`test_coalesce_returns_list_when_one_alternative_matches_many` establishes that multi-match alternatives are supported usage, so this is reachable inside the feature as designed. Reproduced on this HEAD:

```python
d = {"items": [{"v": None}, {"v": 2}]}
mapper(d, {"k": "$.items[*].v"})               # {'k': [None, 2]}
mapper(d, {"k": "$.items[*].v || fallback"})   # {'k': 2}     <- list collapsed to scalar

d = {"items": [{"v": 1}, {"v": None}, {"v": 2}]}
mapper(d, {"k": "$.items[*].v"})               # {'k': [1, None, 2]}
mapper(d, {"k": "$.items[*].v || fallback"})   # {'k': [1, 2]} <- element silently dropped
```

A downstream node bound to such a field receives `2` where it expects a list, or a list whose positions no longer line up with the collection it was derived from. The emptiness test needs the `None` filter; the returned value should be built from the unfiltered matches.

---

vitalii-dynamiq (commented):
Reviewed 81bb0e5277e0235b9f9c25c9491148f3e3ae0c7d — 2 medium findings.

---

dynamiq/nodes/utils/utils.py
```diff
@@ -65,3 +66,31 @@ class Output(Pass):
...
+        A workflow output is a join point: when a Choice sends execution down one branch,
+        the other branch is SKIPPED, and the default rule would skip the output node too.
+        Here a SKIPPED dependency is ignored as long as at least one dependency succeeded.
+
+        Only the dependency's own SKIP status is tolerated; per-option and per-condition
+        gates are still evaluated on every dependency, so conditional branching is
+        unaffected. With a single dependency there is no successful sibling, so skip
+        propagation behaves exactly as before.
+        """
+        tolerate_skips = any(
```
vitalii-dynamiq, line 107: ### An ungated dependency makes skip-tolerance unconditional, so an all-skipped Output reports success with a null payload

**Medium severity**

`tolerate_skips` is computed over *every* dependency, including ones that are not gated on a branch and therefore always run. When the Output node depends on any such node — `input` being the obvious one — the `any(...)` is permanently `True`, so every skipped dependency is `continue`d regardless of how many branches actually executed. The "at least one dependency succeeded" safeguard degenerates into "always tolerate", and the invariant pinned by `test_output_skips_when_every_dependency_is_skipped` holds only for Outputs whose dependencies are *all* branch-gated.

The reachable topology is an ordinary one: a conditional sub-chain that may not run at all, plus an ungated edge because the selector also wants a field from the original input. Reproduced on this HEAD with `input → choice -{go}→ branch`, `output` depending on `[branch, input]`, and `go=False`:

```python
# selector={"answer": "$.branch.output.v", "question": "$.input.output.go"}
branch: skip
output: success {'answer': None, 'question': False}
```

Dropping the `input` edge from `output.depends` — the same graph otherwise — correctly yields `output: skip`, as `test_single_dependency_skip_still_propagates` asserts.

So the workflow returns `RunnableStatus.SUCCESS` carrying `answer: None` when no branch produced anything. Because `_get_failed_nodes_with_raise_behavior` never counts `SKIP`, nothing downstream flags it — this is the same silent success-with-null-data failure the PR was opened to remove, reintroduced for Outputs that carry an ungated dependency. Restricting the successful-sibling scan to dependencies that could have been skipped (branch-gated ones) keeps the join fix while preserving the all-skipped case.

---

vitalii-dynamiq (commented):
Reviewed 3c7699a: 1 medium-severity finding.

---

dynamiq/utils/jsonpath.py
tyaroshko: Confirmed and fixed in d6aed1ff.

`is_expression` was conflating two different things: a candidate that is not rooted (a literal default) and one that is rooted but fails to parse. Split into `_is_rooted` (shape only) and `is_expression` (shape + parses). `_coalesce` now returns a candidate as a literal only when it is not rooted; a rooted candidate that fails to parse is skipped so the remaining alternatives still get their turn. The guard in `mapper` moved to `_is_rooted` for the same reason — otherwise the whole selector fell through to the plain path and leaked there instead.

Your repro now returns `{"output": "BLOCKED"}`, and `$.264f… || no result` returns `no result`.

Covered by `test_unparseable_rooted_alternative_does_not_abort_the_chain`, `..._falls_through_to_the_literal_default`, `..._never_leaks_as_the_value` (parametrized over three forms), and `test_coalesce_is_a_no_op_when_only_unparseable_paths_are_offered`.

One case deliberately left alone: a plain, non-coalesced unparseable selector still returns the raw expression as its value. That is pre-existing on `main` (verified by stashing) and out of scope for this PR, so it is noted in a test docstring rather than changed here.

dynamiq/utils/jsonpath.py
tyaroshko: Confirmed and fixed in d6aed1ff.

The `None` filter now only decides *whether* a candidate resolved; the returned value is built from the unfiltered matches, so shape and length match the plain path exactly.

Both repros now agree with the non-coalesced selector: `[None, 2]` and `[1, None, 2]`.

Pinned by `test_multi_match_shape_matches_the_plain_path`, parametrized over four collections and asserting equality against `mapper(data, {"k": "$.items[*].v"})` rather than hardcoded values, so the two paths cannot drift apart again. `test_all_null_multi_match_falls_through` keeps the emptiness semantics you noted are still needed.

dynamiq/nodes/utils/utils.py
tyaroshko: Confirmed and fixed in d6aed1ff. Reproduced exactly as described: with the `input` edge, `output: success {"answer": None, "question": False}`; dropping it, `output: skip`.

`tolerate_skips` now scans only branch-gated dependencies, as you suggested. The one wrinkle is that gating cannot be read off the Output’s own edges — `NodeDependency(agent)` carries no `option`, since the gate sits on the branch node’s edge to the `Choice`. So gating is inherited through the depends graph: `_is_branch_gated` walks a node’s ancestry for any edge with an `option` or `condition`, with a visited set for cycles. An ungated dependency such as `input` is excluded from both the successful-sibling scan and the tolerated set.

Covered by `test_ungated_dependency_does_not_make_skip_tolerance_unconditional` (both topologies, asserts `skip`), `test_ungated_dependency_still_allows_the_taken_branch_through` (the `go=True` path still succeeds, so the fix the PR exists for is intact), and `test_skip_tolerance_survives_an_intermediate_ungated_hop` (gating inherited through a plain intermediate node between the branch and the Output).

mykhailobuleshnyi (approved):

PR #902, opened 2026-08-18
fix(agents): execute every requested tool call and pair results by id

mykhailobuleshnyi:
Six defects in the agent's function-calling loop, each with a regression test verified to fail without its fix.

Tool calls were being discarded. With parallel_tool_calls_enabled=False -- the default -- a step containing several tool calls executed the first and dropped the rest. The loss was silent: the extra calls were erased before the assistant message was written, so the transcript showed one call requested and one answered, the provider raised nothing, and the model could not tell work had gone missing. This is not an edge case -- we never ask the provider to disable parallel calls, so a model that decides to batch will do so. The flag now means "do not run these concurrently" rather than "throw the extras away": every call runs, one after another, in the order requested.

Results were paired to calls by position. Nothing in a result identified which call it answered; the link held only because four separate code paths happened to agree on ordering, across a thread pool that returns in completion order. The provider's tool_call_id now travels with the call -- through ToolCallItem, as a SkipJsonSchema field so it stays out of the prompt the model sees -- into each result, and replies are matched by that id. A result arriving without one now costs its call the content rather than handing it another tool's output.

Requested calls could go unanswered, which kills the run. A request carrying an unanswered tool_call_id is rejected outright, discarding every completed loop. Two paths that end a batch early -- a refused sub-agent budget, a batch that failed to parse -- wrote a bare user-role observation and left their ids owing. Every pending id is now answered in one place.

A final answer batched alongside tool calls was returned as though those results had informed it. They cannot have: the model wrote the answer in the same step that requested them. Anchoring on the first call made the outcome depend on ordering that carries no meaning -- a batch placing the answer last had it discarded and looped to max_loops, one placing it first returned a guess and threw the tools away. An answer is now accepted only when nothing else was requested; batched with tools, the tools run and the reply to the declined call says why. The trade-off is deliberate: a model that attaches an answer to the same repeated calls now exhausts max_loops instead of returning a guess.

Structured-output mode crashed the run on valid-but-unexpected JSON. `null`, a bare list, or a quoted string all parse cleanly and then failed a membership test with TypeError; a response omitting `action_input` failed the read below it with KeyError. Neither is recoverable, so both ended the run. They are parse failures like any other and are now reported as such, naming what is missing.

Tool concurrency was unbounded -- the number of threads spawned was whatever the model asked for. max_parallel_tool_calls (default 8) bounds the pool; calls beyond the limit queue rather than being dropped.

Tool failures were indistinguishable from successes in function-calling mode. The status is computed either way and used to build SUCCESS/ERROR headers for the other inference modes, but the function-calling path reads the per-call results and never those headers. Failures now carry a short marker; successful results are unchanged byte-for-byte.

Also: `run-parallel` is no longer injected into an agent constructed with no tools, where it advertised a tool that could never run, along with the whole tools prompt block.

Behaviour change to note: a function-calling agent with parallel_tool_calls_enabled=False now performs N tool executions per step where it previously performed one. That is work the model requested, but it is a real cost and latency change.

Not included: defaulting parallel_tool_calls_enabled by inference mode. Enabling it outside FUNCTION_CALLING injects `run-parallel`, and workflow schema generation requires every tool to expose a `connection` field, which that tool lacks (node_generation.py:127). That is reachable today by setting the flag explicitly, and should be fixed separately.

### DIFF
--- a/dynamiq/nodes/agents/agent.py
+++ b/dynamiq/nodes/agents/agent.py
@@ -610,22 +610,6 @@ class Agent(HistoryManagerMixin, BaseAgent):
 
         return self
 
-    def _acknowledge_pending_fc_tool_calls(self, content: str) -> None:
-        """Answer and clear any pending FC tool_call ids with a neutral stub.
-
-        Keeps OpenAI's function-calling protocol valid when tool calls are
-        abandoned without execution (e.g. the model returned ``provide_final_answer``
-        alongside other tool calls), so no unpaired ``tool_call`` is left to 400 a
-        subsequent request and recovery for unrelated errors is not misdirected to
-        these ids.
-        """
-        pending_ids = getattr(self, "_pending_fc_tool_call_ids", None) or []
-        for tc_id in pending_ids:
-            self._prompt.messages.append(
-                Message(role=MessageRole.TOOL, content=content, tool_call_id=tc_id, static=True)
-            )
-        self._pending_fc_tool_call_ids = []
-
     def _append_recovery_instruction(
         self,
         *,
@@ -890,10 +874,8 @@ class Agent(HistoryManagerMixin, BaseAgent):
                                 static=True,
                             )
                         )
-                        # An answer batched with tool calls was written before those results
-                        # existed, so it cannot be informed by them and is not accepted.
-                        # Saying so is what lets the model answer for real on the next step
-                        # instead of re-emitting the same batch.
+                        # Declined when batched with tools: the answer predates their
+                        # results. Saying so is what stops the model re-emitting the batch.
                         final_answer_reply = (
                             "Not accepted: you requested tools in this same step, so this answer was "
                             "written before their results existed. Their results follow — answer "
@@ -996,7 +978,6 @@ class Agent(HistoryManagerMixin, BaseAgent):
         final_answer_call = next((tc for tc in tool_calls if tc.function.name.strip() == "provide_final_answer"), None)
 
         if final_answer_call is not None and not actual_tool_calls:
-            self._acknowledge_pending_fc_tool_calls("Skipped — superseded by a final-answer call.")
             final_args = final_answer_call.function.parse_as_final_answer()
             thought = final_args.thought
             self._requested_output_files = self._parse_output_files_csv(final_args.output_files)
@@ -1524,7 +1505,6 @@ class Agent(HistoryManagerMixin, BaseAgent):
         self,
         tool_result: Any,
         ordered_results: list[dict[str, Any]] | None = None,
-        tool_name: str | None = None,
         success: Any = None,
     ) -> None:
         """Append tool observations to prompt history.
@@ -1566,7 +1546,7 @@ class Agent(HistoryManagerMixin, BaseAgent):
                             else mark_tool_failure(str(tool_result), success)
                         ),
                         tool_call_id=tc_id,
-                        name=result.get("tool_name") if result else tool_name,
+                        name=result.get("tool_name") if result else None,
                         static=True,
                     )
                 )
@@ -1595,7 +1575,7 @@ class Agent(HistoryManagerMixin, BaseAgent):
             logger.error(error_message)
             # Emitted, not added as a bare observation: the batch never ran, so every id
             # in it is answered with the parse error that applies to all of them.
-            self._emit_tool_observations(error_message, tool_name=PARALLEL_TOOL_NAME, success=False)
+            self._emit_tool_observations(error_message, success=False)
             return None
 
     def _check_subagent_limits(self, tools_data: list[dict[str, Any]], action: str) -> str | None:
@@ -1723,7 +1703,7 @@ class Agent(HistoryManagerMixin, BaseAgent):
             if subagent_error:
                 # Emitted, not added as a bare observation: no call ran, so every id in
                 # the batch is answered with the refusal that applies to all of them.
-                self._emit_tool_observations(subagent_error, tool_name=action, success=False)
+                self._emit_tool_observations(subagent_error, success=False)
                 return None
 
             ordered_results: list[dict[str, Any]] = []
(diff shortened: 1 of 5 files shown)


dynamiq/nodes/agents/agent.py
```diff
@@ -1650,24 +1701,40 @@ def _execute_tools_and_update_prompt(
...
 
             ordered_results: list[dict[str, Any]] = []
-            if (
-                self.sanitize_tool_name(action) == PARALLEL_TOOL_NAME
-                and self.parallel_tool_calls_enabled
-                and not skip_parallel
-            ):
+            # Stays None on the batch path, where each result carries its own status.
+            tool_success: Any = None
+            if self.sanitize_tool_name(action) == PARALLEL_TOOL_NAME and not skip_parallel:
```
vitalii-dynamiq, line 1677: ### `delegate_final` is silently dropped once a step is routed through the batch path

**Medium severity**

Removing the `self.parallel_tool_calls_enabled` guard from this branch means a multi-call FUNCTION_CALLING step now reaches `_execute_tools` under the default `parallel_tool_calls_enabled=False`. Previously such a step fell through to the `else` branch, where `is_delegated` is checked and the delegated answer is returned (line 1720).

`_execute_tools` has no such check: `_execute_single_tool_to_result` unpacks `_execute_single_tool` as `tool_result, tool_files, _, success, dependency` (line 2427) and discards the delegation flag.

Failure chain: with `delegation_allowed=True`, a model batches a `SubAgentTool` call carrying `delegate_final: true` alongside any other tool call. `_execute_single_tool` takes the `if delegate_final:` branch (line 1383) — it calls `log_final_output` and, when streaming is enabled, emits the sub-agent's answer with `step="answer"` — then returns `True` as the third element. That `True` is thrown away here, so the delegated result is appended as an ordinary observation and the ReAct loop continues, producing a second final answer (or running to `max_loops`). The client receives two answers for one run, and the run's actual output is not the delegated one.

Before this change the same batch executed only the first call, so a delegating call in first position was honoured correctly; that path is now unreachable at the default flag value.

---

dynamiq/nodes/tools/parallel_tool_calls.py
```diff
@@ -22,6 +23,11 @@ class ToolCallItem(BaseModel):
         description="Input parameters for the tool as key-value pairs",
     )
     thought: str = Field(default="", description="Reasoning for this tool call.")
+    # The provider's id for this call, in FUNCTION_CALLING mode. Carried with the call so
+    # its result can be paired back to it by id rather than by position, which execution
+    # order does not preserve. SkipJsonSchema keeps it out of the schema shown to the
+    # model — the other inference modes author this object themselves and have no ids.
+    tool_call_id: SkipJsonSchema[str | None] = None
```
vitalii-dynamiq: ### `SkipJsonSchema` does not keep `tool_call_id` out of the prompt shown to the model

**Low severity**

The comment above states that `SkipJsonSchema` keeps this field out of the schema the model sees. The non-FUNCTION_CALLING prompt is not built from a JSON schema, so it does not.

`generate_input_formats` (`dynamiq/nodes/agents/components/schema_generator.py:141-145`) describes a `list[NestedModel]` field by enumerating `args[0].model_fields.items()` directly. `SkipJsonSchema` only annotates the field for JSON-schema generation; it does not remove it from `model_fields`. Verified against pydantic 2.12.5 — the `run-parallel` entry rendered into `input_formats` becomes:

```
list[{name: str - Name of the tool..., input: dict - ..., thought: str - ..., tool_call_id: str | None - }, ...]
```

`run-parallel` is only injected outside FUNCTION_CALLING mode (`base.py:449-453`), which is exactly where the field is meaningless: those modes author the object themselves and have no provider ids. So every agent that gets this tool now advertises a fourth, description-less parameter the model may populate with a fabricated value, which is then carried into `ordered_results` and never used.

The `is_accessible_to_agent` flag this function already honours is only consulted for top-level fields, not for the nested-model branch, so suppressing the field here needs a change on the generator side as well.

---

vitalii-dynamiq (commented):
Reviewed e4ffaaad819a183f215a4e756b35ed8e96f9dee4: 2 findings (1 medium, 1 low).

---

dynamiq/nodes/agents/base.py
```diff
@@ -439,7 +447,10 @@ def __init__(self, **kwargs):
             use_native_parallel = inference_mode == InferenceMode.FUNCTION_CALLING
             if not use_native_parallel:
                 self.tools = [t for t in self.tools if t.name != PARALLEL_TOOL_NAME]
-                self.tools.append(ParallelToolCallsTool())
+                # Nothing to batch without real tools -- injecting it would advertise a
+                # tool, and the whole tools prompt block, to an agent that has none.
+                if self.tools:
```
vitalii-dynamiq, line 455: ### The empty-tools guard runs before `SkillsTool` is added, so a skills-only agent loses `run-parallel`

**Low severity**

`self.tools` is not complete at this point. `_init_skills()` runs four lines below (line 456) and appends `SkillsTool` to `self.tools` (line 586). The guard is evaluated before that append, so it reads an empty list for an agent whose tool list is about to become non-empty.

Failure chain: construct an agent with `tools=[]`, `skills.enabled=True` and `skills.source` set, `parallel_tool_calls_enabled=True`, and any inference mode other than `FUNCTION_CALLING`. The `PARALLEL_TOOL_NAME` filter above is a no-op on the empty list, `if self.tools:` is False, and `ParallelToolCallsTool()` is not appended. `_init_skills()` then adds `SkillsTool`, and `_init_prompt_blocks()` (line 457) renders a tools block containing exactly one real, batchable tool and no `run-parallel`. The agent can only invoke skills one per loop; before this change it was given the batching tool.

The guard's stated intent — "nothing to batch without real tools" — does not hold here, because there is a real tool. Evaluating the injection after `_init_skills()` makes the check see the final tool list.

---

vitalii-dynamiq (commented):
Reviewed baf2b93: 1 low-severity finding.

---

TrachukT (dismissed):

dynamiq/nodes/agents/agent.py
```diff
@@ -2338,12 +2378,44 @@ def _execute_tools(
...
+                    {
+                        "order": tool_payload["order"],
+                        "tool_call_id": tool_payload.get("tool_call_id"),
+                        "tool_name": tool_payload["name"],
+                        "success": False,
+                        "result": refusal,
+                        "files": [],
+                    }
+                )
+            prepared_tools = executable
```
vitalii-dynamiq: ### Refused `delegate_final` calls are erased from the batch streaming events

**Low severity**

`prepared_tools` is rebound to `executable` here, dropping the refused entries, and both batch streaming calls below read that same rebound list — `_stream_batch_reasoning_event(prepared_tools, ...)` at 2448 and `_stream_batch_tool_result_event(batch_tool_run_id, prepared_tools, ...)` at 2516. The refusal only ever reaches `all_results`, which the stream never consults for membership.

Two consequences, both reachable with `delegation_allowed=True`, streaming enabled, and a model batching a `SubAgentTool` call carrying `delegate_final: true` alongside other tools:

Two calls, one refused → `prepared_tools` has length 1, so the `len(prepared_tools) == 1` branch at 2444 runs and neither batch event is emitted at all. The client sees an ordinary single-tool run; nothing in the stream indicates a second call was requested and declined, while the model's history does carry the refusal.

Three or more calls, one refused → `per_tool_summary` (759-789) is built by iterating `prepared_tools`, so the refused call has no entry, and `overall_status` is computed as `all(s.get("status") == RunnableStatus.SUCCESS for s in per_tool_summary)` over exactly that list. The refusal's `success: False` in `all_results` is not consulted, so the batch streams `RunnableStatus.SUCCESS` and an `input`/`result` array one element short of what the model requested.

Building the summary from `all_results` (which already carries `order`, `tool_name` and `success` for the refusals) rather than from the filtered `prepared_tools` keeps the streamed batch consistent with what was requested.

---

vitalii-dynamiq (commented):
Reviewed 1b47e00 — 1 low-severity finding.

---

TrachukT (approved):

PR #907, opened 2026-08-19
fix: file edit ambiguous anchors

mykhailobuleshnyi:


dynamiq/nodes/tools/file_tools.py
```diff
@@ -1363,25 +1423,51 @@ def _execute_edit(self, input_data: FileWriteInputSchema) -> dict[str, Any]:
...
                 skipped.append(edit.find)
                 continue
-            count = occurrences if edit.replace_all else min(1, occurrences)
+            if len(positions) > 1 and not edit.replace_all:
+                ambiguous.append((edit.find, len(positions), [_line_of(content, p) for p in positions]))
+                continue
             if edit.replace_all:
+                hits = _find_positions(content, edit.find, overlapping=False)
+                total += len(hits)
+                changed_lines.extend(_line_of(content, offset) for offset in hits)
```
vitalii-dynamiq: ### Reported "(lines ...)" are computed against intermediate content and never re-mapped to the written file

**Medium severity**

Each line number is recorded from the `content` snapshot as it exists *before* that replacement runs (here and at line 1445), but the summary at line 1470 presents them as locations in the file that was just written. Any replacement that changes the line count invalidates every number recorded before it, and within a single `replace_all` it invalidates every hit after the first.

Two reproductions against this code:

- `a\nb\n` with edits `[{find:'b', replace:'B'}, {find:'a', replace:'x\ny\na'}]` writes `x\ny\na\nB\n`. The real changes are on lines 3 and 4; the summary says `(lines 1, 2)`.
- `foo\nbar\nfoo\n` with `{find:'foo', replace:'A\nB', replace_all:true}` writes `A\nB\nbar\nA\nB\n`. The second replacement lands on lines 4–5; the summary says `(lines 1, 3)`.

Since the point of the new location string is to let the caller re-read a narrow window instead of the whole file, a caller acting on it reads the wrong region. Offsets need to be tracked and adjusted for the delta each replacement introduces (or resolved after all edits are applied) before being converted to line numbers.

---

dynamiq/nodes/tools/file_tools.py
```diff
@@ -227,11 +255,30 @@ class FileWriteAction(str, enum.Enum):
...
+    )
+
+
+def _align_line_endings(edit: EditOperation, content: str) -> EditOperation:
+    """Rewrite an edit's bare newlines to CRLF, for a file that uses CRLF.
+
+    Left alone when the find string already matches as written, or already carries a
+    carriage return -- a caller converting line endings means what it wrote.
+    """
+    if "\r" in edit.find or "\r" in edit.replace or edit.find in content:
```
vitalii-dynamiq: ### CRLF alignment skips `replace` whenever `find` already matches, splicing bare-LF lines into a CRLF file

**Medium severity**

The `edit.find in content` guard short-circuits the whole conversion, so `edit.replace` is left with bare newlines even though only `edit.find` was checked. A single-line anchor always matches as written in a CRLF file, so any edit that adds lines through such an anchor writes mixed line endings.

Reproduction: file `a\r\nb\r\n`, edit `{find: 'b', replace: 'b\nc'}`. `'b' in content` is true, the edit is returned unconverted, and the result written back is `a\r\nb\nc\r\n` — the `b` line now terminates with a bare LF while its neighbours keep CRLF. The equivalent multi-line anchor (`find: 'a\nb'`) does not match as written, so it takes the conversion path and correctly produces `a\r\nX\r\nb\r\n`; the outcome depends on whether the anchor happens to span a newline.

The rationale in the docstring — that a caller writing its own line endings means what it wrote — is carried by the `"\r" in edit.find or "\r" in edit.replace` checks. The `edit.find in content` term should gate only whether `find` is rewritten; `replace` still needs its bare newlines converted to match the file.

---

vitalii-dynamiq (commented):
Reviewed d6212005cee01f8fdcb610f0c0b9afb319054603 — 2 findings: 2 medium.

---

maksymbuleshnyi (dismissed):

dynamiq/nodes/tools/file_tools.py
```diff
@@ -37,6 +37,60 @@
...
+    pieces: list[str] = []
+    cursor = 0
+    for offset in positions:
+        pieces.append(content[cursor:offset])
+        pieces.append(replace)
+        cursor = offset + len(find)
+    pieces.append(content[cursor:])
+
+    def shifted(offset: int) -> int:
+        return offset + delta * sum(1 for position in positions if position < offset)
```
vitalii-dynamiq: ### Offset bookkeeping makes `replace_all` quadratic in the number of matches

**Medium severity**

`shifted` is evaluated once per tracked offset, and each evaluation walks the whole `positions` list, so a `replace_all` producing *k* hits costs O(k²). The line resolution at line 1494 adds a second quadratic path: `_line_of` rescans `content` from offset 0 for every one of the *k* offsets, so fixing `shifted` alone does not remove the cost. Before this change the same edit was a single `str.replace()`.

Measured with these helpers on a CSV of `user_N,old_status,value_N` rows, replacing `old_status` with `replace_all: true`:

| rows | file size | `_replace_at` | `_line_of` loop | old `str.replace` |
|---|---|---|---|---|
| 10,000 | 318 KB | 2.05 s | 0.57 s | 0.0009 s |
| 20,000 | 658 KB | 7.79 s | 2.32 s | 0.0019 s |
| 50,000 | 1.6 MB | 48.8 s | 14.6 s | 0.0050 s |

A 1.6 MB file takes just over a minute of CPU where it previously took 5 ms, which is long enough to stall or time out the calling agent. Nothing in the edit path bounds file size, so this is reachable with any ordinary bulk rename in a large data or log file.

Both paths can be linear: `positions` is already ascending, so the shift for each tracked offset can come from a single merge walk over the two sorted lists instead of a per-offset scan, and the line numbers can be resolved in one pass over `content` that counts newlines between consecutive offsets rather than restarting at 0 each time. Note also that only the first 10 resolved lines are ever displayed by `_format_lines`.

---

vitalii-dynamiq (commented):
Reviewed 0c23a94: 1 medium finding.

---

dynamiq/nodes/tools/file_tools.py
```diff
@@ -1363,17 +1432,43 @@ def _execute_edit(self, input_data: FileWriteInputSchema) -> dict[str, Any]:
...
-            occurrences = content.count(edit.find)
-            if occurrences == 0:
+            # Located against current content, not the original: a prior edit can
+            # add or remove candidate positions, and shift the lines they sit on.
+            positions = _find_positions(content, edit.find)
+            if not positions:
                 skipped.append(edit.find)
                 continue
-            count = occurrences if edit.replace_all else min(1, occurrences)
+            if len(positions) > 1 and not edit.replace_all:
```
vitalii-dynamiq, line 1446: ### Ambiguity judged on intermediate content aborts batches the stored file would accept, with a message the caller cannot act on

**Medium severity**

`positions` is computed against `content` as mutated by preceding edits, but the abort at line 1457 writes nothing, so the reported count and lines describe a state that never reaches disk. When an earlier edit's replacement introduces text matching a later edit's `find`, the batch is refused and the caller is told to disambiguate an anchor that is already unique in the only content it can read.

Running this loop against the current HEAD:

- `"alpha\nbeta\n"` with `[{find:"alpha",replace:"beta"}, {find:"beta",replace:"gamma"}]` aborts with `'beta' matches 2 places (lines 1, 2)`. The stored file is unchanged and contains one `beta`, at line 2.
- `"x\nfoo\ny\nfoo\n"` with `[{find:"x\n",replace:"x\nA\nB\n"}, {find:"foo",replace:"FOO"}]` aborts with `'foo' matches 2 places (lines 4, 6)`. The stored file has four lines; `foo` is at lines 2 and 4.

The caller's only documented recourse is the instruction at line 1464 — "include enough surrounding text to make each find string match exactly one place". Re-reading the file shows the anchor already matching one place, so the same request is retried and fails identically. Before this change the first batch applied cleanly, writing `"gamma\nbeta\n"`.

The mirror-image conflict is already handled the other way: when a prior edit *removes* a later `find`, line 1441 records it in `skipped` and the write still proceeds with a warning in the summary. Only the case where a prior edit *adds* candidate positions is fatal.

Deciding ambiguity against the original file content (which the `missing` pre-check at line 1424 already reads) would keep the reported counts and lines consistent with what the caller can observe, and would confine the abort to anchors that are genuinely ambiguous in the file being edited.

---

vitalii-dynamiq (commented):
Reviewed 587864a: 1 medium severity finding.

---

maksymbuleshnyi (approved):

PR #913, opened 2026-08-28
feat(composio): add Composio tool node

vitalii-dynamiq:
Adds `dynamiq.nodes.tools.Composio`, the runtime counterpart of the Composio Action node, alongside the existing Pipedream tool node.

Part of a three-repo change:
- **dynamiq** (this PR) — the executing node
- **mono** — https://github.com/dynamiq-ai/mono/pull/1330
- **ui** — https://github.com/dynamiq-ai/ui/pull/1538

Without this, a Composio Action node can be built and saved but not run.

## What changed

- `Composio` connection in `dynamiq/connections/connections.py` — API key auth against `https://backend.composio.dev/api/v3`.
- `Composio` node in `dynamiq/nodes/tools/composio.py`. It builds its pydantic input schema from the tool's `input_parameters` JSON Schema, reusing the helper the MCP node already uses for that same shape, and merges design-time argument values with runtime input.
- `resolve_root_object_schema` in `dynamiq/nodes/tools/mcp.py` — exposes the `$ref`/`allOf` root resolution the schema builder already performs, so callers that rewrite properties operate on the schema the model is actually built from.
- Examples under `examples/components/tools/composio/` — direct execution and an agent, in Python and declarative YAML.

## Four details worth review

**One node is one action, so a node is named after its action.** An agent keys its tool registry, its prompt listing and its result cache on `name`. An unnamed node takes its lowercased `tool_slug`, so two actions on the same agent stay distinct; an explicit `name` is kept.

**A failed action still returns HTTP 200.** Composio reports the real outcome in `successful` in the response body, so the status code alone would treat a failed action as a success. The node reads `successful` and raises `ToolExecutionException`.

It passes `recoverable=False` to record that an execute should not be replayed — it may have side effects — but that argument does not currently stop one, and the docstring says so. `Node._handle_failure` derives the flag from the exception type (`isinstance(e, RecoverableAgentException)`), not from the value at the raise site, so every failure is reported as recoverable and the agent may re-issue the call. The same is true of the `RECOVERABLE_CODES` split and of every other tool in the repo that passes the argument, `pipedream.py` included. Making it effective is a framework-level change to `_handle_failure` that would alter behaviour for every tool, so it belongs in its own PR rather than here.

**The tool version is sent explicitly.** Omitting it lets Composio resolve its own default, which can differ from the schema the arguments were configured against; the node sends the version captured when the action was picked.

**A configured argument outranks the schema default.** A parameter configured at design time is dropped from the schema's `required` list, so the runtime input may override it but need not supply one. Its JSON Schema `default` is dropped along with it — otherwise the built model emits that default whenever the caller stays silent, and it wins over the configured value when the payload is assembled, silently sending Composio's default instead of what was configured. Parameters that are *not* configured keep their schema default unchanged. Both rewrites run against the resolved root, so they apply equally to schemas expressed through a root-level `$ref` or `allOf`.

## Review findings fixed after the first approval

- `description` compounded a fresh copy of the parameter listing on every save/load cycle; `to_dict` now persists the description Composio supplied rather than the generated one.
- The declarative example defined no `flows:`/`workflows:`, so `Workflow.from_yaml_file` raised `IndexError` on it.
- Configured arguments were silently ignored when the schema root was a `$ref`/`allOf` composition — the action executed against Composio's default instead of the configured value, with no error raised.
- The docstring claimed `recoverable=False` prevented replay; it does not (see above).
- Nested objects, and objects inside arrays, posted an explicit `null` for every sub-field the caller left unset.
- Every node took the same default `name`, so two actions on one agent collapsed to a single entry in `BaseAgent.tool_by_names` and the first became unreachable.

## Verification

`pytest tests/unit` — 2715 passed, 1 skipped, including 45 tests for this node. `black` and `isort` clean at the repo's configured line length.

### DIFF
--- a/dynamiq/nodes/tools/composio.py
+++ b/dynamiq/nodes/tools/composio.py
@@ -200,8 +200,13 @@ class Composio(ConnectionNode):
         """Fail on a tool-level error.
 
         Composio answers 200 even when the action itself failed, so `successful` in the body -
-        not the HTTP status - carries the real outcome. The failure is reported as unrecoverable
-        because an execute must never be replayed: it may have side effects.
+        not the HTTP status - carries the real outcome.
+
+        `recoverable=False` records that an execute should not be replayed - it may have side
+        effects - but it does not currently stop one: `Node._handle_failure` derives the flag
+        from the exception type rather than from this value, and every `ToolExecutionException`
+        is a `RecoverableAgentException`. The same is true of `RECOVERABLE_CODES` above and of
+        every other tool in the repo that passes the argument.
         """
         if not response_json.get("successful", True):
             error_message = f"Composio tool {self.tool_slug} execution failed"
--- a/dynamiq/nodes/tools/composio.py
+++ b/dynamiq/nodes/tools/composio.py
@@ -6,7 +6,12 @@ from dynamiq.connections import Composio as ComposioConnection
 from dynamiq.nodes import NodeGroup
 from dynamiq.nodes.agents.exceptions import ToolExecutionException
 from dynamiq.nodes.node import ConnectionNode, ensure_config
-from dynamiq.nodes.tools.mcp import create_input_schema_from_json_schema, rename_keys_recursive
+from dynamiq.nodes.tools.mcp import (
+    create_input_schema_from_json_schema,
+    get_json_schema_definitions,
+    rename_keys_recursive,
+    resolve_root_object_schema,
+)
 from dynamiq.runnables import RunnableConfig
 from dynamiq.types.cancellation import check_cancellation
 from dynamiq.utils.logger import logger
@@ -140,6 +145,13 @@ class Composio(ConnectionNode):
             schema_dict = {}
         schema_dict = rename_keys_recursive(schema_dict, {"type_": "type"})
 
+        # A root composed purely of $ref/allOf is resolved to the object schema it stands for before
+        # anything below inspects it. The schema builder resolves the same way, so rewriting the
+        # unresolved root instead would edit keys it goes on to discard. The definitions live on the
+        # original root and have to be carried across, since the resolved schema no longer holds them.
+        definitions = get_json_schema_definitions(schema_dict)
+        schema_dict = resolve_root_object_schema(schema_dict, definitions)
+
         # Composio marks optionality the Pydantic way, so only `required` is authoritative. A parameter
         # already configured at design time is dropped from it: the runtime input may override the
         # configured value, but does not have to supply one.
@@ -157,7 +169,7 @@ class Composio(ConnectionNode):
                 if isinstance(prop, dict) and "default" in prop:
                     schema_dict["properties"][name] = {key: value for key, value in prop.items() if key != "default"}
 
-        return create_input_schema_from_json_schema(schema_dict, "ComposioToolSchema")
+        return create_input_schema_from_json_schema(schema_dict, "ComposioToolSchema", definitions=definitions)
 
     def _build_request(self, input_data: BaseModel) -> tuple[str, dict]:
         base_url = (self.connection.url or DEFAULT_COMPOSIO_URL).rstrip("/")
--- a/dynamiq/nodes/tools/composio.py
+++ b/dynamiq/nodes/tools/composio.py
@@ -1,6 +1,6 @@
 from typing import Any, Literal
 
-from pydantic import BaseModel
+from pydantic import BaseModel, PrivateAttr
 
 from dynamiq.connections import Composio as ComposioConnection
 from dynamiq.nodes import NodeGroup
@@ -49,6 +49,11 @@ class Composio(ConnectionNode):
     connected_account_id: str | None = None
     arguments: dict[str, Any] = {}
 
+    # The description Composio supplied for the action, before `_generate_description` appends the
+    # parameter listing onto it. `None` means it was never captured, which only happens if the node
+    # was built without running `__init__`.
+    _source_description: str | None = PrivateAttr(default=None)
+
     def __init__(self, input_props: dict[str, Any] | None = None, **kwargs):
         arguments = kwargs.get("arguments") or {}
         # A tool that declares no parameters is persisted with an empty or absent schema - the node
@@ -64,7 +69,17 @@ class Composio(ConnectionNode):
             input_props=input_props,
             **kwargs,
         )
+        self._source_description = self.description
         self.description = self._generate_description()
+
+    def to_dict(self, **kwargs) -> dict:
+        data = super().to_dict(**kwargs)
+        # At runtime `description` carries the parameter listing that `_generate_description` appended
+        # to the description Composio supplied. Persisting that listing would feed it back in on the
+        # next load, where it is appended again, so it would compound on every save/load cycle.
+        if self._source_description is not None and "description" in data:
+            data["description"] = self._source_description
+        return data
 
     @property
     def to_dict_exclude_params(self):
--- a/dynamiq/nodes/tools/composio.py
+++ b/dynamiq/nodes/tools/composio.py
@@ -40,7 +40,7 @@ class Composio(ConnectionNode):
     connection: ComposioConnection
     timeout: float = 180
     input_schema: type[BaseModel]
-    input_props: dict[str, Any]
+    input_props: dict[str, Any] = {}
     description: str = ""
     user_id: str
     toolkit_slug: str
@@ -49,8 +49,14 @@ class Composio(ConnectionNode):
     connected_account_id: str | None = None
     arguments: dict[str, Any] = {}
 
-    def __init__(self, input_props: dict[str, Any], **kwargs):
+    def __init__(self, input_props: dict[str, Any] | None = None, **kwargs):
         arguments = kwargs.get("arguments") or {}
+        # A tool that declares no parameters is persisted with an empty or absent schema - the node
+        # type serializes `input_props` with `omitempty`, and Composio itself reports "no declared
+        # schema" as either `{}` or `null` - so every such spelling has to yield an empty schema
+        # rather than a construction error.
+        if not isinstance(input_props, dict):
+            input_props = {}
         input_props = rename_keys_recursive(input_props, {"type": "type_"})
         input_schema = self.get_input_schema(input_props, arguments=arguments)
         super().__init__(
@@ -111,9 +117,9 @@ class Composio(ConnectionNode):
             schema_dict (dict[str, Any]): A JSON schema dictionary describing the tool's expected input.
             arguments (dict[str, Any]): Argument values already configured at design time.
         """
-        schema_dict = rename_keys_recursive(schema_dict, {"type_": "type"})
         if not isinstance(schema_dict, dict):
             schema_dict = {}
+        schema_dict = rename_keys_recursive(schema_dict, {"type_": "type"})
 
         # Composio marks optionality the Pydantic way, so only `required` is authoritative. A parameter
         # already configured at design time is dropped from it: the runtime input may override the
--- a/dynamiq/nodes/tools/composio.py
+++ b/dynamiq/nodes/tools/composio.py
@@ -1,0 +1,234 @@
+from typing import Any, Literal
+
+from pydantic import BaseModel
+
+from dynamiq.connections import Composio as ComposioConnection
+from dynamiq.nodes import NodeGroup
+from dynamiq.nodes.agents.exceptions import ToolExecutionException
+from dynamiq.nodes.node import ConnectionNode, ensure_config
+from dynamiq.nodes.tools.mcp import create_input_schema_from_json_schema, rename_keys_recursive
+from dynamiq.runnables import RunnableConfig
+from dynamiq.types.cancellation import check_cancellation
+from dynamiq.utils.logger import logger
+
+DEFAULT_COMPOSIO_URL = "https://backend.composio.dev/api/v3"
+SUCCESS_CODES = [200]
+RECOVERABLE_CODES = [400, 401, 402, 422, 429]
+
+
+class Composio(ConnectionNode):
+    """
+    A tool for executing a single Composio action (tool) on behalf of a Composio user.
+
+    Attributes:
+        name (str): Name of the tool
+        description (str): Description of the tool
+        group (Literal[NodeGroup.TOOLS]): The group the node belongs to
+        connection (ComposioConnection): The Composio API connection
+        timeout (float): Request timeout in seconds
+        input_props (dict[str, Any]): The tool's `input_parameters` JSON Schema
+        user_id (str): The Composio user id the action runs as
+        toolkit_slug (str): Slug of the Composio toolkit (app), e.g. `gmail`
+        tool_slug (str): Slug of the Composio tool (action), e.g. `GMAIL_SEND_EMAIL`
+        tool_version (str | None): Pins the tool version the arguments were configured against
+        connected_account_id (str | None): Pins the action to a specific connected account
+        arguments (dict[str, Any]): Argument values configured at design time
+    """
+
+    name: str = "composio"
+    group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
+    connection: ComposioConnection
+    timeout: float = 180
+    input_schema: type[BaseModel]
+    input_props: dict[str, Any]
+    description: str = ""
+    user_id: str
+    toolkit_slug: str
+    tool_slug: str
+    tool_version: str | None = None
+    connected_account_id: str | None = None
+    arguments: dict[str, Any] = {}
+
+    def __init__(self, input_props: dict[str, Any], **kwargs):
+        arguments = kwargs.get("arguments") or {}
+        input_props = rename_keys_recursive(input_props, {"type": "type_"})
+        input_schema = self.get_input_schema(input_props, arguments=arguments)
+        super().__init__(
+            input_schema=input_schema,
+            input_props=input_props,
+            **kwargs,
+        )
+        self.description = self._generate_description()
+
+    @property
+    def to_dict_exclude_params(self):
+        parent_dict = super().to_dict_exclude_params.copy()
+        parent_dict.update(
+            {
+                "input_schema": True,
+            }
+        )
+        return parent_dict
+
+    def _generate_description(self) -> str:
+        """
+        Generates a detailed description of the tool based on the input schema.
+
+        Returns:
+            str: A formatted description of the tool and its capabilities
+        """
+        schema_fields: dict[str, Any] = self.input_schema.model_fields
+        logger.debug(f"Tool {self.name} - Generating description from schema fields")
+
+        desc: list[str] = [self.description if self.description else f"{self.name} tool"]
+
+        required_fields: list[str] = [name for name, field in schema_fields.items() if field.is_required() is not False]
+        if required_fields:
+            desc.append("\nRequired Parameters:")
+            for field_name in sorted(required_fields):
+                field = schema_fields[field_name]
+                desc.append(f"- {field_name} ({str(field.annotation)}): {field.description}")
+
+        optional_fields: list[str] = [name for name, field in schema_fields.items() if field.is_required() is False]
+        if optional_fields:
+            desc.append("\nOptional Parameters:")
+            for field_name in sorted(optional_fields):
+                field = schema_fields[field_name]
+                desc.append(f"- {field_name} ({str(field.annotation)}): {field.description}")
+
+        if self.arguments:
+            desc.append("\nAlready configured parameters, that can be overridden:")
+            for field_name in sorted(self.arguments):
+                desc.append(f"- {field_name}: {self.arguments[field_name]}")
+
+        return "\n".join(desc)
+
+    def get_input_schema(self, schema_dict: dict[str, Any], arguments: dict[str, Any]) -> type[BaseModel]:
+        """
+        Creates an input schema based on the tool's `input_parameters` JSON Schema.
+
+        Args:
+            schema_dict (dict[str, Any]): A JSON schema dictionary describing the tool's expected input.
+            arguments (dict[str, Any]): Argument values already configured at design time.
+        """
+        schema_dict = rename_keys_recursive(schema_dict, {"type_": "type"})
+        if not isinstance(schema_dict, dict):
+            schema_dict = {}
+
+        # Composio marks optionality the Pydantic way, so only `required` is authoritative. A parameter
+        # already configured at design time is dropped from it: the runtime input may override the
+        # configured value, but does not have to supply one.
+        required = schema_dict.get("required") or []
+        schema_dict = {**schema_dict, "required": [name for name in required if name not in arguments]}
+
+        return create_input_schema_from_json_schema(schema_dict, "ComposioToolSchema")
+
+    def _build_request(self, input_data: BaseModel) -> tuple[str, dict]:
+        base_url = (self.connection.url or DEFAULT_COMPOSIO_URL).rstrip("/")
+        url = f"{base_url}/tools/execute/{self.tool_slug}"
+        payload = {
+            "arguments": {
+                **self.arguments,
+                # by_alias restores the original JSON Schema property names for parameters whose
+                # names are not valid Python identifiers.
+                **{k: v for k, v in input_data.model_dump(by_alias=True).items() if v is not None},
+            },
+            "user_id": self.user_id,
+            # Omitting the version lets Composio resolve its own default, which could differ
+            # from the schema the arguments were configured against.
+            **({"version": self.tool_version} if self.tool_version else {}),
+            **({"connected_account_id": self.connected_account_id} if self.connected_account_id else {}),
+        }
+        return url, payload
+
+    def _check_response_status(self, response: Any) -> None:
+        if response.status_code not in SUCCESS_CODES:
+            error_message = f"Composio API request failed with status code: {response.status_code}"
+            logger.error(f"Tool {self.name} - {error_message}")
+            recoverable = response.status_code in RECOVERABLE_CODES
+            raise ToolExecutionException(f"{error_message} and response: {response.text}", recoverable=recoverable)
+
+    def _check_execution_result(self, response_json: dict) -> None:
+        """Fail on a tool-level error.
+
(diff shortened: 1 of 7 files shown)


vitalii-dynamiq:
`Integration Tests With Creds` went red on `c5698f9`, on a single test:

```
FAILED tests/integration_with_creds/agents/test_agent_input.py::test_react_agent_role_with_special_characters[code_examples_with_braces]
===== 1 failed, 98 passed, 158 skipped, 1172 warnings in 191.69s =====
```

This is not this PR's failure, and there is no fix to port — nothing appears to be broken.

That test builds an `Agent` with `tools=[]` and asserts on a live model's free text (`"4" in content` and `"have a nice day!" in content.lower()`) under `RunnableConfig(request_timeout=30)`. Both assertions are nondeterministic against a live provider.

The diff cannot reach it. It is one new node (`dynamiq/nodes/tools/composio.py`), its unit tests, two examples, and two purely additive edits to shared files: a single `from .composio import Composio` line in `dynamiq/nodes/tools/__init__.py`, and a new `Composio` connection class appended to `dynamiq/connections/connections.py`. Nothing in prompt rendering, the agent loop, or LLM invocation is touched.

This suite runs only on `workflow_dispatch` or the `run-integration-tests-with-creds` label, so it has no baseline on `main` to compare against, and it needs provider credentials I cannot run locally. Re-running the job once to confirm.

---

vitalii-dynamiq:
Correcting my earlier comment: I called this a flake and re-ran it. It failed again identically, and the suite's own retries (`4 rerun`) did not save it, so "flake" was wrong — it is consistent.

It is still not this PR's failure, and that is now established rather than argued. I dispatched the same workflow against `main` at `ae0fce3` — this PR's merge base — and it fails the same way:

**`main` @ `ae0fce3`** ([run 33453224338](https://github.com/dynamiq-ai/dynamiq/actions/runs/33453224338))
```
FAILED tests/integration_with_creds/agents/test_agent_input.py::test_react_agent_role_with_special_characters[code_examples_with_braces]
===== 1 failed, 98 passed, 158 skipped, 1172 warnings in 191.73s =====
```

**This PR @ `c5698f9`** ([run 33449339992](https://github.com/dynamiq-ai/dynamiq/actions/runs/33449339992))
```
FAILED tests/integration_with_creds/agents/test_agent_input.py::test_react_agent_role_with_special_characters[code_examples_with_braces]
===== 1 failed, 98 passed, 158 skipped, 1206 warnings, 4 rerun in 336.31s =====
```

Same test, same pass/skip counts. The suite runs only on `workflow_dispatch` or the `run-integration-tests-with-creds` label, which is why a pre-existing breakage on `main` went unnoticed.

The failing assertion is `"have a nice day!" in content.lower()`, where the phrase comes from `extra_instructions` interpolated into an agent role full of Jinja `{% raw %}` blocks and braces. The model returns a valid JSON answer — `assert "4" in content` passes — but omits the trailing phrase. So it is a live-model expectation that no longer holds, in a test with `tools=[]` that this PR's diff (a new tool node, plus an additive import and connection class) cannot reach.

There is no fix to port: nothing is broken by a change, and repairing that expectation belongs on `main` rather than widening this PR. Flagging it for whoever owns that suite.

---

olbychos (dismissed):

dynamiq/nodes/tools/composio.py
```diff
@@ -0,0 +1,255 @@
...
+        if not isinstance(input_props, dict):
+            input_props = {}
+        input_props = rename_keys_recursive(input_props, {"type": "type_"})
+        input_schema = self.get_input_schema(input_props, arguments=arguments)
+        super().__init__(
+            input_schema=input_schema,
+            input_props=input_props,
+            **kwargs,
+        )
+        self.description = self._generate_description()
```
vitalii-dynamiq, line 84: ### `description` grows without bound across every save/load cycle

**Medium severity**

`__init__` assigns the generated description back onto the persisted `description` field, and `_generate_description` builds its output by appending the parameter listing onto whatever `self.description` already holds (line 93). `description` is not in `to_dict_exclude_params` — only `input_schema` is — so the compounded text is serialized and fed straight back in on the next load, where it is compounded again.

Each round trip appends another full copy of the `Required Parameters:` / `Optional Parameters:` / `Already configured parameters:` blocks. Running this PR's own `examples/components/tools/composio/agent_composio.yaml` through `WorkflowYAMLLoader` → `WorkflowYAMLDumper` repeatedly:

```
gen 0: len=434  ReqBlocks=1 OptBlocks=1
gen 1: len=801  ReqBlocks=2 OptBlocks=2
gen 2: len=1168 ReqBlocks=3 OptBlocks=3
gen 3: len=1535 ReqBlocks=4 OptBlocks=4
```

On the platform a node is saved and reloaded on every edit, so this compounds over the normal editing lifecycle. The description is what an agent reads when choosing between tools, so the result is steadily inflating prompt token cost and a tool description that repeats its own parameter list N times — degrading tool selection rather than helping it.

Keeping the generated text out of the persisted field fixes it: store the Composio-supplied description as-is and compute the augmented text on read (e.g. a property or a cached non-persisted attribute), or strip the generated suffix before regenerating.

---

vitalii-dynamiq (commented):
Reviewed 3933e2fd9e932317687b8683a2ba2c1a6a4f8fab — 1 medium finding.

---

dynamiq/nodes/tools/composio.py
vitalii-dynamiq: Confirmed and fixed in f84d5a3.

I reproduced it both ways. Direct construction, feeding `to_dict()["description"]` back in:

```
gen 0: len=240 ReqBlocks=1 OptBlocks=1
gen 1: len=421 ReqBlocks=2 OptBlocks=2
gen 2: len=602 ReqBlocks=3 OptBlocks=3
gen 3: len=783 ReqBlocks=4 OptBlocks=4
```

and through `WorkflowYAMLLoader` → `WorkflowYAMLDumper` on `agent_composio.yaml`, which matched your numbers exactly (434 → 801 → 1168 → 1535).

The fix takes the first option you suggested. `__init__` now captures the Composio-supplied description into a non-persisted `_source_description` before `_generate_description` runs, and `to_dict` writes that value back in place of the generated one. `description` still carries the augmented text at runtime — agents read `tool.description` directly (`nodes/agents/base.py:2498`, `components/schema_generator.py:433`) so it has to stay there — but the listing is now rebuilt from a stable base on every load instead of being appended to itself.

`_source_description` defaults to `None` rather than `""` deliberately: if a future path builds the node without running `__init__`, `to_dict` leaves `description` alone rather than blanking it.

Both reproductions are flat after the fix (240 and 434 across all four generations). Two regression tests added; negative control run with the fix reverted fails both and passes the other 29. `tests/unit/nodes/tools/` is 400 passed, 1 skipped.

One note beyond this PR: `dynamiq/nodes/tools/pipedream.py` has the identical pattern (`__init__` line 77, `_generate_description` line 89, `to_dict_exclude_params` excluding only `input_schema`), so it compounds the same way today. I've left it alone since it's pre-existing and outside this PR's scope — happy to send a separate PR for it if you want.

---

examples/components/tools/composio/agent_composio.yaml
```diff
@@ -0,0 +1,57 @@
...
+            subject:
+              anyOf:
+                - type: string
+                - type: "null"
+              default: null
+              description: Subject of the email.
+            is_html:
+              type: boolean
+              default: false
+              description: Whether the body is HTML.
```
vitalii-dynamiq, line 57: ### The declarative example defines no flow or workflow, so it cannot be loaded

**Low severity**

The file ends after `nodes:` — it has no `flows:` and no `workflows:` block. Both entry points that consume an example YAML fail on it:

- `Workflow.from_yaml_file(...)` reaches `list(file_data.workflows.values())[0]` in `dynamiq/workflow/workflow.py:99` with an empty `workflows` mapping and raises `IndexError`.
- `Workflow.from_yaml_file_data(..., wf_id=...)`, the form the sibling example uses, falls through to `raise ValueError(f"Workflow '{wf_id}' not found in YAML")` at line 104.

Every other tool example YAML in `examples/components/tools/` defines both blocks, including the `agent_pipedream.yaml` this example is modelled on, which closes with a `pipedream-agent-flow` referencing its node and a `pipedream-agent-workflow` referencing that flow. Adding the equivalent pair for `agent-composio` makes the example runnable as the PR description advertises.

---

vitalii-dynamiq (commented):
Reviewed f84d5a3 — 1 low-severity finding.

---

examples/components/tools/composio/agent_composio.yaml
vitalii-dynamiq: Confirmed and fixed in a31cdbe.

Reproduced before the fix — `Workflow.from_yaml_file` on the example raised exactly what you described:

```
ERROR - Failed to load workflow from YAML. list index out of range
IndexError list index out of range
```

Added the flow and workflow pair mirroring `agent_pipedream.yaml`:

```yaml
flows:
  composio-agent-flow:
    name: Composio Agent Flow
    nodes:
      - agent-composio

workflows:
  composio-agent-workflow:
    flow: composio-agent-flow
    version: 1
```

Both entry points now load it:

```
from_yaml_file OK: composio-agent-workflow nodes: ['Mailer Agent']
from_yaml_file_data OK: composio-agent-workflow
```

Worth noting for context, since it explains how this got through: nothing in the PR loads this file today — `use_composio.py` builds its nodes programmatically, unlike `agent_pipedream_jira.py` which loads `agent_pipedream.yaml` through `WorkflowYAMLLoader`. So the example was structurally incomplete rather than breaking a code path, but it is advertised as a declarative example and every sibling defines both blocks, so it should load.

The `WorkflowYAMLLoader`/`WorkflowYAMLDumper` round trip from the previous finding still holds on the completed file (434 chars flat across four generations), and `tests/unit/nodes/tools/test_composio.py` is 31 passed.

---

dynamiq/nodes/tools/composio.py
```diff
@@ -0,0 +1,270 @@
...
+            arguments (dict[str, Any]): Argument values already configured at design time.
+        """
+        if not isinstance(schema_dict, dict):
+            schema_dict = {}
+        schema_dict = rename_keys_recursive(schema_dict, {"type_": "type"})
+
+        # Composio marks optionality the Pydantic way, so only `required` is authoritative. A parameter
+        # already configured at design time is dropped from it: the runtime input may override the
+        # configured value, but does not have to supply one.
+        required = schema_dict.get("required") or []
```
vitalii-dynamiq, line 164: ### Configured-argument precedence is skipped when the root schema is a `$ref`/`allOf` composition

**Medium severity**

Both halves of the configured-argument handling read the root dict directly: `required` on line 146 and `properties` on line 149. `create_input_schema_from_json_schema` — the helper this node deliberately reuses — resolves a root defined purely through `$ref`/`allOf` down to its object schema first (`_resolve_to_object_schema`, `mcp.py:318`, called from `build` at line 338). On such a schema the root carries neither key, so:

- the `required` rewrite on line 147 filters an empty list and injects `required: []` at the root, which `_resolve_to_object_schema` then discards along with the rest of the root — the resolved schema keeps its own `required`, still naming the configured parameter;
- the `default`-stripping loop is guarded by `isinstance(properties, dict)` and no-ops entirely.

Running this file's `get_input_schema` and `_build_request` verbatim against one schema expressed three ways, with `arguments={"recipient": "configured@example.com"}`:

```
# recipient in `required`
inline root        recipient_required=False  -> sent 'configured@example.com'
root-level $ref    recipient_required=True   -> ValidationError: recipient Field required
root-level allOf   recipient_required=True   -> ValidationError: recipient Field required

# recipient optional, carrying default 'composio-default@example.com'
inline root        -> Composio receives 'configured@example.com'
root-level $ref    -> Composio receives 'composio-default@example.com'
root-level allOf   -> Composio receives 'composio-default@example.com'
```

The second case is the one that matters: no error surfaces, and the action executes against Composio's default instead of the configured value — the exact failure the PR description calls out as the reason the `default` is stripped, on an execute path that may have side effects. `tests/unit/nodes/tools/test_composio.py` covers only flat inline schemas, so neither case is caught.

Resolving the root once before inspecting it — the same way `build` does — makes both blocks operate on the schema whose `required` and `properties` actually reach the model.

---

vitalii-dynamiq (commented):
Reviewed a31cdbe — 1 medium finding.

---

dynamiq/nodes/tools/composio.py
vitalii-dynamiq: Confirmed and fixed in b442576. I reproduced your table verbatim before touching anything:

```
# recipient in `required`
inline root        recipient_required=False  -> sent 'configured@example.com'
root-level $ref    recipient_required=True   -> ValidationError: recipient
root-level allOf   recipient_required=True   -> ValidationError: recipient

# recipient optional, carrying default 'composio-default@example.com'
inline root        -> Composio receives 'configured@example.com'
root-level $ref    -> Composio receives 'composio-default@example.com'
root-level allOf   -> Composio receives 'composio-default@example.com'
```

and all six rows now behave identically (`recipient_required=False`, `configured@example.com` sent).

The fix resolves the root once, up front, the same way `build` does. Rather than re-implement the `$ref`/`allOf` walk here — which would duplicate `_merge_all_of` and drift from it — I exposed the existing resolution as `resolve_root_object_schema` in `mcp.py`, delegating to the same `_resolve_to_object_schema`, so there stays exactly one implementation.

One consequence worth calling out, since it is the trap in this fix: resolving the root drops the `$defs` that lived on it. Passing the resolved schema on its own would break any property that `$ref`s a nested definition. So the definitions are read from the original root and handed to `create_input_schema_from_json_schema` through its existing `definitions` parameter. `test_definitions_survive_a_resolved_root` covers that — and to be precise about what it is worth: it passes with and without the fix, so it is a guard against the fix regressing nested refs, not a reproduction of the reported bug.

The four tests that *do* reproduce the bug are the parametrized `$ref`/`allOf` cases. Negative control with only `composio.py` reverted: those four fail, the two `inline` variants still pass (so the parametrization is not failing trivially), and the other 34 pass.

Full unit suite is 2708 passed, 1 skipped; `tests/unit/nodes/tools/` is 407 passed, 1 skipped. black and isort clean at the project's 120-char config.

---

dynamiq/nodes/tools/composio.py
```diff
@@ -0,0 +1,282 @@
...
+        """Fail on a tool-level error.
+
+        Composio answers 200 even when the action itself failed, so `successful` in the body -
+        not the HTTP status - carries the real outcome. The failure is reported as unrecoverable
+        because an execute must never be replayed: it may have side effects.
+        """
+        if not response_json.get("successful", True):
+            error_message = f"Composio tool {self.tool_slug} execution failed"
+            logger.error(f"Tool {self.name} - {error_message}")
+            raise ToolExecutionException(f"{error_message} with error: {response_json.get('error')}", recoverable=False)
```
vitalii-dynamiq, line 222: ### `recoverable=False` does not prevent replay — a failed execute can still be reissued by the agent

**Medium severity**

`ToolExecutionException` subclasses `RecoverableAgentException` (`dynamiq/nodes/agents/exceptions.py:31`), and `Node._handle_failure` derives the flag from the type rather than from the value passed at the raise site:

```python
recoverable = isinstance(e, RecoverableAgentException)   # dynamiq/nodes/node.py:1287
```

That is unconditionally `True` here, so the `recoverable=False` argument is discarded before it reaches the result. Running the node against a stubbed response on each failure path:

```
200 + successful:false   status=failure recoverable=True
500 server error         status=failure recoverable=True
400 client error         status=failure recoverable=True
```

Downstream, `Agent._run_tool` (`dynamiq/nodes/agents/base.py:1762`) branches on that flag, and because it is `True` it raises `ToolExecutionException` rather than `ValueError`. The agent loop catches `RecoverableAgentException` at `dynamiq/nodes/agents/agent.py:1417`, appends the error as an observation, and continues — so the LLM can re-issue the identical call. This is exactly the replay the docstring says is prevented, on a path that has already reached Composio and may have produced a side effect (a sent email, a created issue).

The repo already documents the correct idiom in `dynamiq/nodes/tools/bedrock_agentcore_sandbox.py:227`: *"agents only retry RecoverableAgentException, so a raw error here would abort the run instead of allowing another attempt."* Raising a non-`RecoverableAgentException` type for the tool-level failure gets the intended behavior; the `RECOVERABLE_CODES` split in `_check_response_status` is inert for the same reason.

---

vitalii-dynamiq (commented):
Reviewed b442576 — 1 finding (1 medium).

---

dynamiq/nodes/tools/composio.py
vitalii-dynamiq: The diagnosis is correct and I've fixed what is mine in b43c496 — but I've deliberately not made the behaviour change, and I want to be explicit about that rather than quietly do half of it.

**Confirmed.** Reproduced with a stubbed client, and my first attempt at measuring it was wrong in a way worth flagging: a bare `MagicMock` has truthy `closed`/`is_closed`, so `ConnectionNode` decided the client was closed and replaced it with a real one — the run then failed on a genuine 401 from Composio, which is in `RECOVERABLE_CODES` and so produced `recoverable=True` for the wrong reason. With `closed = False` and `is_closed()` stubbed, the real numbers match yours exactly:

```
200 + successful:false   status=failure recoverable=True calls=1
500 server error         status=failure recoverable=True calls=1
400 client error         status=failure recoverable=True calls=1
```

The 500 is the sharpest case: `RECOVERABLE_CODES` explicitly excludes it and it still reports recoverable. `RecoverableAgentException.__init__` does store `self.recoverable`, so `_handle_failure` is discarding a value the exception carries faithfully.

**What I changed.** The defect that is genuinely this PR's is the docstring: I wrote that the failure "is reported as unrecoverable because an execute must never be replayed", and that is not what happens. It now states the actual behaviour and names why. I also fixed the inline comment on the existing test — `call_count == 1` pins that *the node* does not retry, not that the call is never replayed — and added `test_run_reports_every_failure_as_recoverable`, a characterization test over all three paths so the gap is visible in the suite rather than assumed away. Its comment says to update it when `_handle_failure` starts honouring the argument.

**What I did not change, and why.** Switching to a non-`RecoverableAgentException` type would abort the entire agent run on any failed Composio action — a single failed read kills the run. Two things argue against doing that here:

1. It would make Composio behave differently from Pipedream behind the same Action node. `pipedream.py` passes `recoverable=False` on exactly the same paths (lines 258, 298) and is equally inert; this PR's whole remit is to add Composio *the way Pipedream works*.
2. The `bedrock_agentcore_sandbox.py:227` comment is good evidence for the mechanism, but read in full it argues the other way — it explains why that code deliberately *uses* `ToolExecutionException`: "agents only retry RecoverableAgentException, so a raw error here would abort the run instead of allowing another attempt."

The real fix is framework-level: have `_handle_failure` honour `getattr(e, "recoverable", ...)` instead of the `isinstance` check, which would make every existing `recoverable=` argument in the repo finally mean something. That changes behaviour for every tool that passes it, so it wants its own PR and its own review rather than riding along here. Happy to open it — say the word.

`tests/unit` is 2711 passed, 1 skipped. black and isort clean.

---

dynamiq/nodes/tools/composio.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def _build_request(self, input_data: BaseModel) -> tuple[str, dict]:
+        base_url = (self.connection.url or DEFAULT_COMPOSIO_URL).rstrip("/")
+        url = f"{base_url}/tools/execute/{self.tool_slug}"
+        payload = {
+            "arguments": {
+                **self.arguments,
+                # by_alias restores the original JSON Schema property names for parameters whose
+                # names are not valid Python identifiers.
+                **{k: v for k, v in input_data.model_dump(by_alias=True).items() if v is not None},
```
vitalii-dynamiq: ### Nested object arguments send `null` for every field the caller left unset

**Medium severity**

The null filter is a flat comprehension over the top level of `model_dump()`, so it only drops parameters left unset at the root. Every object-typed property is built as a nested `BaseModel` whose non-required fields are annotated `X | None` with default `None` (`mcp.py:291-293`), and `model_dump` serializes those nested models in full — so unset sub-fields pass straight through the filter as explicit `null`s.

For an `Addr` declaring `city`, `zip` and `street`, a caller supplying `{"city": "Kyiv"}` posts:

```json
{"addr": {"city": "Kyiv", "zip": null, "street": null}}
```

to `/tools/execute/<slug>`. The action receives explicit nulls for two parameters nobody supplied, against a schema declaring both as `{"type": "string"}` — precisely the payload shape the top-level filter exists to prevent. Objects nested inside arrays behave the same way.

The existing coverage cannot catch this: `test_definitions_survive_a_resolved_root` asserts `payload["arguments"]["addr"] == {"city": "Kyiv"}` and holds only because its `Addr` declares exactly one property, which the test supplies. Adding a second optional property to that definition breaks the assertion.

`model_dump(by_alias=True, exclude_none=True)` applies the same rule at every depth.

---

vitalii-dynamiq (commented):
Reviewed b43c496 — 1 medium finding.

---

dynamiq/nodes/tools/composio.py
vitalii-dynamiq: Confirmed and fixed in 5856055, including your point about my own test being unable to catch it — that part was the most useful bit of the finding.

Reproduced before the fix, and the array case behaves exactly as you predicted:

```
nested : {'city': 'Kyiv', 'zip': None, 'street': None}
array  : [{'city': 'Lviv', 'zip': None, 'street': None}]
root   : ['addr', 'stops']          # `note`, unset at the root, was correctly dropped
```

`exclude_none=True` replaces the comprehension. It is equivalent at the root — the comprehension dropped root-level `None`s and so does this — and extends the same rule to every depth. Values the caller did supply are untouched, including falsy ones: `is_html=False` still reaches the payload, since `False is not None`.

```
nested : {'city': 'Kyiv'}
array  : [{'city': 'Lviv'}]
explicit: {'addr': {'city': 'Kyiv', 'zip': '01001'}, 'note': 'n'}
```

On the test critique — you were right, and it was a real gap in what I added one commit earlier, not just a theoretical one. The shared `$defs` fixture now gives `Addr` three properties (`city` required, `zip` and `street` optional), so `test_definitions_survive_a_resolved_root` can no longer pass merely because there was nothing left to leak. Negative control confirms it: with only `composio.py` reverted, that test now **fails** where it previously passed, alongside the two new ones (`test_unset_fields_are_dropped_at_every_depth`, `test_supplied_nested_values_are_kept`) — 3 failed, 40 passed.

`tests/unit` is 2713 passed, 1 skipped. black and isort clean.

---

olbychos (dismissed):

dynamiq/nodes/tools/composio.py
```diff
@@ -0,0 +1,289 @@
...
+        input_props (dict[str, Any]): The tool's `input_parameters` JSON Schema
+        user_id (str): The Composio user id the action runs as
+        toolkit_slug (str): Slug of the Composio toolkit (app), e.g. `gmail`
+        tool_slug (str): Slug of the Composio tool (action), e.g. `GMAIL_SEND_EMAIL`
+        tool_version (str | None): Pins the tool version the arguments were configured against
+        connected_account_id (str | None): Pins the action to a specific connected account
+        arguments (dict[str, Any]): Argument values configured at design time
+    """
+
+    name: str = "composio"
```
vitalii-dynamiq, line 43: ### Two Composio actions on one agent collide on the default `name`, making one unreachable

**Medium severity**

One node is one action, but every node defaults to the same `name`. An agent given two Composio actions — the natural shape here, e.g. `GMAIL_FETCH_EMAILS` plus `GMAIL_SEND_EMAIL` — carries two tools both named `composio`, and every dispatch path keys on that name:

- `BaseAgent.tool_by_names` (`base.py:2511`) is a dict comprehension over `_runtime_tools`, so the second tool overwrites the first. Every lookup (`agent.py:1018`, `1263`, `1558`, `2325`) then resolves to the same node, and the first action can never be invoked no matter which one the LLM names.
- `ToolCacheEntry` (`agents/utils.py:1233`) keys on `action=tool.name` plus the serialized input, so with caching on, a result cached for one action is served for the other whenever their arguments coincide.

`tool_description` also lists both entries under the identical name, so the LLM has no way to distinguish them in the first place.

Neither shipped example sets `name`, so both reproduce the collision as written. `MCPTool` — the repo's other one-node-per-tool type — declares `name: str` with no default (`mcp.py:389`) for this reason. Defaulting from `tool_slug`, or requiring `name`, keeps the two nodes distinct.

---

vitalii-dynamiq (commented):
Reviewed 5856055: 1 medium finding.

---

dynamiq/nodes/tools/composio.py
vitalii-dynamiq: Confirmed and fixed in 8118049. This is the one I should have caught myself — one node is one action is the whole design of this node, and I still gave every instance the same name.

Reproduced before the fix, with the two actions you named:

```
names   : composio | composio
by_names: ['composio']
collide : GMAIL_SEND_EMAIL - first action reachable: False
```

and after:

```
names   : gmail_fetch_emails | gmail_send_email
by_names: ['gmail_fetch_emails', 'gmail_send_email'] - both reachable: True
explicit: Mailer
```

I took the first of your two options — defaulting from `tool_slug` rather than requiring `name` — because the platform always supplies a name (the builder derives it from the action slug), so requiring it would break only bare Python construction and both shipped examples for no gain. An explicit `name` still wins, so nothing that sets one changes behaviour. The lowercased slug also survives `sanitize_tool_name` unchanged, since underscores are in its allowed set.

Two regression tests. Negative control with only `composio.py` reverted: `test_two_actions_do_not_collide_on_the_default_name` fails, 44 pass. To be precise about the other one — `test_an_explicit_name_is_kept` passes with and without the fix; it guards the fix against clobbering a caller-supplied name rather than reproducing the bug.

`tests/unit` is 2715 passed, 1 skipped. black and isort clean.

Worth noting for context, not as a request: `pipedream.py` carries the same `name: str = "pipedream"` default and would collide the same way given two actions on one agent. It is pre-existing and outside this PR, so I have left it — happy to send it with the other pipedream fix if you want them.

---

olbychos (approved):

TrachukT (approved):

olbychos (approved):

PR #915, opened 2026-08-31
feat: add google document ai

olbychos:


dynamiq/components/converters/google_document_ai.py
```diff
@@ -0,0 +1,374 @@
...
+        Returns:
+            Document: The processed Google Document AI document.
+        """
+        from google.cloud import documentai
+
+        request = documentai.ProcessRequest(
+            name=self.processor_name,
+            raw_document=documentai.RawDocument(content=content, mime_type=mime_type),
+            process_options=documentai.ProcessOptions(
+                ocr_config=documentai.OcrConfig(enable_native_pdf_parsing=self.enable_native_pdf_parsing),
```
vitalii-dynamiq: ### `ocr_config` is sent unconditionally, so every non-OCR processor fails

**High severity**

`ProcessOptions.ocr_config` is only accepted by two processor types. From the pinned `google-cloud-documentai==3.15.0` proto docs for this field:

> Only applicable to `OCR_PROCESSOR` and `FORM_PARSER_PROCESSOR`. Returns error if set on other processor types.

Here `ocr_config` is constructed on every request with no way to omit it. Setting a submessage in proto-plus marks the field present on the wire even when all its scalars are default, so `enable_native_pdf_parsing=False` does not suppress it — the request still carries `ocr_config {}`.

`processor_id` is a free-form field and the node is documented generically as "extract plain text from documents". A Layout Parser, Custom Extractor, or Invoice Parser processor all return the `document.text` plus `pages` that `_page_texts` slices, so pointing this converter at one is a realistic configuration — and it gets `INVALID_ARGUMENT` on every single call, with no field on the node or component that can turn the option off.

Build `ProcessOptions` without `ocr_config` unless the caller opts into it.

---

dynamiq/components/converters/google_document_ai.py
```diff
@@ -0,0 +1,374 @@
...
+
+        Returns:
+            str: The guessed type, or PDF when the name carries no usable extension.
+
+        Raises:
+            ValueError: If the file type is one Document AI cannot process.
+        """
+        guessed_mime_type = mimetypes.guess_type(file_path)[0]
+        if guessed_mime_type is None:
+            return PDF_MIME_TYPE
```
vitalii-dynamiq, line 230: ### WebP files are silently declared as PDF on Python 3.10

**Medium severity**

`image/webp` is advertised in `SUPPORTED_MIME_TYPES`, but CPython 3.10 has no `.webp` entry in `mimetypes.types_map` — it was added in 3.11. `pyproject.toml` declares `requires-python = ">=3.10,<3.14"`.

On 3.10, in any environment without a system `/etc/mime.types` (python:3.10-slim, Alpine, macOS), `mimetypes.guess_type("scan.webp")` returns `None`, so this line declares the WebP bytes as `application/pdf` and Document AI is asked to parse a WebP as a PDF. The same path is reached for unnamed `BytesIO` input: `get_filename_for_bytesio` sniffs the magic bytes with `filetype` and produces `<uuid>.webp`, which then also degrades to PDF.

The new `test_every_supported_type_is_declared_verbatim[image/webp]` case fails there too — `mimetypes.guess_extension("image/webp")` returns `None`, so the fixture builds `scanNone`. CI only runs 3.13, so this stays invisible.

Map the supported types explicitly rather than relying on the interpreter's `mimetypes` table.

---

vitalii-dynamiq (commented):
Reviewed c1f07b2: 2 findings (1 high, 1 medium).

---

dynamiq/components/converters/google_document_ai.py
```diff
@@ -0,0 +1,389 @@
...
+        Document AI is left to report the real error.
+
+        Args:
+            content (bytes): The raw file content.
+            mime_type (str): The MIME type to declare to Document AI.
+
+        Yields:
+            bytes: The raw content of each request, in page order.
+        """
+        if mime_type != PDF_MIME_TYPE:
```
vitalii-dynamiq, line 276: ### Multi-page TIFFs bypass the page-limit splitting and are rejected by the API

**Medium severity**

`_request_batches` short-circuits on anything that is not `application/pdf`, on the stated assumption that images are "single-page by nature". TIFF is not: `.tif` and `.tiff` are both in `MIME_TYPES_BY_EXTENSION`, so `_resolve_mime_type` happily returns `image/tiff` for them, and Document AI counts each TIFF sub-image as a page against the same online limit this method exists to work around.

A 40-page scanned TIFF — the ordinary output of fax and bulk-scanner archives — therefore reaches `_process_bytes` unsplit and `process_document` fails the whole `run()` call with a page-limit `InvalidArgument`, while the identical 40-page document as a PDF is split and processed successfully. The page limit is the exact condition the splitting path was written for, and for one of the accepted input types it never engages.

Either derive the page count for TIFF and split it the way PDFs are split, or stop advertising `.tif`/`.tiff` in the accepted extensions.

---

dynamiq/components/converters/google_document_ai.py
```diff
@@ -0,0 +1,389 @@
...
+            filepath (str): The path of the processed file.
+            elements (list[str]): The text of each page, in page order.
+            document_creation_mode (DocumentCreationMode): How to group pages into Documents.
+            metadata (dict[str, Any]): Metadata to attach to the documents.
+
+        Returns:
+            list[Document]: The created Documents.
+        """
+        document_metadata = copy.deepcopy(metadata)
+        document_metadata["file_path"] = filepath
```
vitalii-dynamiq: ### A caller-supplied `file_path` is discarded and replaced with a per-run random UUID

**Medium severity**

This assignment overwrites `file_path` unconditionally. Every other file converter in this package — `docx.py:142`, `excel.py:240`, `pptx.py:92`, `text.py:88`, `textract.py:689` — routes the same metadata through `build_source_metadata`, which deliberately keeps the caller's value: `effective_file_path = result.get("file_path") or file_path`, a behaviour pinned by `tests/unit/components/test_converter_metadata.py`.

The divergence is observable on the `BytesIO` path. For a nameless stream, `get_filename_for_bytesio` mints `f"{generate_uuid()}.{ext}"`, so a caller that does `run(files=[BytesIO(pdf_bytes)], metadata={"file_path": "contracts/2026/invoice.pdf"})` gets documents stamped with a fresh random UUID instead — a different one on every run of the same input.

That value is load-bearing downstream: `DEFAULT_SEARCHABLE_TEXT_METADATA_FIELDS` in `dynamiq/storages/vector/utils.py` embeds `file_path` into the indexed searchable text, and `_DocumentFeatures.from_document` in `dynamiq/components/splitters/auto.py` reads it first when choosing a splitting rule. So re-ingesting the same document through this converter appends a duplicate copy under a new key rather than replacing the old one, filters keyed on the caller's path match nothing, and the UUID is indexed as search text.

Use `build_source_metadata(metadata, filepath)` here, as the sibling converters do.

---

vitalii-dynamiq (commented):
Reviewed 469ad67e3e4906265f63e31dffe29f729d411bde — 2 findings (2 medium).

---

dynamiq/components/converters/google_document_ai.py
```diff
@@ -0,0 +1,422 @@
...
+                for start in range(0, page_count, pages_per_batch):
+                    frames = []
+                    for page_number in range(start, min(start + pages_per_batch, page_count)):
+                        image.seek(page_number)
+                        frames.append(image.copy())
+
+                    batch = BytesIO()
+                    frames[0].save(batch, format="TIFF", save_all=True, append_images=frames[1:])
+                    yield batch.getvalue()
+        except Exception as e:
```
vitalii-dynamiq, line 314: ### TIFF batching `try` wraps its own `yield`s, so a mid-loop failure re-sends the whole file

**Medium severity**

The `try` here spans the entire splitting loop, including the `yield` on line 330. `_extract_page_texts` consumes this generator lazily and `extend`s per batch (line 252-253), so batches are already processed and their text collected before a later iteration runs.

`image.n_frames` walks the IFD chain and counts frames without decoding pixel data, so a frame that cannot be decoded or re-encoded surfaces only later, inside the loop: `image.copy()` triggers `load()` on the sought frame and raises on a truncated stream or an unsupported per-frame compression, and `frames[0].save(...)` raises when a frame's mode cannot be written as TIFF. That is ordinary input for the scanner and fax archives this splitting path exists to handle.

When it happens after k batches have been yielded, control falls into this handler and yields the complete unsplit `content`. The first `k * max_pages_per_request` pages are then appended a second time from the full document, and the full document is exactly what exceeds the online page limit this method works around — so the run either returns duplicated page text or fails with a page-limit `InvalidArgument`, in both cases after logging "Could not read the TIFF page count locally", which is not what went wrong.

The PDF path at lines 281-287 gets this right: only the page-count read sits inside `try`, and the splitting loop runs outside it. Narrow this `try` to the `Image.open`/`n_frames` read the same way.


---

vitalii-dynamiq (commented):
Reviewed 78ff323: 1 finding (1 medium).

---

dynamiq/components/converters/google_document_ai.py
```diff
@@ -0,0 +1,425 @@
...
+        description="Create one document per file, with page texts joined, or one per page.",
+    )
+    enable_native_pdf_parsing: bool | None = Field(
+        default=None,
+        description=(
+            "Read embedded text from born-digital PDFs instead of OCR-ing rendered pages. "
+            "When unset, OCR-specific process options are omitted."
+        ),
+    )
+    imageless_mode: bool = Field(
```
delamainer: What about `enabled_imageless_mode`? To keep it consistent with the rest of the platform.

TrachukT (approved):

PR #919, opened 2026-09-02
feat(memory): add titled agent notes

mykhailobuleshnyi:


dynamiq/nodes/tools/notes.py
```diff
@@ -0,0 +1,292 @@
...
+
+        try:
+            note, outcome = self.backend.write(
+                user_id=self.user_id,
+                title=input_data.title,
+                description=input_data.description,
+                content=input_data.content,
+            )
+            verb = "Created" if outcome is NoteWriteOutcome.CREATED else "Overwrote"
+            return {"content": f"{verb} note {note.title!r}.\n\nYour notes:\n{self._index_block()}"}
```
vitalii-dynamiq, line 211: ### A failing index refresh turns a committed note mutation into a fatal, run-ending error

**Medium severity**

`self._index_block()` is evaluated inside the same `try` as `backend.write(...)`, and it reaches `NotesBackend.index()`, which wraps any storage failure in `NotesStorageError` (`memory/notes/base.py:209-211`).

If `write` commits and the subsequent `index()` statement then fails — the connection is dropped between the two statements, or the second query hits a statement timeout — control falls into `except NotesError` and `_fail` classifies `NotesStorageError` as non-recoverable, raising it instead of a `ToolExecutionException`. `Agent._execute_single_tool` only catches `RecoverableAgentException` (`nodes/agents/agent.py:1419`), so the exception escapes the ReAct loop and aborts the entire agent run, even though the note was already persisted. The model never learns the write succeeded, and on a later run it sees the note it was told it failed to create.

`DeleteNoteTool.execute` (line 281) has the identical shape, and `ReadNoteTool.execute` (line 234) additionally discards the note bodies it already fetched successfully.

The index refresh is cosmetic in all three paths — its own comment at line 233 calls it a self-correction aid — so it belongs in its own `try` whose failure degrades the message rather than losing the completed operation.

---

vitalii-dynamiq (commented):
Reviewed 794a0f24d1f395656e85f314e5fabc442b4952ec — 1 finding (1 medium).

---

dynamiq/nodes/tools/notes.py
```diff
@@ -0,0 +1,292 @@
...
+        data["backend"] = self.backend.to_dict(include_secure_params=include_secure_params, **kwargs)
+        return data
+
+    def _start_run(self, config: RunnableConfig | None, **kwargs) -> None:
+        config = ensure_config(config)
+        check_cancellation(config)
+        self.run_on_node_execute_run(config.callbacks, **kwargs)
+
+    def _index_entries(self):
+        return self.backend.index(user_id=self.user_id)
```
vitalii-dynamiq, line 162: ### The tool-echoed index ignores both `NotesConfig` caps and diverges from the injected index

**Medium severity**

`_index_entries()` calls `backend.index()` with no `limit`, so it falls through to `effective = self.max_index_entries` — the *backend's* cap (`memory/notes/base.py:198`, default 200) — and no character cap is applied to the rendered block at all.

The prompt-injected index is built differently: `_render_notes_index` passes `limit=self.notes.max_index_entries` and then hard-caps the rendered string at `self.notes.max_index_chars` (`nodes/agents/base.py:1119-1130`).

The two therefore disagree for any user whose note count grows, even on stock configuration. With defaults, a user with 200 notes gets a system-prompt index trimmed to 6000 chars, while every `write_note` / `delete_note` observation echoes all 200 entries at up to ~505 chars per line. Lowering `NotesConfig.max_index_entries` to 20 while leaving the backend at its default widens the gap further: 20 lines injected, 200 echoed. Neither `max_index_entries` nor `max_index_chars` has any effect on the echoed block.

That is the exact failure `memory/notes/formatting.py` exists to prevent — its module docstring notes that if the injected index and a tool's echoed index differed, the model would read them as two different lists. Here the echoed list is a strict superset carrying titles the injected snapshot omitted, and it lands in the observation stream on every mutation.

Passing the configured entry limit and applying the same character cap in `_index_block()` keeps the two renders identical.

---

vitalii-dynamiq (commented):
Reviewed e79b873: 1 medium severity finding.

---

PR #923, opened 2026-09-07
feat: cli commands

maksymbuleshnyi:


dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,523 @@
...
+    "dynamiq.memory.backends.Dynamiq",
+    "dynamiq.memory.backends.PostgreSQL",
+    "dynamiq.memory.backends.Pinecone",
+    "dynamiq.memory.backends.Qdrant",
+    "dynamiq.memory.backends.Weaviate",
+    "dynamiq.memory.backends.DynamoDB",
+}
+
+# Declared by dynamiq/nodes/tools/pipedream.py. `props` is a component schema, not a field.
+PIPEDREAM_FIELDS = {
```
vitalii-dynamiq, line 464: ### `PIPEDREAM_FIELDS` omits `connection` and `timeout`, so valid Pipedream tools fail validation

**High severity**

`connection` is a *required* field on `dynamiq.nodes.tools.Pipedream` (`dynamiq/nodes/tools/pipedream.py:57`: `connection: PipedreamConnection`), and `timeout: float = 180` is declared right below it. Neither is in this set, so `check_pipedream` reports them as unknown — and that message is appended to `problems`, which callers at lines 336 and 357 splice straight into `errors`, not `warnings`.

The repository's own canonical example carries exactly this shape (`examples/components/tools/pipedream/pipedream_with_files.yaml:54`, `connection: 667abff4-...` on a `dynamiq.nodes.tools.Pipedream` tool). Running `flowcheck.validate()` on that tool inside an agent produces:

```
tool gmail on node 'agent': unknown field(s) connection - they are ignored.
```

`workflow validate` then raises `ClickException` and exits 1, and `workflow save` (line 768) reports `1 problem(s) in the saved flow` after a save that actually succeeded. The message itself states the fields "are ignored", so even for genuinely inert keys this belongs in `warnings` rather than in the blocking error list.


---

dynamiq/cli/commands/workflow.py
```diff
@@ -423,29 +519,125 @@ def test_workflow(
         form["dry_run"] = "true"
     if last_node_output:
         form["last_node_output"] = "true"
-    echo_response(api.post("/v1/workflows/test", data=form))
+    # This endpoint takes multipart; a urlencoded body is answered with 415.
+    response = api.post("/v1/workflows/test", files={k: (None, v) for k, v in form.items()})
```
vitalii-dynamiq: ### `workflow test` re-executes the flow up to 5 times on a slow run

**High severity**

`--dry-run` now defaults to on, and per the docstring above that is the form the endpoint actually executes: "the flow executes and its tools really act, so a Notion tool creates a real page". Before this change the default omitted `dry_run` and the call answered 400, so no execution happened.

That execution now runs through `ApiClient._request`, which applies `timeout: float = 30.0` (`dynamiq/cli/client.py:92`) under `@retry(stop=stop_after_attempt(5), retry=retry_if_exception_type(requests.RequestException), reraise=True)`. `requests.exceptions.ReadTimeout` is a `RequestException`, so any flow taking longer than 30 seconds — routine for an agent doing LLM plus tool calls — is silently re-POSTed four more times. Each attempt starts a fresh server-side execution whose tools really act, so the user gets up to five duplicated real side effects, and then a raw `ReadTimeout` traceback rather than a message.

Pass a timeout suited to execution for this call, and keep the non-idempotent `/v1/workflows/test` POST out of the timeout retry path.


---

dynamiq/cli/commands/workflow.py
```diff
@@ -460,3 +652,118 @@ def release_workflow(
...
+    for path in custom_paths:
+        payload = read_json_arg(path)
+        entries = payload.get("custom_node_data", payload) if isinstance(payload, dict) else None
+        if not isinstance(entries, dict):
+            raise click.ClickException(f"{path}: expected an object of node id -> entry.")
+        custom.update(entries)
+
+    flow_ui = flow_ui_for(flow_body)
+    for node_id, entry in custom.items():
+        flow_ui["custom_node_data"][node_id] = {**flow_ui["custom_node_data"].get(node_id, {}), **entry}
```
vitalii-dynamiq: ### `--custom` entries can never reach a top-level node

**Medium severity**

`--custom` is documented as taking "a bare {node_id: entry} mapping", but for a top-level node `flow_ui_for` keys `custom_node_data` by a freshly generated canvas uuid (line 378: `custom_node_data[ui_id] = {**node, "id": ui_id, "name": slug}`), not by the flow node id. Only nodes nested inside an agent are keyed by their flow id (`nested_custom_entries`, line 311).

So for a Pipedream node placed directly in the DAG — a shape `flowcheck` explicitly supports (`flowcheck.py:356`) — a `--custom` entry keyed by the flow node id matches nothing here and merges into a new orphan key instead. The real uuid-keyed entry keeps no `pipedreamComponent`, so the warning at line 721 still fires and the tool still saves and draws with no logo, account picker or configuration form. The user cannot work around it either: the uuid is regenerated on every invocation, so there is no key they can supply.

Resolve each incoming key through `ui_ids` (falling back to the key itself for nested nodes) before merging.


---

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,523 @@
...
+            if isinstance(target, dict):
+                target = target.get("id")
+            if target not in known:
+                errors.append(f"node {label!r} depends on {target!r}, which is not a node in this flow.")
+        if node_type == INPUT_TYPE and node.get("depends"):
+            errors.append(f"Input node {label!r} must not depend on anything.")
+        if node_type != INPUT_TYPE and not node.get("depends"):
+            errors.append(f"node {label!r} has no `depends`, so it never runs. Wire it to an upstream node.")
+
+        selector = (node.get("input_transformer") or {}).get("selector") or {}
```
vitalii-dynamiq: ### A non-dict `input_transformer` crashes the validator instead of being reported

**Medium severity**

`or {}` only guards falsy values, so a truthy non-dict `input_transformer` reaches `.get` and raises. Confirmed by running `flowcheck.validate()`:

```python
flow = {"nodes": [
    {"id": "input", "type": "dynamiq.nodes.utils.Input"},
    {"id": "a", "type": "dynamiq.nodes.agents.Agent",
     "depends": [{"node": "input"}], "input_transformer": "$.input.output"},
]}
flowcheck.validate(flow)
# AttributeError: 'str' object has no attribute 'get'
```

A list value fails the same way. `workflow validate` exists precisely to diagnose malformed hand- or model-written flows, and writing the JSONPath directly into `input_transformer` is the same class of mistake the checker already handles at line 219 (top-level `selector`). Here the user gets an uncaught `AttributeError` traceback instead of the diagnosis, and every other problem in the flow goes unreported. Guard with `isinstance(..., dict)` and emit an error for the malformed shape, as the `selector` check below already does.


---

vitalii-dynamiq (commented):
Reviewed 8c2c8b7ada9e0517413a1b7dc2394bbbb038303c: 2 high, 2 medium.

---

dynamiq/cli/commands/knowledgebase.py
```diff
@@ -0,0 +1,173 @@
...
+@with_api_and_settings
+def upload_files(*, api: ApiClient, settings: Settings, knowledgebase_id: str, paths: tuple[str, ...]):
+    """Upload one or more files into a knowledge base.
+
+    The endpoint takes a multipart form with a repeated `files` field, so several files go up
+    in one call. Ingestion is ASYNCHRONOUS: a 2xx here means the files were accepted, not that
+    they are searchable. Poll `knowledge-base items <id>` until each item's status settles
+    before telling anyone the content is available.
+    """
+    handles = [open(path, "rb") for path in paths]  # noqa: SIM115 - closed below
```
vitalii-dynamiq, line 72: ### Retried uploads send empty files and report success

**High severity**

These handles are opened once and handed to `api.post(..., files=files)`, but the retry lives on `ApiClient._request` (`client.py:73`): `stop_after_attempt(5)` with `retry_if_exception_type(requests.RequestException) | retry_if_result(r.status_code in {502, 503, 504})`.

Every retry re-enters `self._client.request` with the *same* file objects. `requests` has already read them to EOF on the first attempt and nothing seeks back to 0, so attempts 2-5 build a multipart body whose `files` parts are zero bytes. A gateway 503 or a dropped connection mid-upload — the exact conditions the retry exists for — therefore uploads empty files.

`echo_response` then sees a 2xx and prints the body, so the user is told the files were accepted. Since ingestion is asynchronous, the empty items surface later as successfully-ingested-but-empty knowledge base content, with nothing pointing back at the retry.

`skill.py:63` and `skill.py:123` pass an open handle through the same retry path and fail identically.

Re-open or `handle.seek(0)` per attempt, or read the bytes once and pass the buffer.

---

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,523 @@
...
+            "for an agent, add a dynamiq.nodes.agents.Agent node and put the tools inside it."
+        )
+
+    for node in nodes:
+        if not isinstance(node, dict):
+            continue
+        label = node.get("id", "?")
+        node_type = str(node.get("type") or "")
+
+        for dependency in node.get("depends") or []:
```
vitalii-dynamiq: ### `depends: "input"` is iterated character by character

**Medium severity**

`node.get("depends") or []` yields the string itself when `depends` is a bare string, so this loop walks its characters. `validate_flow_command` (`workflow.py:669`) validates the raw JSON — it never calls `normalize_flow` — while `save` (`workflow.py:454`) does, and `normalize_flow` explicitly accepts this shorthand: `* depends: "input" or ["input"] -> [{"node": "input"}]` (`workflow.py:168`).

So a flow that `save` accepts is rejected by `validate` with one bogus error per character:

```python
flowcheck.validate({"nodes": [
    {"id": "input", "type": "dynamiq.nodes.utils.Input"},
    {"id": "a", "type": "dynamiq.nodes.agents.Agent", "depends": "input"},
]})
```

```
node 'a' depends on 'i', which is not a node in this flow.
node 'a' depends on 'n', which is not a node in this flow.
node 'a' depends on 'p', which is not a node in this flow.
node 'a' depends on 'u', which is not a node in this flow.
node 'a' depends on 't', which is not a node in this flow.
```

`validate` exits 1 with five invented dependency names, and the real problems in the flow are buried under them. Normalize the flow before validating, or coerce a string `depends` to a single-element list here.

---

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,523 @@
...
+        if not text.startswith("dynamiq.nodes.") or len(parts) < 4:
+            hint = " Did you mean a `dynamiq.nodes.<group>.<Class>` path?" if "." not in text else ""
+            errors.append(
+                f"type {node_type!r} is not a node type. Types are fully qualified dotted paths "
+                f"like `dynamiq.nodes.agents.Agent`, `dynamiq.nodes.llms.OpenAI` or "
+                f"`dynamiq.nodes.tools.Pipedream` - never a short label.{hint} "
+                "Copy the exact string from `dynamiq workflow get <id>` on a workflow that works."
+            )
+            continue
+        if groups and parts[2] not in groups:
```
vitalii-dynamiq: ### The namespace check rejects real SDK modules

**Medium severity**

`node_groups()` reads `NodeGroup`, but that enum is the node *group* taxonomy, not the module layout. `dynamiq/nodes/` contains `knowledgebases/` and `knowledge_graphs/`, and neither has a `NodeGroup` member — those classes declare `group: Literal[NodeGroup.TOOLS]` (`knowledgebases/knowledge_graph.py:47`), `EXTRACTORS` or `RETRIEVERS` while living in their own package.

`Node.type` is built from the module path, not the group (`nodes/node.py:502`: `f"{self.__module__.rsplit('.', 1)[0]}.{self.__class__.__name__}"`), so the SDK's own emitted type for these nodes is `dynamiq.nodes.knowledgebases.DynamiqKnowledgebaseGraphSearch` — exactly the string this rejects:

```
type 'dynamiq.nodes.knowledgebases.DynamiqKnowledgebaseGraphSearch' is not a real namespace -
the SDK has no `dynamiq.nodes.knowledgebases` module.
```

The message is false, and the docstring above states the check only catches "a namespace the SDK has no module for at all". A flow using a knowledge base node fails `workflow validate` with exit 1, and `workflow verify` (`workflow.py:750`) raises `ClickException` on a workflow that saved and runs correctly.

Derive the set from the `dynamiq.nodes` package's submodules rather than from `NodeGroup`.

---

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,523 @@
...
+                errors.extend(check_pipedream(tool, where))
+            elif tool.get("connection") is not None and not UUID_RE.match(str(tool.get("connection"))):
+                errors.append(
+                    f"{where}: connection {tool.get('connection')!r} is not a UUID. "
+                    "Run `dynamiq connection list`."
+                )
+
+        # The API reports these as `cannot be blank` with no node name.
+        if node_type.startswith("dynamiq.nodes.llms."):
+            if not UUID_RE.match(str(node.get("connection") or "")):
```
vitalii-dynamiq: ### A `connection` requirement placeholder cannot be validated or saved

**Medium severity**

`requirement_problems` documents that `{"$type": "requirement", "$id": ...}` "is a legal value ANYWHERE" (line 415), and the new `requirement-add` command documents the `connection` requirement type plus how to reference it (`workflow.py:559-564`):

```
type "connection"        -> spec {"type": "dynamiq.connections.<X>"}
Reference the returned id from the flow as
{"$type": "requirement", "$id": "<id>", "value_path": "$.account_id"}.
```

But none of the `connection` checks call `requirement_problems` — they go straight to `UUID_RE.match(str(...))`. A dict stringifies to something that never matches, so the placeholder is reported as a bad UUID:

```python
flowcheck.validate({"nodes": [
    {"id": "input", "type": "dynamiq.nodes.utils.Input"},
    {"id": "llm", "type": "dynamiq.nodes.llms.OpenAI", "model": "gpt-4o",
     "depends": [{"node": "input"}],
     "connection": {"$type": "requirement", "$id": "req-1", "value_path": "$.account_id"}},
]})
# node 'llm': connection {'$type': 'requirement', ...} is not a connection UUID.
```

Lines 250 and 339 have the same gap. `save` fails even earlier: `check_connection` (`workflow.py:149`) raises `ClickException` outright on any non-UUID `connection`, and `normalize_flow` runs it over node, tool and llm connections (lines 193-199).

So the `connection` requirement flow this PR documents can be declared through the API but never validated or saved through the CLI. Route these values through `requirement_problems` first, as `external_user_id` already does at line 458.

---

dynamiq/cli/commands/workflow.py
```diff
@@ -460,3 +652,118 @@ def release_workflow(
...
+
+@workflow.command("flow-ui")
+@click.argument("flow")
+@click.option("--out", "out_path", default=None, help="Write here instead of stdout.")
+@click.option(
+    "--custom",
+    "custom_paths",
+    multiple=True,
+    help="JSON of extra custom_node_data entries (repeatable). Accepts the output of "
+    "`integration tool-node`, or a bare {node_id: entry} mapping.",
```
vitalii-dynamiq: ### The `integration tool-node` command referenced here does not exist

**Medium severity**

`integration.py` registers `list`, `connect`, `accounts`, `connect-project`, `app`, `components`, `component`, `component-props` and `component-options` — there is no `tool-node`. Nor is there a `pipedream_node` binary, referenced at line 330 here and at `flowcheck.py:470` and `flowcheck.py:483`.

The remnants of the dropped command are still in the tree: `PIPEDREAM_NODE_TYPE` (`integration.py:31`), `pipedream_fetch` (`integration.py:347`) and `map_configurable_props` (`integration.py:367`) have no callers.

This is not only stale help text. The warning at line 721 fires when a Pipedream tool has no `custom_node_data` entry and tells the user to "Build each with `integration tool-node` and pass it via --custom" — running that gets `Error: No such command 'tool-node'`. `--custom` is the only way to supply the `pipedreamApp` / `pipedreamComponent` records, and nothing in the CLI produces them, so the tool saves and then draws on the canvas with no logo, account picker or configuration form, and the user has no way out of the loop.

Either restore the command or point these three messages at whatever now produces `custom_node_data`.

---

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,523 @@
...
+                warnings.append(
+                    f"node {label!r} is a standalone step that runs AFTER agent {after[0]!r} and receives "
+                    'its finished prose. To give the agent a tool it can call, move this into that '
+                    "agent's \"tools\" array instead."
+                )
+
+    for path, text in walk_strings(flow):
+        if path.rsplit(".", 1)[-1] in PROSE_KEYS or len(text) > 200:
+            continue
+        if looks_like_placeholder(text):
```
vitalii-dynamiq, line 446: ### Any short string containing `<` blocks the save

**Medium severity**

`PLACEHOLDER_HINTS` includes a bare `"<"` (line 28), so `looks_like_placeholder` returns True for every string under 200 characters that contains that character anywhere, unless its key is one of the seven `PROSE_KEYS`. The result is appended to `errors`, not `warnings`.

Slack's message format is the common case: a pinned `input_props.text` of `"Hi <@U12345>"` is ordinary, correct content, and `text` is not a prose key.

```python
flowcheck.validate({"nodes": [
    {"id": "input", "type": "dynamiq.nodes.utils.Input"},
    {"id": "a", "type": "dynamiq.nodes.agents.Agent", "depends": [{"node": "input"}],
     "tools": [{"type": "dynamiq.nodes.tools.Pipedream", "id": "t1", "name": "slack",
                "input_props": {"text": "Hi <@U12345>"}}]},
]})
# $.nodes[1].tools[0].input_props.text is still the placeholder 'Hi <@U12345>' - replace it with a real value.
```

`workflow validate` exits 1 and `workflow verify` raises, and the user cannot comply: the value they are told to replace is already the real one. Channel mentions (`<#C123>`), HTML in an email body and XML-ish content in any non-prose field hit the same rule.

Match an actual template shape — `<...>` with no whitespace, or a leading `<` — rather than the character on its own, or demote this one hint to a warning.

---

vitalii-dynamiq (commented):
Reviewed 2ee579b: 6 findings — 1 high, 5 medium.

---

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/commands/workflow.py
maksymbuleshnyi: Fixed

dynamiq/cli/commands/workflow.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/commands/knowledgebase.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/commands/workflow.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/client.py
```diff
@@ -60,6 +106,10 @@ def delete(
     ) -> Any:
         return self._request("DELETE", path, headers=headers, json=json, data=data)
 
+    def _request_once(self, method, path, **kwargs):
+        """One attempt, no retry. Same request, same error handling."""
+        return self._request.retry_with(stop=stop_after_attempt(1))(self, method, path, **kwargs)
```
vitalii-dynamiq: ### `_request_once` raises `RetryError` on a 502/503/504 instead of returning the response

**Medium severity**

`retry_with(stop=stop_after_attempt(1))` narrows only the stop condition. The `retry` predicate is carried over intact, including `retry_if_result(lambda r: r is not None and r.status_code in _RETRY_STATUS)`.

So on a 502/503/504 the predicate says "retry", the stop says "stop", and tenacity takes its stop path. `reraise=True` only re-raises when the outcome is an exception; here the outcome is a successful *result*, so there is nothing to re-raise and tenacity raises `RetryError` instead. Reproduced against the pinned tenacity 9.1.2:

```python
c = C()                      # _request stubbed to return a 503
c._request_once("POST", "/v1/workflows/test")
# RetryError[<Future at 0x... state=finished returned Resp>]
```

`workflow test` (`workflow.py:569`) is the only caller. A gateway 503 therefore propagates a `RetryError` out of `api.post`: nothing after that line runs, so the response body never reaches `echo_response` and the user gets a traceback naming a tenacity Future instead of the error the gateway returned. The docstring's contract — "One attempt, no retry. Same request, same error handling." — holds for exceptions but not for the retryable status codes.

Override the `retry` predicate alongside `stop` so a single attempt returns its response whatever the status code.

---

dynamiq/cli/commands/deployment.py
```diff
@@ -0,0 +1,298 @@
...
+@with_api_and_settings
+def test_inference(*, api: ApiClient, settings: Settings, inference_id: str, prompt: str):
+    """Send one chat completion - the proof a deployment actually serves.
+
+    The endpoint is OpenAI-compatible, so anything that speaks that API can call it directly.
+    Only meaningful for a `text_generation` deployment.
+    """
+    echo_response(
+        api.post(
+            f"/v1/inferences/{inference_id}/chat/completions",
```
vitalii-dynamiq, line 224: ### `inference test` re-runs the generation five times and then fails on any completion over 30s

**Medium severity**

This POST leaves both new `post()` knobs at their defaults — `retry=True`, no `timeout` — so it lands on `_request` with `timeout: float = 30.0` under `stop_after_attempt(5)` and `retry_if_exception_type(requests.RequestException)`.

`requests.exceptions.ReadTimeout` is a `RequestException`. A chat completion that runs longer than 30 seconds — routine for a large model or a long prompt — is therefore re-POSTed four more times, each attempt starting a fresh generation on the deployment's GPU, and the command ends by re-raising `ReadTimeout` as a traceback rather than printing a completion. A deployment whose generations are slower than 30s can never pass this check, which the docstring describes as "the proof a deployment actually serves".

`post()`'s own docstring names this hazard, and `workflow test` guards against it with `timeout=EXECUTION_TIMEOUT, retry=False` (`workflow.py:569`). `voice simulate` (`voice.py:146`) has the same gap and is worse on the side-effect axis: a simulation run is created per attempt, so a slow simulation leaves up to five records behind before the traceback.

Give both calls an execution-sized timeout and `retry=False`.

---

vitalii-dynamiq (commented):
Reviewed cc23a4af28089b3551dc00d8cd5fc77008903185: 2 medium findings.

---

dynamiq/cli/commands/voice.py
```diff
@@ -0,0 +1,172 @@
...
+@voice.command("simulate")
+@click.argument("agent_id")
+@click.argument("payload")
+@with_api_and_settings
+def simulate(*, api: ApiClient, settings: Settings, agent_id: str, payload: str):
+    """Run a simulated call against the agent - a test that needs no phone.
+
+    Use this as the "does it work" gate before telling anyone the agent is ready.
+    """
+    echo_response(api.post(f"{BASE}/agents/{agent_id}/simulations", json=read_json_arg(payload)))
```
acoola: [P2] Disable automatic retries for simulation creation

This POST inherits the default retry policy, which resubmits on a ReadTimeout. If the server has already accepted the simulation and only its response times out, a retry can start a second run; the client has no idempotency key to identify it as the same request. A mocked timeout followed by 200 produced two POSTs and a successful CLI exit. Pass `retry=False` here, as `workflow test` does, and check the other new non-idempotent creation/execution commands for the same issue.

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,559 @@
...
+                    "like an id (lowercase letters, digits, single hyphens). Use e.g. "
+                    f"{slug_for(name)!r}. "
+                    "Human-readable text belongs in `role`/`description`."
+                )
+
+        if node_type.startswith("dynamiq.nodes.agents."):
+            llm = node.get("llm")
+            if not isinstance(llm, dict):
+                errors.append(f"agent {label!r} has no `llm` object.")
+            elif requirement_problems(llm.get("connection"), f"agent {label!r}: llm.connection") is not None:
```
acoola, line 296: [P2] Carry requirement references through workflow normalization

The new requirement support accepts an object such as `{"$type": "requirement", "$id": "<requirement UUID>"}` as a connection, but `normalize_flow()` still calls `check_connection()`, which requires a UUID string. Consequently, a flow can pass the new validator and then fail in `workflow save` before any HTTP request; the same normalization is used by create/test and explicit release flows. Reproduced this with a requirement-backed LLM connection. Update normalization to preserve valid requirement objects so this new support is usable through the CLI.

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,559 @@
...
+                f'{{"selector": {{"field": "$.node.output.x"}}}}, got {type(transformer).__name__}. '
+                "The JSONPath goes inside `selector`, not on `input_transformer` itself."
+            )
+            transformer = {}
+        selector = transformer.get("selector") or {}
+        if not isinstance(selector, dict):
+            errors.append(f"node {label!r}: input_transformer.selector must be an object of field -> JSONPath.")
+            selector = {}
+        for field, expression in selector.items():
+            if not isinstance(expression, str) or not expression.startswith("$."):
```
acoola: [P2] Match the runtime's supported selector semantics

The runtime accepts more than `$.<node-id>...`: bracket notation such as `$['input'].output.message`, literal selector values, and paths relative to `input_transformer.path` are supported. For example, `{"path": "$.input.output", "selector": {"input": "$.message"}}` correctly resolves to the message, but this checker treats `message` as a missing node. All three cases were reproduced through `Node.transform()` and rejected by `validate()`. Honor the path-selection step and supported JSONPath/literal forms to avoid false failures in `workflow validate` and `workflow verify`.

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,559 @@
...
+        problems.append(
+            f"{where}: needs `input_props.configurableProps` - the FULL prop list from the "
+            f"component, not a subset. Build it with `pipedream_node <app> {tool.get('action_id') or '<key>'} "
+            "--out tool.json`, which fetches the real record instead of reconstructing it."
+        )
+    else:
+        mistyped = [
+            prop.get("name")
+            for prop in declared_props
+            if isinstance(prop, dict) and "type_" not in prop and "type" in prop
```
acoola: [P2] Accept Pipedream property schemas using `type`

This reports a property such as `{"name": "text", "type": "string"}` as an error, claiming the SDK ignores it. However, `Pipedream.__init__()` explicitly converts `type` to `type_` with `rename_keys_recursive()`, and `get_input_schema()` converts it back when constructing the fields. The input representation is therefore supported by the node. Accept both keys or normalize them before checking, otherwise valid Pipedream configurations fail the new validate/verify commands.

dynamiq/cli/client.py
acoola: [P2] Return gateway responses from the single-attempt path

Reducing Tenacity to one attempt does not bypass its result predicate. A 502/503/504 still matches `retry_if_result`, so exhausting the single attempt raises `RetryError` even with `reraise=True`, rather than returning the response. Reproduced `workflow test` receiving 503: `echo_response()` never runs and the readable HTTP status/body is replaced by an uncaught RetryError. Call the undecorated request implementation for the no-retry path, or otherwise disable retry evaluation, so callers can handle the response normally.

acoola (dismissed):
Found five issues in the new CLI behavior: execution retries can submit duplicate work, requirement references do not survive the save path, validation rejects supported selector and Pipedream schema formats, and single-attempt requests lose gateway error details. Details are attached inline.

Validation: CLI help smoke check and focused reproductions with mocked HTTP requests; multipart replay preserved file contents for both dictionary and list payloads. Full testing was limited by an unrelated Stagehand dependency mismatch in the local environment.

acoola (commented):
Submitting this as a non-blocking Comment review. The five inline findings remain attached to the relevant code in the earlier review: https://github.com/dynamiq-ai/dynamiq/pull/923#pullrequestreview-5134479958. The earlier Request changes recommendation has been dismissed to correct the review type.

dynamiq/cli/commands/workflow.py
```diff
@@ -136,13 +147,24 @@ def require_project(settings: Settings) -> str:
...
+    """
+    if isinstance(value, dict) and value.get("$type") == "requirement":
+        if not value.get("$id"):
+            raise click.ClickException(
+                f"requirement reference on {where} has no `$id`. Declare it with "
+                "`workflow requirement-add` and reference the id it returns."
+            )
         return
-    if not isinstance(value, str) or not _is_uuid(value):
+    if not _is_uuid(str(value or "")):
```
vitalii-dynamiq, line 182: ### `check_connection` lost its `None` guard, so every node without a connection now fails

**High severity**

The base version returned early on a missing connection (`if value is None: return`, `workflow.py:140` on `e709f67`). That guard is gone, and the replacement folds `None` into the UUID branch: `str(None or "")` is `""`, `uuid.UUID("")` raises `ValueError`, `_is_uuid` returns False, and the `ClickException` fires.

`normalize_flow` calls this unconditionally for every node, every entry in `tools`, and `llm` (lines 202-209) — none of which is guarded by the node actually having a `connection`. `Input`, `Output`, and `Pipedream` tools carry no `connection` at all, so every real DAG trips it.

The CLI's own documented minimal invocation breaks on the flow it generates for itself:

```
$ dynamiq workflow create '{"name": "my-workflow"}'
Error: connection None on node 'input' is not a connection UUID. Run `dynamiq connection list` ...
```

`create_workflow` fills in `starter_flow()` (line 506, one `dynamiq.nodes.utils.Input` node with no `connection`) and hands it to `normalize_flow` on the next line. `save` (line 527), `test`, `release` and `flow-ui` (line 762) all call `normalize_flow` first thing and fail the same way on any flow containing a connectionless node, which is every flow with an Input node.

Restore the early return for `None` before the UUID check.

---

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,568 @@
...
+        if missing:
+            problems.append(
+                f"{where}: input_props prop(s) {', '.join(map(str, missing))} declare no type. "
+                'Each needs `type_` (or `type`, which the SDK renames): {"name": "title", '
+                '"type_": "string"}.'
+            )
+
+    props = tool.get("configurable_props")
+    if not isinstance(props, dict) or not props:
+        nested = find_nested(tool, "configurable_props")
```
vitalii-dynamiq, line 588: ### An empty `configurable_props` is reported as nested inside itself

**Medium severity**

`find_nested` starts its walk at the object it is given and matches the key on the first level (`flowcheck.py:447-449`: `for k, v in obj.items(): if k == key: return f"{path}.{k}"`). Passing `tool` therefore matches `tool["configurable_props"]` itself.

The guard above only requires a *non-empty dict*, so `configurable_props` present but empty, `null`, or a non-dict falls into this branch with the key sitting at the top level:

```python
tool = {"type": "dynamiq.nodes.tools.Pipedream", "action_id": "notion-create-page",
        "configurable_props": {}}
# find_nested(tool, "configurable_props") -> 'tool.configurable_props'
```

The user is then told:

```
... `configurable_props` exists but at tool.configurable_props - it belongs at the TOP level
of the tool object, a sibling of `action_id`, not nested inside another key. Move it up one level.
```

The instruction cannot be followed — the key is already a sibling of `action_id` — and because this branch returns, the correct diagnostic in the `else` (needs `configurable_props` binding the account, with the `authProvisionId` example and the `integration accounts` pointer) never runs. `"configurable_props": {}` and `"configurable_props": null` are exactly what a partially scaffolded Pipedream tool carries, so this hits the case the message was written for. `validate` exits 1 and `verify` raises with a self-contradictory message in place of the actionable one.

Start the search below the top level, or check `key in tool` first and fall through to the `else` when it is.

---

vitalii-dynamiq (commented):
Reviewed 1073eb6: 1 high, 1 medium.

---

dynamiq/cli/client.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/commands/voice.py
maksymbuleshnyi: Fixed

dynamiq/cli/commands/deployment.py
maksymbuleshnyi: Fixed

dynamiq/cli/client.py
maksymbuleshnyi: Fixed

dynamiq/cli/commands/workflow.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: fixed

dynamiq/cli/commands/integration.py
```diff
@@ -136,3 +149,197 @@ def connect_project(*, api: ApiClient, settings: Settings, app_slug: str):
...
+    GET /v1/pipedream/connect/apps, filtered to one entry. This is the platform's own
+    catalogue, and it is deliberately not Pipedream's `/v1/apps/<slug>`: the catalogue entry
+    carries `actions[]` and `triggers[]` alongside `id`, `name`, `name_slug`, `img_src` and
+    `description`, and the whole entry is what the editor keeps as a node's `pipedreamApp`.
+    It is also the only source of the logo the canvas draws.
+
+    Verified against a builder-authored workflow: its stored `pipedreamApp` has this shape,
+    down to the nested action and trigger lists.
+    """
+    response = api.get("/v1/pipedream/connect/apps", params={"page_size": 500})
```
vitalii-dynamiq: ### `integration app` searches one page and rejects every app past it

**Medium severity**

This reads a single page of `/v1/pipedream/connect/apps` and then treats it as the whole catalogue. `page_size=500` is the maximum the API accepts — `pagination_options` documents it as "max 500; API default 25" (`workflow.py:49`) — so the response is capped at 500 entries however many apps exist, and Pipedream's connect catalogue is in the thousands.

The response carries the count needed to detect this. `echo_list` reads it off the same shape for the sibling `/v1/pipedream/connect/accounts` endpoint and warns when the page is partial (`workflow.py:101-105`):

```python
pagination = body.get("pagination") or {}
total = pagination.get("total_count")
items = body.get("data") or []
if total is not None and len(items) < total:
    click.echo(f"note: showing {len(items)} of {total}. Use --all to fetch every page.", err=True)
```

Here `payload.get("data", payload)` takes the items and drops `pagination` on the floor, so an app on page 2 falls through `next(...)` as `None` and the command hard-fails:

```
Error: no app matching 'zoom' in the catalogue.
```

The "did you mean" hint is built by scanning the same truncated page, so it comes back empty and the message names no alternative. Per this command's own docstring the catalogue entry is what a node stores as `pipedreamApp` and "the only source of the logo the canvas draws", so for any app beyond the first page there is no way to obtain that record through the CLI.

Walk the pages until `len(items) >= pagination["total_count"]`, as `echo_list`'s `--all` branch does, or push the filter to the server if the endpoint accepts a slug parameter.

---

vitalii-dynamiq (commented):
Reviewed e6c8cab: 1 medium-severity finding.

---

dynamiq/cli/commands/integration.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,588 @@
...
+                                f"agent {label!r}: memory.backend.memory_id "
+                                f"{backend.get('memory_id')!r} is not a UUID. Create one with "
+                                "`POST /v1/memories` and use the id it returns."
+                            )
+                    if memory.get("save_mode") not in (None, "full", "input_output"):
+                        errors.append(
+                            f"agent {label!r}: memory.save_mode {memory.get('save_mode')!r} is not valid. "
+                            'Use "full" or "input_output".'
+                        )
+                    if not ({"user_id", "session_id"} & set(selector)):
```
vitalii-dynamiq: ### The memory check rejects agents that receive `user_id` without a selector

**High severity**

This keys entirely on `input_transformer.selector`, but that is not what decides whether memory switches on. `Agent.run` reads the field off its own input data — `use_memory = self.memory and (input_data.user_id or input_data.session_id)` (`dynamiq/nodes/agents/base.py:759`) — and `user_id`/`session_id` are ordinary declared fields on `AgentInputSchema` (`base.py:182-183`). A selector is only one of several ways to populate them.

`Node.transform_input` (`dynamiq/nodes/node.py:664-668`) shows the two shapes this rejects:

- **No `input_transformer` at all.** The `else` branch runs `inputs = input_data | {...depends...}`, where `input_data` is the flow-level input — the payload posted to run the workflow. `{"input": "...", "user_id": "u1"}` therefore reaches the agent as `input_data.user_id` and memory engages. This is the canonical shape: an agent depending directly on Input with no transformer.
- **`input_transformer.path` with no selector.** Line 664 applies the transformer when *either* `path` or `selector` is set, and `transform` runs `jsonpath_filter(data, path)` before the mapper. `{"path": "$.input.output"}` hands the agent the Input node's whole output dict, `user_id` included. This function already models that form 100 lines above (`relative = bool(transformer.get("path"))`, line 229), but the memory check does not consult it.

Since `selector` is `{}` in both cases, `{"user_id", "session_id"} & set(selector)` is empty and this appends to `errors`, not `warnings`. `workflow validate` exits 1 and `workflow verify` (`workflow.py:750`) raises `ClickException` on a flow that saves and runs with working memory. The message asserts memory "can never switch on — it is ignored at runtime with no error", which is false for both shapes, and its remedy tells the user to add a selector mapping they do not need.

Treat the check as satisfied when the agent has no `input_transformer`, or when `input_transformer.path` is set, and otherwise demote it to a warning.

---

vitalii-dynamiq (commented):
Reviewed ef82334: 1 high severity finding.

---

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,597 @@
...
+        problems.append(
+            f'{where}: a Pipedream tool needs `action_id` (e.g. "notion-create-page"). '
+            "List the real keys with `dynamiq integration components <app_slug>`."
+        )
+    external_user = tool.get("external_user_id")
+    from_requirement = requirement_problems(external_user, f"{where}: external_user_id")
+    if from_requirement is not None:
+        problems.extend(from_requirement)
+    elif not external_user:
+        problems.append(f"{where}: needs `external_user_id` (the project id). The CLI fills this in on save.")
```
vitalii-dynamiq: ### The `external_user_id` error promises a fill-in that the CLI never performs

**Medium severity**

Nothing in `dynamiq/cli` ever writes `external_user_id` into a flow node. `normalize_flow` (`commands/workflow.py`) is the only place a flow is mutated before it is sent, and it fills `flow.id`, node `name`, `depends`, `selector` and sub-object ids — not this. The only other occurrences in the package are `integration.py:298` and `integration.py:334`, which put the project id into a Pipedream *API request body*, not into a node. `grep -rn external_user_id dynamiq/cli/` returns exactly these plus this check.

The field is genuinely required — `Pipedream.external_user_id: str` has no default (`dynamiq/nodes/tools/pipedream.py:62`) and is sent in the action-run payload (`pipedream.py:168`) — so the check is right to flag it. The remedy sentence is what fails, and it directs the user into persisting a broken flow:

1. `workflow validate` reports this via `problems`, which line 378 splices into `errors`, so the command exits 1.
2. The message says the CLI fills it in on save, so the user runs `workflow save`. Nothing injects the field, and the flow is persisted as written.
3. `workflow verify` then runs `flowcheck.validate` over the saved flow (line 826) and raises `ClickException(f"{len(errors)} problem(s) in the saved flow.")` at line 845 — after the save has already happened.

A Pipedream tool is now stored missing a field its node requires at construction time. Either inject the project id in `normalize_flow` alongside the other fixes, or state the real remedy: add `"external_user_id"` to the tool with the project id from `dynamiq config`.

---

vitalii-dynamiq (commented):
Reviewed a47669b: 1 medium finding.

---

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

dynamiq/cli/flowcheck.py
```diff
@@ -0,0 +1,602 @@
...
+                        )
+
+            # STRUCTURED_OUTPUT without a schema is the mode change without the guarantee.
+            response_format = node.get("response_format")
+            if response_format is not None and not isinstance(response_format, dict):
+                errors.append(
+                    f"agent {label!r}: `response_format` must be a JSON Schema object, got "
+                    f"{type(response_format).__name__}."
+                )
+            elif isinstance(response_format, dict) and not response_format.get("properties"):
```
vitalii-dynamiq: ### The `response_format` check rejects the litellm-wrapped shape the Agent explicitly accepts

**Medium severity**

`.get("properties")` is read off the value as written, but the agent normalizes before storing it. `Agent.response_format` carries a `mode="before"` validator (`dynamiq/nodes/agents/agent.py:246-260`) that routes every non-`None` input through `schema_generator.unwrap_response_format`, whose docstring states it strips `{"type": "json_schema", "json_schema": {"schema": ...}}` — "matching litellm's response_format shape" — and whose implementation returns `inner["schema"]` for exactly that form (`components/schema_generator.py:56-61`).

So the wrapped form is a supported input. Its top-level keys are `type` and `json_schema`, never `properties`, so this branch fires on a schema that is well-formed and does constrain the answer:

```python
flowcheck.validate({"nodes": [
    {"id": "input", "type": "dynamiq.nodes.utils.Input"},
    {"id": "a", "type": "dynamiq.nodes.agents.Agent", "depends": [{"node": "input"}],
     "llm": {"type": "dynamiq.nodes.llms.OpenAI", "model": "gpt-4o",
             "connection": "11111111-1111-1111-1111-111111111111"},
     "response_format": {"type": "json_schema", "json_schema": {"name": "ans", "strict": True,
         "schema": {"type": "object", "properties": {"answer": {"type": "string"}},
                    "required": ["answer"]}}}},
]})
```

```
agent 'a': `response_format` has no `properties`, so it constrains nothing. Use e.g. {"type": "object", "properties": {"answer": {"type": "string"}}, "required": ["answer"]}.
```

This lands in `errors`, not `warnings`, so `workflow validate` exits 1 and `workflow verify` (`workflow.py:845`) raises `ClickException` on a workflow that saves and runs with the structured output working. The suggested remedy also tells the user to write the raw form, discarding the `name`/`strict` fields litellm reads. The repository writes schemas in the wrapped form itself (`examples/components/agents/orchestrators/graph_orchestrator/code_assistant.py:30-47`).

Unwrap the value the way `unwrap_response_format` does before looking for `properties`.



---

vitalii-dynamiq (commented):
Reviewed 8e228b26391ced66b252c72704ff4464e85ecebb — 1 medium finding.

---

dynamiq/cli/flowcheck.py
maksymbuleshnyi: Fixed

mykhailobuleshnyi (dismissed):

dynamiq/cli/client.py
```diff
@@ -19,6 +19,53 @@ class HTTPError(RuntimeError):
...
+
+
+def ok(response) -> bool:
+    """True for any 2xx.
+
+    Checking `status_code == 200` alone reports a created resource as a failure: `POST /v1/apps`
+    answers 201 and the app is deployed, but the caller sees `HTTP 201: {...}` raised as an
+    error and cannot tell it apart from a real one.
+    """
+    return 200 <= response.status_code < 300
```
vitalii-dynamiq, line 66: ### The `ok()` migration did not reach `_request`, so every 201 logs a false failure

**Low severity**

`_request` still logs on `status_code != 200` (`client.py:174`):

```python
if response.status_code != 200:
    logging.error(f"{method} {path} failed with {response.status_code}: {response.text.strip()}")
```

`dynamiq/utils/logger.py:6` calls `logging.basicConfig`, so the root logger carries a stderr handler and this line prints.

The docstring right here states the premise: `POST /v1/apps` answers 201 and the app is deployed. `app_create` posts to exactly that endpoint (`commands/app.py:166`) and hands the response to `ok()`, which passes it. The user gets both lines:

```
ERROR:root:POST /v1/apps failed with 201: {"data": {...}}
{
  "data": { ... }
}
```

Every 201/202/204 the new commands produce prints a failure line next to its own success output. Compare against `ok(response)` here, as the rest of the PR now does.

---

vitalii-dynamiq (commented):
Reviewed 4b0b321: 1 low.

---

tyaroshko (dismissed):

dynamiq/cli/client.py
maksymbuleshnyi: Fixed

tyaroshko (approved):

PR #924, opened 2026-09-07
fix(tools): treat unknown todo ids as inserts on merge

mykhailobuleshnyi:


dynamiq/nodes/tools/todo_tools.py
```diff
@@ -176,15 +179,15 @@ def execute(
...
-                )
-
             for todo in new_todos:
-                existing_by_id[todo["id"]]["status"] = todo["status"]
+                stored = existing_by_id.get(todo["id"])
+                if stored is None:
+                    # Unknown id is an insert, not an error: the first write of a run has
+                    # nothing to merge into, and an agent that discovers work mid-run must
+                    # be able to add a task without resending (and rewording) the whole list.
+                    existing_by_id[todo["id"]] = todo
```
vitalii-dynamiq, line 188: ### A swallowed load failure now silently destroys the stored todo list

**Medium severity**

`_load_todos()` catches every exception and returns `[]` (lines 133-135), and also returns `[]` when the parsed `todos` key is not a list (lines 123-125). It cannot distinguish "no todos yet" from "the read failed". Every concrete backend reaches this through a remote call — `E2BSandbox`, `DaytonaSandbox` and the Bedrock sandboxes each implement `exists`/`retrieve` over the network — so a transient retrieve/decode failure, or a truncated `todos.json` from an interrupted save, is reachable.

With that empty list, a routine `merge=true` status update finds no id in `existing_by_id`, takes this insert branch for each item, and `final_todos` becomes only the items in that one call. `_save_todos` then writes with `overwrite=True`, so every todo not named in that call is gone from the file permanently, and the response reports the truncated list as the current plan.

Before this change the unknown-id guard raised on exactly this state, so the file was left intact and the agent could retry. Distinguishing a failed load from an empty one — having `_load_todos` signal the failure and skipping the merge-write when it does — keeps the new insert behavior without turning a read blip into data loss.

---

dynamiq/nodes/tools/todo_tools.py
```diff
@@ -91,9 +90,13 @@ class TodoWriteTool(Node):
...
   {"id": "2", "content": "ignored", "status": "in_progress"}], "merge": true}
 
+ADD (merge=true): An id that does not exist yet is appended, using the content you send.
+  {"todos": [{"id": "3", "content": "Fix rate limiting", "status": "pending"}], "merge": true}
+
 RULES:
 - Use merge=false ONLY for initial list creation. First task should be "in_progress", rest "pending".
-- Use merge=true for ALL subsequent updates — only status is applied, content stays unchanged.
+- Use merge=true for ALL subsequent updates — only status is applied to existing todos, content stays unchanged.
+- Use merge=true to add newly discovered work; never resend the whole list just to add one item.
```
vitalii-dynamiq, line 99: ### The agent system prompt still forbids the capability this rule adds

**Low severity**

`TODO_TOOLS_INSTRUCTIONS` in `dynamiq/nodes/agents/prompts/secondary_instructions.py:45` says:

> After initial creation, ONLY update status via merge=true - do not restructure the plan

`PromptManager` appends that block whenever `config.todo_management_enabled` (`dynamiq/nodes/agents/prompts/manager.py:322-323`), which is the same condition under which this tool is available. The model therefore receives both this new rule and a categorical instruction not to do what it describes, in one context. An agent that follows the system prompt never adds a mid-run todo, so the new insert path is unreachable in normal agent use; one that follows the tool description contradicts its own instructions. Updating that line to permit adding newly discovered work makes the change effective.

---

vitalii-dynamiq (commented):
Reviewed cea2e8d: 2 findings (1 medium, 1 low).

---

PR #925, opened 2026-09-07
feat(mcp): pass per-run HTTP headers via tool_params

delamainer:
## Summary
- Allow a shared agent to send per-run MCP HTTP headers through `tool_params` (including by owning `MCPServer` name/id) without mutating the connection, so concurrent users stay isolated.
- Headers are hidden from the LLM, merged over connection-level headers on `connect()`, and stripped from MCP tool arguments.
- Stdio connections accept the same `headers` argument and ignore it.

## Test plan
- [ ] `uv run pytest tests/integration/nodes/tools/test_mcp_tool.py tests/unit/nodes/tools/test_mcp_pipedream_mocking.py`
- [ ] Run an agent with an `MCPServer` and `tool_params.by_name[<server name>].headers` and confirm the header is on the HTTP request, not in `tools/call` arguments
- [ ] Confirm two concurrent runs with different headers do not overwrite `connection.headers`


dynamiq/nodes/tools/mcp.py
```diff
@@ -343,11 +343,49 @@ def build(self, schema_dict: dict[str, Any], model_name: str) -> type[BaseModel]
...
+    if field is None or not is_mcp_http_headers_field(field):
+        return input_dict, None
+    headers = input_dict.pop(MCP_HTTP_HEADERS_FIELD, None)
+    return input_dict, headers or None
+
+
+def attach_mcp_http_headers_field(model: type[BaseModel]) -> type[BaseModel]:
+    """Add a hidden ``headers`` field unless the MCP tool already declares that name."""
+    if MCP_HTTP_HEADERS_FIELD in model.model_fields:
+        return model
```
vitalii-dynamiq, line 384: ### Per-run headers are sent as tool arguments when the MCP tool declares its own `headers` property

**Medium severity**

`attach_mcp_http_headers_field` returns the model unchanged when the tool's JSON Schema already declares `headers`, so that field carries no `is_accessible_to_agent: False` marker. `split_mcp_http_headers` then takes the `field is None or not is_mcp_http_headers_field(field)` branch and returns the whole payload as tool arguments with `None` headers.

A `tool_params` entry such as `{"headers": {"Authorization": "Bearer <user token>"}}` targeting such a tool is therefore serialized into the `tools/call` arguments payload instead of the HTTP request: the per-run credential never reaches the transport and the call goes out with only the connection-level headers. Reproduced on this branch — for a schema with `properties: {url, headers}`, `split_mcp_http_headers` returns `({'url': 'http://x', 'headers': {'Authorization': 'Bearer user-token'}}, None)`. If the declared type is not dict-compatible (`"type": "string"`), the same configuration instead fails schema validation at run time. Neither case is reported when the collision happens.

---

dynamiq/nodes/tools/mcp.py
```diff
@@ -343,11 +343,49 @@ def build(self, schema_dict: dict[str, Any], model_name: str) -> type[BaseModel]
...
 
 
+MCP_HTTP_HEADERS_FIELD = "headers"
+MCP_HTTP_HEADERS_SCHEMA_EXTRA = {"is_accessible_to_agent": False}
+
+
+def is_mcp_http_headers_field(field: Any) -> bool:
+    """True when ``headers`` is Dynamiq's hidden per-call HTTP header field, not a tool argument."""
+    extra = getattr(field, "json_schema_extra", None)
+    return isinstance(extra, dict) and extra.get("is_accessible_to_agent") is False
```
vitalii-dynamiq: ### A tool's own `headers` argument hidden via `input_param_modes` is diverted to the transport

**Medium severity**

The marker for "this is Dynamiq's field" is the plain name `headers` plus `is_accessible_to_agent is False`, and neither part is exclusive to this feature. `apply_param_modes` sets exactly that flag on any field marked hidden (`dynamiq/nodes/schema_utils.py:100-102`), so an MCP tool that declares a real `headers` argument and is configured with `input_param_modes={"headers": "hidden"}` has that argument popped out of the tool-call payload by `split_mcp_http_headers` and injected into the HTTP connection instead.

Reproduced on this branch: with `properties: {url, headers}` and `apply_param_modes(model, {"headers": "hidden"})`, `split_mcp_http_headers` returns `({'url': 'http://x'}, {'Accept': 'text/plain'})` — the server never receives the argument it declared. Marking the Dynamiq-owned field with a dedicated key in `json_schema_extra`, and checking that key here, keeps the two apart.

---

dynamiq/nodes/tools/mcp.py
```diff
@@ -343,11 +343,49 @@ def build(self, schema_dict: dict[str, Any], model_name: str) -> type[BaseModel]
...
+    return create_model(
+        model.__name__,
+        __base__=model,
+        headers=(
+            dict[str, Any] | None,
+            Field(
+                default=None,
+                description="HTTP headers for this MCP request. Not forwarded as a tool argument.",
+                json_schema_extra=MCP_HTTP_HEADERS_SCHEMA_EXTRA,
+            ),
```
vitalii-dynamiq: ### Per-run credentials travel through node input, which is traced and logged unredacted

**Medium severity**

Connection-level MCP headers are shielded from tracing: `BaseConnection.to_dict(for_tracing=True)` returns only `{"id", "type"}` (`dynamiq/connections/connections.py:86-88`). This field routes the same class of secret through node input data, which has no such shielding. `Node.run` hands the validated input to `run_on_node_start` (`dynamiq/nodes/node.py:1358`), and `TracingCallbackHandler.on_node_start` stores `format_value(input_data, for_tracing=True)` (`dynamiq/callbacks/tracing.py:363`); `format_value` does no key-based redaction, so the header dict is serialized into the run trace as-is. `Agent._apply_parameters` additionally appends `f"  - From {source}: Set {key}={value}"` to the debug lines emitted when `verbose` is enabled (`dynamiq/nodes/agents/base.py:1458`, `1695`).

A per-user bearer token supplied through `tool_params.by_name[<server>].headers` — the exact usage this PR documents — therefore ends up in exported traces and verbose logs in cleartext.

---

vitalii-dynamiq (commented):
Reviewed fa276e442e10c18bc9d77e9feabe823f3b8d35ec: 3 findings (3 medium).

---

dynamiq/nodes/agents/base.py
```diff
@@ -1658,9 +1658,15 @@ def _run_tool(
...
 
                 # 2. Apply parameters by tool name (medium priority)
+                # MCPServer is replaced by the tools it discovers, so match by_name/by_id against
+                # the owning server as well. Callers can key headers (and other params) once by
+                # server name/id instead of listing every remote tool.
+                owner = getattr(tool, "_owner_server", None)
                 name_params_any = (
                     tool_params.by_name_params.get(tool.name)
                     or tool_params.by_name_params.get(self.sanitize_tool_name(tool.name))
+                    or (owner and tool_params.by_name_params.get(owner.name))
```
vitalii-dynamiq: ### Server-keyed headers are silently dropped when the same tool also has a tool-keyed `tool_params` entry

**Medium severity**

The owner-server lookup is spliced into an `or` chain, so resolution is first-match-wins rather than a merge. `tool_params.by_name_params.get(tool.name)` is evaluated first, and any non-empty dict there short-circuits the chain before `tool_params.by_name_params.get(owner.name)` is reached.

With `by_name = {"github-mcp": {"headers": {"Authorization": "Bearer <user token>"}}, "create_issue": {"labels": ["bug"]}}`, the `create_issue` tool discovered from the `github-mcp` `MCPServer` resolves to `{"labels": ["bug"]}` only. `headers` is never applied to `merged_input`, `split_mcp_http_headers` yields `None`, and the call goes out with the shared connection-level credentials instead of the per-run ones — the isolation this PR adds is lost for that tool, with no error raised. Every other tool on the same server, having no tool-keyed entry, does get the header, so the failure is per-tool and easy to miss.

This contradicts the intent stated in the comment directly above: keying params once by server name does not compose with per-tool entries. The `by_id` chain at line 1684 has the same shape. Merging the server-level dict under the tool-level dict, rather than selecting between them, is what the documented usage requires.

---

vitalii-dynamiq (commented):
Reviewed 9511753: 1 medium severity finding.

---

dynamiq/nodes/tools/mcp.py
delamainer: Addressed in 7995798d by using a dedicated `mcp_http_headers` field (and `tool_params` key) instead of `headers`. A tool-declared `headers` argument stays in `tools/call`; per-run HTTP headers go on the connection.

dynamiq/nodes/tools/mcp.py
delamainer: Addressed by identifying Dynamiq's field with `json_schema_extra.is_mcp_http_headers` (7995798d). Hiding a tool's own `headers` via `input_param_modes` no longer diverts it to the HTTP connection; covered in 04024af0.

dynamiq/nodes/tools/mcp.py
delamainer: Addressed in 80b43f82. `format_value(..., for_tracing=True)` redacts `mcp_http_headers` (including nested `tool_params`), and verbose `_apply_parameters` logs `***` instead of the header values.

dynamiq/nodes/agents/base.py
delamainer: Addressed in c1bcd05b. Matching `by_name` / `by_id` entries are applied in order (owner server first, then the tool), so server-level `mcp_http_headers` still apply when the tool also has its own `tool_params` entry.

dynamiq/nodes/tools/mcp.py
```diff
@@ -357,11 +357,50 @@ def resolve_root_object_schema(
...
+        ),
+    )
+
+
 def create_input_schema_from_json_schema(
     schema_dict: dict[str, Any], model_name: str = "MCPToolSchema", definitions: dict[str, Any] | None = None
 ) -> type[BaseModel]:
     """Create a Pydantic input-schema model from an MCP tool's JSON Schema."""
-    return _SchemaModelBuilder(dict(definitions or {})).build(schema_dict, model_name)
+    return attach_mcp_http_headers_field(_SchemaModelBuilder(dict(definitions or {})).build(schema_dict, model_name))
```
vitalii-dynamiq: ### The MCP-only `mcp_http_headers` field is injected into every Composio tool schema

**Medium severity**

`attach_mcp_http_headers_field` is applied inside `create_input_schema_from_json_schema`, which is not MCP-specific: `Composio.get_input_schema` calls it too (`dynamiq/nodes/tools/composio.py:178`). Every Composio action now gets the hidden field, and Composio has no code path that honours it.

Three consequences follow on the Composio side:

- `_generate_description` (`composio.py:105-125`) iterates `input_schema.model_fields` and lists every optional field without checking `is_accessible_to_agent`, so `Composio.__init__` bakes `- mcp_http_headers (dict[str, Any] | None): HTTP headers for this MCP request...` into the description the LLM sees for every action.
- If a value is set — by the LLM it was just advertised to, or via `tool_params.global_params` — `_build_request` does `model_dump(by_alias=True, exclude_none=True)`, so it is posted to Composio's API inside `arguments`, and `execute` logs it at `composio.py:245` in cleartext.
- Three existing assertions break: `tests/unit/nodes/tools/test_composio.py:154` (`sorted(model_fields) == ["addr", "note", "stops"]`) and the two `model_fields == {}` cases at lines 190 and 200.

Confirmed by running the function on this branch: an empty Composio schema comes back with `['mcp_http_headers']`, the nested one with `['addr', 'mcp_http_headers', 'note', 'stops']`, and `model_dump(by_alias=True, exclude_none=True)` yields `{'addr': {...}, 'mcp_http_headers': {'Authorization': 'Bearer tok'}}`.

Attaching the field in `MCPTool.get_input_schema` (line 472) rather than in the shared builder keeps it on the tools that consume it.

---

dynamiq/utils/utils.py
```diff
@@ -426,6 +433,11 @@ def format_value(
         else:
             base_dict = value.model_dump()
 
+        if for_tracing and isinstance(base_dict, dict):
+            base_dict = {
+                k: TRACING_REDACTED_PLACEHOLDER if k in TRACING_REDACTED_KEYS else v for k, v in base_dict.items()
+            }
```
vitalii-dynamiq: ### Trace redaction misses `tool_params`, the documented way to pass per-run headers

**High severity**

The dict branch above redacts at any depth because it recurses, but this BaseModel branch rewrites only the top-level keys of `base_dict` and returns it without recursing. That is the branch the agent's own input takes.

`tool_params` is typed `ToolParams | None` (`dynamiq/nodes/agents/base.py:188`), so after validation it is a model instance, not a dict. `Node.run` passes `dict(transformed_input)` to `run_on_node_start` (`dynamiq/nodes/node.py:1358`), and `dict(model)` leaves nested models as models. `TracingCallbackHandler.on_node_start` then calls `format_value(input_data, for_tracing=True)` (`dynamiq/callbacks/tracing.py:363`): the outer dict's `tool_params` value hits this branch, `model_dump()` produces top-level keys `global_params` / `by_name_params` / `by_id_params` — none in `TRACING_REDACTED_KEYS` — and the nested `by_name_params[<server>]["mcp_http_headers"]` is returned untouched.

Ran the real `format_value` against that shape on this branch:

```
format_value({"input": "hi", "tool_params": ToolParams.model_validate(
    {"by_name": {"github-mcp": {"mcp_http_headers": {"Authorization": "Bearer user-token"}}}})},
    for_tracing=True)

-> {'input': 'hi', 'tool_params': {'global_params': {},
     'by_name_params': {'github-mcp': {'mcp_http_headers': {'Authorization': 'Bearer user-token'}}},
     'by_id_params': {}}}
```

So `tool_params.by_name[<server name>].mcp_http_headers` — the usage this PR's description and test plan call out — writes the per-user bearer token into the agent node's trace in cleartext. `tests/unit/utils/test_format_value.py` passes a plain dict, which takes the recursing dict branch, so it does not cover this path. Recursing into `base_dict` after the model dump (or redacting on the already-formatted result) closes it.

---

vitalii-dynamiq (commented):
Reviewed 29ecd29284bb95b6a44f723886db6169ad899859: 2 findings (1 high, 1 medium).

---

dynamiq/nodes/tools/mcp.py
tyaroshko: Addressed in 66888244. `attach_mcp_http_headers_field` moved out of the shared `create_input_schema_from_json_schema` and into `MCPTool.get_input_schema`, so only MCP tools — the ones with a transport that honours it — carry the field. Composio schemas are back to their declared properties (the four `test_composio.py` assertions pass again), and a regression test asserts the shared builder does not add it.

dynamiq/utils/utils.py
tyaroshko: Addressed in 66888244. Redaction is now recursive: a `redact_tracing_keys` helper walks the already-serialized structure, and the model branch applies it to `base_dict` instead of rewriting only its top-level keys. `tool_params.by_name[<server>].mcp_http_headers` now traces as `***`.

The `RunnableResult` branch above had the same hole for the same reason — `RunnableResult.to_dict` formats its members without `for_tracing`, so a result carrying a node input would have leaked the header dict — so it redacts its result too. `tests/unit/utils/test_format_value.py` covers both paths with a real `ToolParams` instance and a `RunnableResult`.

dynamiq/utils/utils.py
```diff
@@ -16,6 +16,28 @@
 
 CHARS_PER_TOKEN = 4
 
+# Values for these keys are credentials; traces must not persist them in cleartext.
+TRACING_REDACTED_KEYS = frozenset({"mcp_http_headers"})
```
vitalii-dynamiq, line 20: ### Per-run MCP headers are written to DEBUG logs in cleartext

**Medium severity**

`TRACING_REDACTED_KEYS` is consulted only by `format_value` and by `Agent._apply_parameters` (`base.py:1457`). The node lifecycle logging path uses a different serializer, `format_value_for_log` (`utils.py:188`), which recurses through dicts and BaseModels replacing only `BytesIO`/`bytes` and never checks the redaction set.

`Node.execute_with_retry` calls `self.log_execution_start(input_data)` (`node.py:1613`, and `node.py:1738` on the async path) with `transformed_input` — the validated input-schema instance, which for an MCP tool carries the hidden `mcp_http_headers` field. `log_execution_start` renders it through `_dump_for_log` → `format_value_for_log` (`node.py:1186`) whenever the logger is enabled for DEBUG.

Reproduced on this branch with a schema built by `attach_mcp_http_headers_field`:

```
inst = M(query='hi', mcp_http_headers={'Authorization': 'Bearer user-token'})

format_value_for_log(inst)      -> {'query': 'hi', 'mcp_http_headers': {'Authorization': 'Bearer user-token'}}
Node._dump_for_log(inst)        -> {'mcp_http_headers': {'Authorization': 'Bearer user-token'}, 'query': 'hi'}
format_value(inst, for_tracing=True) -> {'query': 'hi', 'mcp_http_headers': '***'}
```

So the per-user bearer token this PR routes through node input is redacted in traces and in the agent's verbose line, but is written verbatim to the application log at DEBUG. Applying `redact_tracing_keys` to the result of `format_value_for_log`, or checking `TRACING_REDACTED_KEYS` inside its dict and BaseModel branches, closes the path.

---

vitalii-dynamiq (commented):
Reviewed 66888244f3874c6778446fe3c24c05e3d2f117c8: 1 medium finding.

---

dynamiq/utils/utils.py
tyaroshko: Addressed in 5c2a0f1b. `format_value_for_log` now consults `TRACING_REDACTED_KEYS` in its dict and BaseModel branches, via a small `_log_value_or_redaction` helper, so the node lifecycle logs redact the same keys the trace does. Re-ran the reported shape:

```
format_value_for_log(inst) -> {'query': 'hi', 'mcp_http_headers': '***'}
Node._dump_for_log(inst)   -> {'mcp_http_headers': '***', 'query': 'hi'}
```

`tests/unit/utils/test_format_value.py` covers both the validated MCP input instance and a nested dict. The constant's comment now reads "traces and logs", since it is no longer trace-only.

dynamiq/utils/utils.py
```diff
@@ -16,6 +16,28 @@
...
+TRACING_REDACTED_PLACEHOLDER = "***"
+
+
+def redact_tracing_keys(value: Any) -> Any:
```
vitalii-dynamiq, line 24: ### Per-run MCP headers are written to DEBUG logs in cleartext

**High severity**

`redact_tracing_keys` and `TRACING_REDACTED_KEYS` are consulted only by `format_value` and by `Agent._apply_parameters`. The node lifecycle logging path uses a different serializer, `format_value_for_log` (`utils.py:188`), whose `BytesIO`, `bytes`, `dict`, `list`, `tuple` and `BaseModel` branches replace only file-like values and never check the redaction set.

`Node.run` validates input into a schema instance (`node.py:1354`) and passes it to `execute_with_retry` (`node.py:1374`), which calls `log_execution_start(input_data)` (`node.py:1615`, and `node.py:1739` on the async path); that renders through `_dump_for_log` → `format_value_for_log` (`node.py:1186`) whenever the logger is enabled for DEBUG. For an MCP tool the instance carries the hidden `mcp_http_headers` field; for the agent node `tool_params` is a `ToolParams` model that the `BaseModel` branch recurses into, exposing the nested header dict. `_LOG_REPR.maxstring` is 200 and `maxdict` is 20, so a typical bearer token is rendered in full rather than truncated.

Reproduced on this branch against a schema built by `attach_mcp_http_headers_field`:

```
inst = M(query='hi', mcp_http_headers={'Authorization': 'Bearer user-token-abcdef'})

format_value_for_log(inst)           -> {'query': 'hi', 'mcp_http_headers': {'Authorization': 'Bearer user-token-abcdef'}}
Node._dump_for_log(inst)             -> {'mcp_http_headers': {'Authorization': 'Bearer user-token-abcdef'}, 'query': 'hi'}
format_value(inst, for_tracing=True) -> {'query': 'hi', 'mcp_http_headers': '***'}
```

So the per-user credential this PR routes through node input is redacted in traces and in the agent's verbose line, but written verbatim to the application log at DEBUG. Applying `redact_tracing_keys` to the result of `format_value_for_log`, or checking `TRACING_REDACTED_KEYS` inside its dict and `BaseModel` branches, closes the path.

An earlier thread on this file reported this defect and was resolved citing commit `5c2a0f1b`. That commit is not in this PR's history and no `_log_value_or_redaction` helper exists in the tree, so the fix was never pushed and the leak is live on `455eab6`.

---

vitalii-dynamiq (commented):
Reviewed 455eab6a4fbd33381a929cbb5ef73ccc5b9791a9 — 1 high severity finding.

---

dynamiq/utils/utils.py
tyaroshko: Addressed in a4fbac28, now on the branch head — apologies for the bounce: the earlier thread was resolved citing `5c2a0f1b`, which had been committed but never pushed, so you were right that the fix was not in the tree and the leak was live on `455eab6`.

`format_value_for_log` now consults `TRACING_REDACTED_KEYS` in its dict and BaseModel branches (including `model_extra`), via a `_log_value_or_redaction` helper, so the node lifecycle logs redact the same keys the trace does. Re-ran the reported shape on `a4fbac28`:

```
format_value_for_log(inst) -> {'query': 'hi', 'mcp_http_headers': '***'}
Node._dump_for_log(inst)   -> {'mcp_http_headers': '***', 'query': 'hi'}
```

Both entry points you named are covered: the MCP tool's validated input instance, and the agent node's `tool_params` — the `BaseModel` branch recurses into the `ToolParams` model, and the nested header dict is now replaced before `_LOG_REPR` ever renders it. `tests/unit/utils/test_format_value.py::test_format_value_for_log_redacts_mcp_http_headers` covers the instance and the nested-dict shape.

The shared `TRACING_REDACTED_KEYS` is reused rather than a second constant, so the three consumers — trace, agent verbose line, node logs — stay in sync; its comment now reads "traces and logs".

dynamiq/nodes/agents/base.py
```diff
@@ -1658,31 +1686,42 @@ def _run_tool(
...
-                    or (resolved_agent and tool_params.by_name_params.get(self.sanitize_tool_name(resolved_agent.name)))
+                # MCPServer is replaced by the tools it discovers, so match by_name/by_id against
+                # the owning server as well. Callers can key mcp_http_headers (and other params)
+                # once by server name/id instead of listing every remote tool. Server-level dicts
+                # are applied first, then the tool's own entry, so the two merge rather than one
+                # short-circuiting the other.
+                owner = getattr(tool, "_owner_server", None)
+                name_lookups: list[tuple[str, Any]] = []
+                if owner and owner.name:
+                    name_lookups.append((f"name:{owner.name}", tool_params.by_name_params.get(owner.name)))
```
vitalii-dynamiq: ### Per-run credentials keyed by server name reach every `MCPServer` sharing the default name `"mcp"`

**High severity**

`MCPServer.name` is `name: str = "mcp"` (`dynamiq/nodes/tools/mcp.py:602`), so an operator who adds two MCP servers to one agent and does not rename them ends up with two distinct servers whose `name` is the same string. Nothing rejects that: `initialize_tools` replaces each server with its discovered `MCPTool`s and stamps `mcp_tool._owner_server = self` (`mcp.py:646`), and `tool_by_names` is keyed by tool name (`base.py:2550`), so server names are never checked for uniqueness.

This lookup then resolves the owner by name alone. For a tool discovered from server A and a tool discovered from server B, `owner.name` is `"mcp"` in both cases, so `tool_params.by_name_params.get("mcp")` returns the same dict for both and `_apply_tool_param_lookups` writes it into each tool's `merged_input`. `split_mcp_http_headers` pops `mcp_http_headers` and `execute_async` passes it to `self.connection.connect(headers=...)` (`mcp.py:547-549`), where `merge_mcp_http_headers` puts it on that server's HTTP request.

So `tool_params.by_name["mcp"] = {"mcp_http_headers": {"Authorization": "Bearer <user token>"}}` — the exact usage this PR documents for keying headers once per server — sends the end user's bearer token to server B's endpoint as well as server A's. Before this change `by_name` matched only tool names, which are distinct per remote tool, so no per-run credential could be routed to an unintended host by a name collision. Matching the owner by `owner.id` only, or requiring the resolved owner name to be unique among the agent's servers before applying, keeps a header on the server it was meant for.

---

vitalii-dynamiq (commented):
Reviewed a4fbac2 — 1 high severity finding.

---

dynamiq/nodes/agents/base.py
tyaroshko: Addressed in 7f77bc61. Confirmed the leak first — two `MCPServer`s left at the default name, one `by_name["mcp"]` entry, and the same bearer token went to both hosts:

```
server names: mcp / mcp | ids differ: True
search_a -> https://tenant-a.example.com/sse gets {'Authorization': 'Bearer USER-TOKEN'}
search_b -> https://tenant-b.example.com/sse gets {'Authorization': 'Bearer USER-TOKEN'}
```

Fixed a little differently than suggested. Matching the owner by `owner.id` only would also drop the convenience this PR is for — keying headers once per server by name — for every correctly-named setup. Instead `_owner_server_name_is_unambiguous` checks that the name picks out exactly one server among the run's tools (`_runtime_tools`), and the name match is skipped when it does not. Ids are unaffected, since they are unique by construction.

When the name is ambiguous the entry is dropped **with a warning** naming the server and pointing at `by_id`, rather than silently. Silently dropping fails the operator's intent invisibly; silently applying is the leak. Neither is a good default without a signal.

Three tests in `tests/integration/nodes/tools/test_mcp_tool.py`: the collision keeps the header off *both* servers and warns, `by_id` still reaches only the intended server, and a unique server name still applies as documented.

TrachukT (approved):

PR #926, opened 2026-09-07
fix(agents): keep the todo list visible across compaction

mykhailobuleshnyi:


maksymbuleshnyi (dismissed):

dynamiq/nodes/agents/components/history_manager.py
```diff
@@ -178,6 +183,11 @@ def _compact_history(
             if notes_path := self.get_notes_file_path():
                 summary = f"{summary}\n\n{NOTES_REVISIT_INSTRUCTION_TEMPLATE.format(notes_path=notes_path)}"
 
+            # The todo-write result carrying the list is itself summarized away, and there is no
+            # read tool - so restate the list here or the agent loses track of its own plan.
+            if todos := getattr(getattr(self, "state", None), "todos", None):
```
vitalii-dynamiq: ### Restated todo list is a stale snapshot from the start of the loop

**Medium severity**

`self.state.todos` is only ever populated by `_refresh_agent_state`, which runs at the *top* of each iteration (`agent.py:1856`, `agent.py:1863`) by reading `todos.json` from the todo backend. `_try_summarize_history` runs at the *bottom* of the same iteration (`agent.py:1972`). `TodoWriteTool.execute` writes only to the file store (`todo_tools.py:211`) and never touches `state.todos`, so a todo-write performed during loop N is not reflected in `state.todos` until loop N+1.

Consequences when compaction fires at the end of a loop that contained a todo-write:

- Statuses are one update behind. The injected message asserts "restated below with its **current** statuses. Continue from it", so after the real tool result has been summarized away the agent is left with an authoritative-sounding list that says a task it just completed is still `in_progress`, and can redo it.
- If that loop is the one that *created* the list, `state.todos` is still `[]` (the file did not exist at `_refresh_agent_state` time), the walrus guard is falsy, and nothing is injected at all — the feature silently no-ops in exactly the case it exists for.

The message is appended as a `static=True` prompt message, so the stale snapshot persists in the prompt for the rest of the run. Reload the todos from the backend here rather than reading the loop-start cache.


---

dynamiq/nodes/tools/todo_tools.py
```diff
@@ -92,9 +102,11 @@ class TodoWriteTool(Node):
   {"id": "2", "content": "ignored", "status": "in_progress"}], "merge": true}
 
 RULES:
-- Use merge=false ONLY for initial list creation. First task should be "in_progress", rest "pending".
-- Use merge=true for ALL subsequent updates — only status is applied, content stays unchanged.
-- Do NOT restructure, reword, or reorder todos when updating status.
+- FIRST call of the session: merge=false, always. Never merge=true against a list you have not created.
+- Use merge=true for ALL subsequent status updates — every id you send must already exist.
+- To ADD newly discovered work, send merge=false with the complete list: the existing items with
```
vitalii-dynamiq: ### New "add work with merge=false" rule contradicts the system prompt

**Low severity**

This rule is new in this PR and directly contradicts `secondary_instructions.py:53`, which still reads "After initial creation, ONLY update status via merge=true - do not restructure the plan". Both strings are fed to the same model: `TODO_TOOLS_INSTRUCTIONS` goes into the system prompt, this text into the tool description.

Before this PR the two agreed ("Use merge=false ONLY for initial list creation"); the PR updated one side and not the other. An agent that follows the system prompt when it discovers new work sends `merge=true` with an id that does not exist, and hits `Todo ids not found` at `todo_tools.py:199` — a wasted loop on the path this rule was added to enable. Update line 53 to match.


---

vitalii-dynamiq (commented):
Reviewed 23bf14091b25cfa2167017799856ea47c14d603c — 2 findings: 1 medium, 1 low.

---

dynamiq/nodes/agents/components/history_manager.py
vitalii-dynamiq: ### Restated todo list is stale by one loop and presented as current

**High severity**

`self.state.todos` is only ever populated by `_refresh_agent_state` (agent.py:2162), which runs at the *start* of each loop (agent.py:1863). Auto-compaction runs at the *end* of the loop body (`_try_summarize_history`, agent.py:1972). Nothing in the `todo-write` execution path writes back to `state.todos` — the tool persists to the backing file only.

So when the loop that trips the token limit is the one that called `todo-write` — which the instructions actively encourage after every step ("Mark completed IMMEDIATELY after finishing each step") — the sequence is: refresh loads the pre-write list into `state.todos` → the model marks step N complete → the tool result carrying the authoritative list is summarized away → `_compact_history` restates the *pre-write* list under `TODO_REVISIT_TEMPLATE`, which asserts "restated below with its **current** statuses. Continue from it". The agent then sees step N as still `in_progress` and redoes work it already finished.

The same ordering breaks the first-compaction case outright: if the list was created in that loop, `state.todos` is still empty, the walrus guard is falsy, and no list is restated at all — the exact loss this PR is meant to prevent.

Prior to this change `state.todos` had no reader (it was written only for checkpointing and end-of-run cleanup), so the staleness was harmless; making it the source of a prompt-visible statement is what surfaces it. Reload the todos from the backend immediately before restating rather than relying on the start-of-loop snapshot.

---

dynamiq/nodes/agents/prompts/secondary_instructions.py
```diff
@@ -38,9 +38,17 @@
...
+
 TODO_TOOLS_INSTRUCTIONS = """## Todo Management
 - Use the todo-write tool for complex 3+ step tasks; skip for simple requests
-- The todo-write result lists the full current todo list - re-read it there
+- Your todo list starts EMPTY. The first todo-write of the session MUST use merge=false - this
+applies even when you are recording steps you have already completed
+- The todo-write result lists the full current todo list - re-read it there; if history is
+compacted, the list is restated in the summary
 - When creating initial list: first task "in_progress", rest "pending"
 - After initial creation, ONLY update status via merge=true - do not restructure the plan
```
vitalii-dynamiq: ### System instruction contradicts the new tool description on adding todos

**Low severity**

This line still says "After initial creation, ONLY update status via merge=true - do not restructure the plan", but the tool description rewritten in this PR now says "To ADD newly discovered work, send merge=false with the complete list" (todo_tools.py:107-108). Both strings reach the model in the same prompt — `TODO_TOOLS_INSTRUCTIONS` via `manager.py:323`, the tool description via the tool schema.

Before this PR the two agreed ("Use merge=false ONLY for initial list creation"); the diff changed one side and left this one. An agent that follows this line when it discovers new work will send the new id with `merge=true`, which hits the unknown-id branch at todo_tools.py:195 and raises `Todo ids not found: [...]`, burning a loop.

---

vitalii-dynamiq (commented):
Reviewed 23bf14091b25cfa2167017799856ea47c14d603c — 2 findings: 1 high, 1 low.

---

dynamiq/nodes/agents/agent.py
```diff
@@ -2146,22 +2146,33 @@ def _refresh_agent_state(self, loop_num: int) -> None:
         """
         self.state.update_loop(loop_num)
 
+        todos = self.load_current_todos()
+        if todos is not None:
+            self.state.update_todos(todos)
```
vitalii-dynamiq, line 2151: ### Todo parsing moved outside the try/except that used to guard it

**Medium severity**

Before this change the whole read was one block: `self.state.update_todos(data.get("todos", []))` sat *inside* the `try`, with `except Exception` logging `"Failed to load todo state (none or invalid)"`. The extraction split them — `load_current_todos()` keeps the `try` but only around the file read and `json.loads`, and returns the raw list; the `TodoItem` construction now happens in `update_todos` (`agent.py:172`, `TodoItem(**t)`), which runs *outside* any handler.

So a file that is readable and valid JSON but whose `todos` value is not a list of well-formed items no longer degrades — it raises out of `_refresh_agent_state`. Both call sites (`agent.py:1856` before the loop, `agent.py:1863` inside it) are unguarded, so the exception ends the run:

- `{"todos": [{"id": "1", "status": "pending"}]}` (missing required `content`) → `ValidationError`
- `{"todos": {"1": {...}}}` → iterating yields the key strings → `TypeError: argument after ** must be a mapping`

This also contradicts the new function's own contract: the docstring states "None means disabled or unreadable", but a readable-and-invalid file now propagates instead of returning None. The tool's own loader treats this state as expected — `_load_todos` (`todo_tools.py:120-140`) checks `isinstance(todos, list)` and skips items that fail validation with a warning.

Constructing the items inside `load_current_todos`'s `try` restores the old behavior and matches the docstring.

---

vitalii-dynamiq (commented):
Reviewed bb16f10e64163324819709dd9687135619ac6fb6: 1 medium finding.

---

PR #929, opened 2026-09-09
feat(agents): add a persistent, cross-conversation store for agents

mykhailobuleshnyi:


dynamiq/nodes/agents/base.py
```diff
@@ -2263,6 +2300,85 @@ def file_store_backend(self) -> FileStore | None:
...
+        when ``agent_file_write_enabled``. Standalone or beside a sandbox, the persistent store
+        carries its own ``write_enabled`` - which is how a sandbox agent gets writable memory over a
+        read-only workspace.
+        """
+        if not self.persistent_stores:
+            return False
+        if self.file_store_backend:
+            return self.file_store.agent_file_write_enabled
+        # One writable memory is enough to justify the tool; the prompt says which accept writes.
+        return any(config.write_enabled for config in self.persistent_stores)
```
vitalii-dynamiq: ### `write_enabled=False` on a memory is never enforced

**High severity**

`write_enabled` only decides whether *a* write tool is attached; it never constrains which routes that tool can reach. The single `FileWriteTool` is bound to the `CompositeFileStore` spanning every memory, and neither the composite nor `FileWriteTool` consults the per-route config — `write_enabled` is referenced nowhere else except to render the word "Read-only." into a description.

With `persistent_store=[handbook(write_enabled=False), personal(write_enabled=True)]`, this property returns `True`, `memory-write` is attached, and a write to `memories/handbook/hacked.md` resolves through the route and lands in the read-only backend. Reproduced against this HEAD:

```
routes: ['memories/handbook/', 'memories/personal/']
composite.store("memories/handbook/hacked.md", ...) -> SUCCESS
handbook store now holds: ['memories/handbook/hacked.md']
```

The file-store branch is worse: it returns `file_store.agent_file_write_enabled` and ignores every config's `write_enabled` entirely, so `file-write` reaches all memories whenever workspace writes are on. A read-only memory (a shared handbook, a curated corpus) is therefore protected only by prompt wording, which the model can violate at any time. Enforcement needs to live at the write path — reject a store/delete whose resolved route is read-only.

---

dynamiq/storages/file/composite.py
```diff
@@ -0,0 +1,213 @@
...
+            if candidate.startswith(prefix) and len(prefix) > len(match):
+                owner, match = store, prefix
+        if owner is not None:
+            return owner, normalized
+
+        # A directory above routes belongs to a single store only when it holds exactly one. Above
+        # several - `memories/` over `memories/handbook/` and `memories/me/` - it belongs to none of
+        # them, and the caller merges instead of picking whichever route happened to come first.
+        below = [store for prefix, store in self.routes.items() if prefix.startswith(candidate)]
+        return (below[0], normalized) if len(below) == 1 else None
```
vitalii-dynamiq: ### Listing `memories/` silently drops files when one store owns the directory

**Medium severity**

`_route_for_directory` returns a single owner, and `list_files` then delegates the whole listing to it, skipping both the default store and any deeper routes. Two reachable configurations lose files, and this is exactly the listing the memory protocol tells the agent to perform first (`ALWAYS {list_tool} memories/ BEFORE ANYTHING ELSE`).

One memory namespaced under the root (`path_prefix="user"`), routed under a file store: `below` has exactly one entry, so the memory backend answers alone and the workspace's own `memories/` files are invisible.

```
c = CompositeFileStore(default=default, routes={"memories/user/": user})
c.store("memories/notes.md", ...)      # -> default store
c.store("memories/user/pref.md", ...)  # -> user store
c.list_files("memories/", recursive=True) -> ['memories/user/pref.md']   # notes.md missing
c.list_files("", recursive=True)          -> ['memories/notes.md', 'memories/user/pref.md']
```

A root memory plus a nested one (`path_prefix=""` and `path_prefix="team"`) is worse, because the longest-prefix branch picks the parent route and the nested memory disappears entirely:

```
c = CompositeFileStore(default=..., routes={"memories/": root_mem, "memories/team/": team_mem})
c.list_files("memories/", recursive=True) -> ['memories/mine.md']   # memories/team/ours.md missing
```

The agent concludes the memory is empty and re-asks the user for facts it already recorded, or overwrites them. The merge path at the bottom of `list_files` already handles this correctly — the early single-owner return should only be taken when no other route or the default store can hold files under the requested directory.

---

dynamiq/nodes/agents/base.py
```diff
@@ -2263,6 +2300,85 @@ def file_store_backend(self) -> FileStore | None:
...
+
+        Composing is idempotent: a caller may build the ``CompositeFileStore`` themselves and still
+        declare ``persistent_store`` (which is what turns on the prompt protocol) without ending up
+        with one composite nested inside another.
+        """
+        stores = self.persistent_stores
+        if not stores:
+            return self.file_store.backend
+
+        routes = {config.normalized_prefix: config.backend for config in stores}
```
vitalii-dynamiq: ### Memories sharing a prefix collapse, silently discarding a backend

**Medium severity**

The dict comprehension keys on `normalized_prefix`, so two memories that resolve to the same prefix leave only the last backend in `routes` — no error, no warning. `path_prefix` defaults to `""`, and `name`/`description` are presented as the way to tell memories apart, so `[PersistentStoreConfig(backend=a, name="handbook"), PersistentStoreConfig(backend=b, name="personal")]` is a natural declaration that both map to `memories/`.

The result is not a no-op: `describe_memory_namespaces` still lists both memories to the model (two lines, same path), so the agent is told the handbook exists and files facts "into" it, while every read and write actually resolves to `b`. The first backend is unreachable for the lifetime of the agent, and nothing surfaces the loss. Raising on a duplicate prefix in `_compose_file_store_backend` (or validating the list on the config) turns a silent misroute into a startup error.

---

dynamiq/nodes/agents/prompts/manager.py
```diff
@@ -321,6 +370,19 @@ def build_react_prompt(self, config: ReactPromptConfig) -> None:
                 ops_parts.append(PERSISTENT_NOTES_INSTRUCTIONS_TEMPLATE.format(notes_path=config.notes_file_path))
         if config.todo_management_enabled:
             ops_parts.append(TODO_TOOLS_INSTRUCTIONS)
+        if config.persistent_store_enabled:
+            # The environment block renders before this one (see AGENT_PROMPT_TEMPLATE), and it
+            # calls the sandbox the agent's long-term memory. Contradict it here, immediately
+            # above the protocol, so the correction is the later of the two.
+            if config.sandbox_base_path:
```
vitalii-dynamiq, line 359: ### Sandbox correction names `memory-write` when no write tool is attached

**Low severity**

This branch checks only `sandbox_base_path`, not `persistent_store_writable`. For a sandbox agent with a read-only memory (`persistent_store=PersistentStoreConfig(..., write_enabled=False)`), `_persistent_store_writable` is `False`, so `build_persistent_file_tools` attaches only `memory-read` and `memory-list` — yet the prompt still renders "`memories/` is, and only `memory-write` reaches it", immediately followed by the read-only block stating the store is read-only for the agent.

The model is handed a tool name that does not exist in its tool list, alongside a direct contradiction of it, and will attempt `memory-write` and take a tool-not-found error. Gating this block on `config.persistent_store_writable` alongside `sandbox_base_path` keeps the correction without inventing the tool.

---

vitalii-dynamiq (commented):
Reviewed f5c2a569c0a27a4c26036ced90e4353c07b5b35e: 1 high, 2 medium, 1 low.

---

examples/components/agents/agents/agent_memory_store.py
```diff
@@ -0,0 +1,103 @@
...
+
+def agent_with_two_memories() -> Agent:
+    """Several memories, each behind its own prefix.
+
+    There is no catch-all: a path matching no prefix is refused, naming the prefixes that exist, so
+    a memory never lands somewhere arbitrary. The agent still gets exactly one tool.
+    """
+    team = DynamiqMemoryStore(
+        connection=DynamiqConnection(),
+        memory_store_id=os.getenv("DYNAMIQ_TEAM_MEMORY_STORE_ID", MEMORY_STORE_ID),
```
vitalii-dynamiq, line 72: ### `agent_with_two_memories()` always raises on the default configuration

**Low severity**

The team store falls back to `MEMORY_STORE_ID` when `DYNAMIQ_TEAM_MEMORY_STORE_ID` is unset, and it uses the same `USER_ID` and a second `DynamiqConnection()` pointing at the same URL. `DynamiqMemoryStore.identity()` is `(type, connection.url, memory_store_id, user_id)`, so both routes produce an identical identity.

`CompositeMemoryStore.normalize_route_prefixes` then hits its duplicate-store guard and raises `ValueError("Routes 'user/' and 'team/' address the same memory store. ...")` while constructing the backend, so the function cannot return an agent at all unless the reader happens to have `DYNAMIQ_TEAM_MEMORY_STORE_ID` exported. This is the only demonstration of the multi-memory feature, and the fallback default is precisely what the new validator rejects.

---

vitalii-dynamiq (commented):
Reviewed 2197b9d8cbe4f69f4423543ca303dd8cc77ddfee — 1 finding (1 low).

---

dynamiq/storages/memory/composite.py
```diff
@@ -0,0 +1,163 @@
...
+        """Put the mount point back on paths coming out of a store."""
+        return [entry.model_copy(update={"path": f"{route}{entry.path}"}) for entry in entries]
+
+    def list(self, prefix: str = "") -> list[MemoryEntry]:
+        """List memories, delegating into one route or merging every route at or below ``prefix``."""
+        normalized = normalize_path(prefix)
+        if normalized:
+            for route in sorted(self.routes, key=len, reverse=True):
+                if normalized.startswith(route):
+                    return self._mounted(route, self.routes[route].list(normalized[len(route) :]))
```
vitalii-dynamiq, line 135: ### Listing a nested route's own prefix delegates to its parent and returns nothing

**Medium severity**

`normalize_path` strips the trailing slash, so listing a route's own mount point never matches that route: `normalized` is `team/private`, which fails `startswith("team/private/")` and matches `"team/"` instead. The listing is then delegated to the *parent* store with inner prefix `private`, and the store actually mounted at `team/private/` is never consulted.

Writes take the opposite path — `_resolve` matches on the full key, so `team/private/pay.md` lands in the deep store (this is what `test_longest_prefix_wins` asserts). With `routes={"team/": team, "team/private/": deep}` at this HEAD:

```
write('team/private/pay.md')  -> deep store
list('team/private/')         -> delegate to TEAM store, inner prefix 'private'  => []
list('team/private/sub/')     -> delegate to DEEP store, inner prefix 'sub'      => correct
```

Only the boundary itself is broken, and the boundary is exactly the prefix the agent is handed: `describe_namespaces()` returns `team/private/`, and the tool description tells the model that `path` "narrows to a prefix", so `{"action": "list", "path": "team/private/"}` is the natural call. It reports "No memories under 'team/private/' yet." while memories exist there, and the agent re-asks for facts it already recorded or overwrites them. The merge branch below handles this case correctly — the delegation should only be taken when the requested prefix is strictly *inside* a route, not equal to it.

---

dynamiq/nodes/tools/memory_store_tool.py
```diff
@@ -0,0 +1,224 @@
...
+    brief: str = Field(default="Using memory", description="Short description of what you are doing.")
+
+    @model_validator(mode="after")
+    def validate_action_fields(self):
+        """Fail fast on the arguments each action needs, rather than part-way through execution."""
+        if self.action != MemoryStoreAction.LIST and not self.path:
+            raise ValueError(f"'path' is required for action '{self.action.value}'")
+        if self.action == MemoryStoreAction.WRITE and self.content is None:
+            raise ValueError("'content' is required for action 'write'")
+        if self.action == MemoryStoreAction.EDIT and (self.find is None or self.replace is None):
```
vitalii-dynamiq, line 77: ### An empty `find` passes validation and shreds the memory it edits

**Medium severity**

The check rejects `find` only when it is `None`, so `find=""` is accepted. `_edit` then evaluates `current.count("")`, which is `len(current) + 1` rather than 0, so the "does not appear" guard never fires. With `replace_all=true` the occurrence-count guard is skipped as well, and `current.replace("", replace)` interleaves the replacement between every character before `self.backend.write(path, updated)` persists it:

```
current = "Prefers British English."
find = "", replace = "X", replace_all = true
-> "XPXrXeXfXeXrXsX XBX..."   written back to the store
```

The original content is gone — this is a persistent store, so there is no working copy to recover from, and the tool reports success ("Updated '...' (25 replacements)."). The arguments come from the model, and the schema advertises both `find` and `replace_all` without constraining the former, so nothing between the tool call and the store rejects it. Treating an empty `find` the same as a missing one in this validator closes it.

---

vitalii-dynamiq (commented):
Reviewed fe5df47ad57229734f0eb791f363a46512070b1c — 2 findings (2 medium).

---

dynamiq/nodes/tools/memory_store_tool.py
```diff
@@ -0,0 +1,224 @@
...
+            ) from None
+
+    def _write(self, path: str, content: str) -> dict[str, Any]:
+        entry = self.backend.write(path, content)
+        return {"content": f"Remembered in '{entry.path}'."}
+
+    def _edit(self, input_data: MemoryStoreToolInputSchema) -> dict[str, Any]:
+        path, find, replace = input_data.path, input_data.find, input_data.replace
+        try:
+            current = self.backend.read(path)
```
vitalii-dynamiq: ### Concurrent `edit` calls silently lose one another's changes

**Medium severity**

`_edit` is an unguarded read-modify-write: it reads the whole document here, computes `updated` from that snapshot, and writes the whole document back at line 218. Nothing between the two serializes against another writer, and the backend offers no compare-and-set — `DynamiqMemoryStore.write` is an unconditional `PUT` with no version or ETag (`dynamiq/storages/memory/dynamiq.py:152`).

The store is deliberately shared across concurrent callers rather than isolated. `MemoryStore` sets `_clone_shared: ClassVar[bool] = True` (`dynamiq/storages/memory/base.py:59`), so when `Node.clone()` copies this tool for a parallel batch it hands every clone the same backend by reference. `MemoryStoreTool` does not override `is_clone_safe_for_parallel`, which defaults to `True`, so `dynamiq/nodes/agents/base.py:1824` clones it for parallel execution rather than serializing it the way the shared browser session does.

Two `edit` calls on one path in the same parallel batch therefore both read the pre-edit content, both apply their own replacement to it, and both write the full document. The second write erases the first replacement:

```
memory: "Prefers British English. Ships on Fridays."
edit A: find "British"  -> "Australian"
edit B: find "Fridays"  -> "Tuesdays"
both read the original; last write wins
result: only one of the two changes survives
```

Both calls report success (`Updated '...' (1 replacement).`), so the agent believes both facts were recorded. This is a persistent cross-conversation store with no working copy behind it, so the dropped edit is not recoverable and surfaces later as a memory the agent is confident it corrected. The same window is open across agents sharing one team memory store.

---

vitalii-dynamiq (commented):
Reviewed 084f98e5398580aeb9bc2929e900f17c122e5426 — 1 medium finding.

---

delamainer:
`user_id` on `DynamiqMemoryStore` should follow the same pattern as conversation `memory`, not be a required field on the store config.

For `memory`, `user_id` is per-run input (`input_data.user_id`). The agent only uses memory when it is present, and it is written onto each request as the tenant scope. We do not need `session_id` here — memory-store is user-scoped only — but `user_id` should still come from the run, not from `MemoryStoreConfig` / `DynamiqMemoryStore`.

That keeps isolation the same as chat memory: one store id in the agent JSON, and the caller supplies `user_id` at runtime. Baking it into the backend means a serialized agent is pinned to a single user and cannot be reused across users.

dynamiq/nodes/agents/prompts/manager.py
```diff
@@ -27,17 +27,44 @@
...
+        if config.memory_store_writable
+        else MEMORY_STORE_READONLY_INSTRUCTIONS_TEMPLATE
+    )
+    block = template.format(tool=config.memory_store_tool_name)
+
+    # Several memories, or one the caller bothered to describe: say what each holds, or the model
+    # has no way to choose between them.
+    listing = render_namespaces(config.memory_store_namespaces)
+    if listing:
+        block += f"\n\nYour memories, each holding something different:\n{listing}"
```
vitalii-dynamiq, line 68: ### Memory descriptions reach Jinja unescaped, unlike `role` and `instructions`

**Low severity**

`listing` is built from `MemoryStore.description`, caller-supplied text, and is concatenated straight into `block`. That block becomes the `operational_instructions` prompt block at line 361, and `generate_prompt` renders every block through Jinja: `Template(content).render(temp_variables)` (line 211). Nothing wraps it in `{% raw %}`.

Twelve lines below, this same function wraps `config.instructions` in `{% raw %}` (line 369), and step 5 does the same for `config.role` — this path is the one that missed the guard.

Confirmed against Jinja2:

```
desc = 'Team conventions, e.g. interpolation is written {{ name }}.'
Template('- user/ - ' + desc).render({})
-> '- user/ - Team conventions, e.g. interpolation is written .'

Template('- team/ - Rules for our {%raw_broken%} macros.').render({})
-> TemplateSyntaxError: Encountered unknown tag 'raw_broken'.
```

A `{{...}}` in a description is silently blanked, so the model is handed a route prefix with no explanation of what it holds and loses the basis for choosing between memories. A `{%...%}` raises `TemplateSyntaxError` out of `generate_prompt` on every run, making the agent unusable. Wrapping the rendered listing in `{% raw %}` matches what `role` and `instructions` already do.

---

dynamiq/nodes/tools/memory_store_tool.py
```diff
@@ -0,0 +1,232 @@
...
+        path, find, replace = input_data.path, input_data.find, input_data.replace
+        user_id = input_data.user_id
+        try:
+            current = self.backend.read(path, user_id)
+        except MemoryNotFoundError:
+            raise ToolExecutionException(
+                f"No memory at '{path}' to edit. Use action 'write' to create it.", recoverable=True
+            ) from None
+
+        occurrences = current.count(find)
```
vitalii-dynamiq: ### `str.count` misses overlapping matches, so an ambiguous `edit` is applied instead of refused

**Low severity**

The uniqueness guard asks "does `find` match exactly one place?" but answers it with `str.count`, which counts non-overlapping occurrences. When candidate positions overlap, `occurrences` is 1, both guards pass, and `str.replace(find, replace, 1)` consumes only the leftmost one:

```
current = 'Section separator: ----- then the next topic.'
find    = '---'
current.count(find)                                  -> 1
[i for i in range(len(current)) if current.startswith(find, i)]  -> [19, 20, 21]
current.replace(find, '===', 1)
  -> 'Section separator: ===-- then the next topic.'
```

`self.backend.write(path, updated, user_id)` then persists that, and the tool reports `Updated '...' (1 replacement).` The store is the cross-conversation one with no working copy behind it, so the mangled text is what the agent reads back next time.

`file_tools.py:40` already carries `_find_positions` built for exactly this, with a docstring stating the case: *"'aa' in 'aaa' reports two candidate positions where `str.count` sees one. Either could be the site the caller meant, which is what makes such a find string ambiguous."* Driving the zero/ambiguous checks off `_find_positions` and keeping `count` only for the replacement tally closes it.

---

vitalii-dynamiq (commented):
Reviewed 0d15f4b: 2 findings (2 low).

---

dynamiq/nodes/llms/base.py
```diff
@@ -997,7 +1010,12 @@ def _build_completion_params(
         if not tools and common_params.get("tool_choice") is not None:
             common_params.pop("tool_choice")
 
-        return self.update_completion_params(common_params)
+        params = self.update_completion_params(common_params)
+        # `update_completion_params` already applied the node's own config; a per-call one
+        # only fills in when the node has none, so the two can never double up.
+        if cache_control is not None and getattr(self, "cache_control", None) is None:
```
vitalii-dynamiq: ### A cross-provider fallback dies on the caller's `cache_control` before it ever reaches the provider

**High severity**

This branch applies a per-call `cache_control` whenever the node has none of its own, without checking that the config belongs to this provider. The fallback path hands it a config from a *different* provider: `_prepare_fallback_run` copies every kwarg except `run_depends` into the fallback's kwargs (line 1301), so the `cache_control` the agent passed for the primary is handed to `fallback.llm` verbatim. The fallback's own `cache_control` field defaults to `None`, so this branch is always taken.

The two config classes do not share fields — `Bedrock._apply_cache_control` reads `cache_control.cache_tools` (`bedrock.py:82`), `Anthropic._apply_cache_control` reads `cache_control.cache_system` (`anthropic.py:187`):

```
Bedrock fallback receives AnthropicCacheControl -> AttributeError: 'AnthropicCacheControl' object has no attribute 'cache_tools'
Anthropic fallback receives BedrockCacheControl -> AttributeError: 'BedrockCacheControl' object has no attribute 'cache_system'
```

`cache_tools` is read before the `params.get("tools")` guard, so it raises whether or not tools are present.

Newly reachable here: `Agent` now resolves a config at construction and passes it on every call (`agent.py:1797`), so an Anthropic agent with a Bedrock fallback (the natural redundancy pairing for Claude) carries an `AnthropicCacheControl` into the Bedrock node by default. Reproduced end to end against this HEAD with the primary failing on a rate limit:

```
agent._cache_control = type='ephemeral' ttl='5m' cache_injection_point_index=-1 cache_system=True
models actually dispatched to the provider: ['anthropic/claude-sonnet-4-5']
FALLBACK RESULT: FAILURE | AttributeError: 'AnthropicCacheControl' object has no attribute 'cache_tools'
```

The fallback is never dispatched — it fails inside `_build_completion_params`, `run_sync` logs "Fallback LLM failed" and returns the primary's error, so the configured resilience path is silently dead exactly when it is needed. Applying the per-call config only when it is an instance of this node's own `cache_control` type confines it to a matching provider.

---

dynamiq/nodes/agents/agent.py
```diff
@@ -188,6 +189,36 @@ def mark_tool_failure(content: str, success: Any) -> str:
...
+    Anthropic-family models (including on Bedrock) read back no cached prefix unless the
+    request carries an explicit breakpoint, unlike OpenAI/Gemini which cache automatically.
+    The field is discovered by annotation, so a node that gains one is picked up here.
+
+    Returns None when the node has no ``cache_control`` field, when the model does not
+    support caching, or when the caller set it themselves -- an explicit
+    ``cache_control=None`` is how you opt out.
+    """
+    field = type(llm).model_fields.get("cache_control")
+    if field is None or "cache_control" in llm.model_fields_set:
```
vitalii-dynamiq: ### A serialization round-trip reads as an opt-out, disabling caching for every YAML-loaded agent

**Medium severity**

`model_fields_set` is used to detect "the caller set it themselves", but it cannot distinguish a real opt-out from a field that was simply round-tripped. `Node.to_dict` calls `model_dump` without `exclude_none` and nothing excludes `cache_control`, so a serialized LLM always carries `cache_control: null`; `WorkflowYAMLLoader.init_node` passes the whole node dict back into the constructor (`serializers/loaders/yaml.py:844`), which marks the field as set.

Full dump-and-load round trip against this HEAD:

```
BEFORE round-trip, agent._cache_control = type='ephemeral' ttl='5m' cache_injection_point_index=-1 cache_system=True
cache_control lines in dumped YAML: ['cache_control: null']
AFTER round-trip, agent._cache_control = None
loaded llm model_fields_set has cache_control: True
```

Every agent loaded from YAML or from a serialized dict therefore runs with no breakpoint, and for Anthropic-family models that means nothing is cached at all — the default this PR introduces applies only to agents constructed in-process, and the difference is invisible at both dump and load time. Distinguishing the two cases needs something other than `model_fields_set` — an explicit sentinel default, or excluding `cache_control` from serialization when it is `None`.

---

vitalii-dynamiq (commented):
Reviewed 2828d76: 2 findings (1 high, 1 medium).

---

vitalii-dynamiq (commented):
Reviewed 0d15f4b — 0 findings (0 critical, 0 high, 0 medium, 0 low).

---

dynamiq/nodes/agents/base.py
```diff
@@ -445,6 +455,16 @@ def __init__(self, **kwargs):
             if self.file_store.agent_file_write_enabled:
                 self.tools.append(FileWriteTool(file_store=self.file_store_backend))
 
+        if self.memory_store_backend:
+            # Memory is independent of the workspace: same single tool whether the agent has a file
+            # store, a sandbox, or neither. Not serialized; rebuilt from `memory_store` on load.
+            memory_tool = MemoryStoreTool(
```
vitalii-dynamiq: ### The run's `user_id` never reaches the memory tool, so a store without a configured `user_id` fails every call

**Medium severity**

`MemoryStoreTool` is constructed once here with only `backend` and `write_enabled`, and nothing ever binds the run's `user_id` to it. `MemoryStoreToolInputSchema.user_id` is `is_accessible_to_agent=False` (`memory_store_tool.py:70`), so the model cannot supply it either, and `execute` passes `input_data.user_id` — which stays `None` — straight through to the backend.

The only place `_run_tool` injects `user_id` into a tool's input is `base.py:1789-1793`, gated on `is_child_agent`; a plain tool never gets it. The run-scoped alternative is the LTM pattern — `_build_long_term_memory_tools(input_data)` reads `input_data.user_id` and builds per-run tools bound to it (`base.py:1053`) — and the memory tool does not use it. Grepping `merged_input[...]` assignments in `base.py` confirms there is no other writer.

So for `DynamiqMemoryStore(connection=..., memory_store_id=..., description=...)` with no `user_id` on the store, `agent.run(input_data={"input": ..., "user_id": "u-42"})` makes `_scope` (`storages/memory/dynamiq.py:70`) raise on every list/read/write/edit/delete:

```
MemoryStoreError: No user_id for this memory store: pass one on the run or set one on the store.
```

relayed to the model as a recoverable `ToolExecutionException`, telling it to do the thing the caller already did. Memory is unusable for the whole run, and the agent burns loops retrying it.

This is the multi-tenant configuration the design describes: the field docstring states "A per-run user_id overrides it; one of the two must be set" (`dynamiq.py:33`), and `MemoryStore.list/read/write` document `user_id` as "the end user this call is for; it overrides whatever default the store was configured with". Only the store-configured path is exercised anywhere — `examples/components/agents/agents/agent_memory_store.py` sets `user_id=USER_ID` on the store, and `tests/unit/nodes/agents/test_memory_store_wiring.py` does the same — so the per-run override has no coverage. Today `tool_params` or a hand-written `input_transformer` are the only ways to deliver it; binding `input_data.user_id` into the tool's input the way LTM tools do closes it.

---

vitalii-dynamiq (commented):
Reviewed a589cfeb2c49ed840aa6cab512f42ec3bb4df016: 1 medium finding.

---

tyaroshko (dismissed):

dynamiq/nodes/agents/agent.py
```diff
@@ -2237,6 +2239,11 @@ def _react_prompt_config(self, tools: list | None = None) -> ReactPromptConfig:
             notes_file_path=self.get_notes_file_path(),
             todo_management_enabled=(self.file_store.enabled and self.file_store.todo_enabled)
             or bool(self.sandbox_backend),
+            memory_store_enabled=bool(self.memory_store_backend),
```
vitalii-dynamiq, line 2245: ### A memory-only agent is told it has no tools, so it can never call `memory-store`

**High severity**

`memory_store_enabled` is added here, but `has_tools` seven lines up (line 2235) is still `bool(tools) or skills or ltm_enabled` — the memory store is not OR-ed in. The memory tool is a per-run overlay (`_build_memory_store_tool` → `_run_extra_tools`), never a member of `self.tools`, so an agent whose only tool is the memory store evaluates `has_tools=False` at init.

`build_react_prompt` then sets `prompt_blocks["tools"] = ""` and picks the no-tools instruction variant (`manager.py:319-320`), and `REACT_BLOCK_INSTRUCTIONS_NO_TOOLS` / `REACT_BLOCK_XML_INSTRUCTIONS_NO_TOOLS` document only `Thought:`/`Answer:` — no action format at all. Meanwhile the `memory_store_enabled` branch below still appends the memory protocol, which opens with "Your `memory-store` tool is the ONLY thing that survives this conversation."

Reproduced against this HEAD with `Agent(llm=..., memory_store=MemoryStoreConfig(enabled=True, backend=DynamiqMemoryStore(...)))` and no other tools:

```
mode DEFAULT | agent.tools: [] | has_tools: False | memory_store_enabled: True
  tools block repr: ''
  instructions describe an action format: False
  ops says: Your `memory-store` tool is the ONLY thing that survives this conversation...
  runtime tool attached: ['memory-store']
mode XML     | identical
```

The tool is attached and reachable by name, but the model is given no syntax to emit a call and no tool listing, so in DEFAULT and XML inference modes it cannot use memory at all while being instructed that it must. This is the PR's own `agent_with_two_memories()` shape (no `tools=`), and `_handle_action` at line 1652 was already changed to `self._runtime_tools` for exactly this reason — "per-run overlays (memory store, LTM) may be the only tools an agent has" — but the prompt side still keys off `self.tools`. OR-ing `bool(self.memory_store_backend)` into `has_tools` the way `ltm_enabled` is matches the dispatch fix.

---

dynamiq/nodes/agents/base.py
```diff
@@ -2340,6 +2370,11 @@ def file_store_backend(self) -> FileStore | None:
         """Get the file store backend from the configuration if enabled."""
         return self.file_store.backend if self.file_store.enabled else None
 
+    @property
+    def memory_store_backend(self) -> MemoryStore | None:
```
vitalii-dynamiq, line 2358: ### A file upload permanently strips the `## Memory` block from the system prompt

**Medium severity**

`_setup_in_memory_file_store_and_tools` rebuilds the ReAct prompt through a second `ReactPromptConfig(...)` construction (base.py:2228) that was not updated alongside `_react_prompt_config`: it passes `inference_mode`, `todo_management_enabled`, `sandbox_base_path`, `role` and `instructions`, but none of `memory_store_enabled`, `memory_store_namespaces` or `memory_store_writable`. Those default to off, so `build_react_prompt` drops the memory protocol and the namespace listing entirely.

The rebuild is gated on `_prompt_blocks.get("tools") == ""`, which is exactly the memory-only agent — so the two conditions co-occur. Reproduced against this HEAD with `Agent(llm=..., memory_store=MemoryStoreConfig(enabled=True, backend=DynamiqMemoryStore(...)), instructions="Be terse.")`:

```
BEFORE upload: memory block present: True   | tools block: ''
AFTER  upload: memory block present: False  | tools block non-empty: True
  'memory-store' in full prompt: False
  instructions preserved: True
```

The path runs from `execute` (base.py:877) whenever files are attached and the agent has no sandbox or file store. `_build_memory_store_tool` still attaches `memory-store` on every run, so the model is handed a memory tool with no protocol, no namespace prefixes and no instruction to consult it first — it stops writing and reading memory, silently. The prompt manager is mutated in place and `file_store_backend` is now set, so the branch never re-runs and the loss persists for the lifetime of that agent instance. Passing the same `memory_store_*` values here (or reusing `_react_prompt_config`) keeps the two sites from drifting.

---

dynamiq/storages/memory/dynamiq.py
```diff
@@ -0,0 +1,206 @@
...
+        include_secure_params = kwargs.pop("include_secure_params", False)
+        exclude = kwargs.pop("exclude", self.to_dict_exclude_params)
+        data = self.model_dump(exclude=exclude, **kwargs)
+        data["type"] = self.type
+        data["connection"] = self.connection.to_dict(for_tracing=not include_secure_params)
+        return data
+
+    def identity(self):
+        """Which remote store this addresses: the same triple means the same data."""
+        return (self.type, self.connection.url, self.memory_store_id, self.user_id)
```
vitalii-dynamiq: ### `identity()` keys on a `user_id` the run overrides, so the duplicate-store guard misses a real collision

**Medium severity**

`user_id` here is only the store's *default* tenant — `_scope` (line 67) prefers the run's `user_id` over it. Including it in the identity tuple therefore makes two stores look distinct when they address the same remote data for the run that actually happens.

`CompositeMemoryStore.normalize_route_prefixes` uses this tuple as its duplicate-store key (`composite.py:78`), and its error states the invariant it is enforcing: "the prefix is stripped before the store is called, so both would write to the same keys." Two routes over one `memory_store_id` that differ only in their default `user_id` pass that check and then violate exactly that invariant. Reproduced against this HEAD:

```
identities differ: True
validator accepted the config: routes = ['user/', 'team/']
c.write("user/notes.md", "A", user_id="u-42") -> remote store ms-1, tenant u-42, key 'notes.md'
c.write("team/notes.md", "B", user_id="u-42") -> remote store ms-1, tenant u-42, key 'notes.md'
```

Both routes resolve to the same remote key under the same tenant, so `team/notes.md` overwrites `user/notes.md` and each subsequent read returns the other memory's content. Nothing surfaces it: `describe_namespaces` still advertises two separate memories to the model, and the write returns a `MemoryEntry` remounted under the route the caller asked for. The run-scoped `user_id` is the documented multi-tenant path ("A per-run user_id overrides it", line 33) and is now bound into the tool on every run (`base.py:1082`), so the overriding value is the normal case rather than the exception. Since a run can override it, `user_id` cannot identify the data a store addresses — dropping it from the tuple makes the guard reject this pair at construction.

---

vitalii-dynamiq (commented):
Reviewed b9a576a: 3 findings (1 high, 2 medium).

---

delamainer (dismissed):

maksymbuleshnyi (dismissed):

maksymbuleshnyi (approved):

PR #931, opened 2026-09-10
feat(audio): add provider-neutral SpeechToText and TextToSpeech nodes

vitalii-dynamiq:
## Summary

- Adds `dynamiq.nodes.audio.SpeechToText` and `dynamiq.nodes.audio.TextToSpeech`: one node per capability, with the provider chosen by the connection type (the same approach `ImageGeneration` already uses).
- Speech to text providers: OpenAI (`gpt-4o-transcribe`, `gpt-4o-transcribe-diarize`, `whisper-1`), Groq, Mistral Voxtral, Deepgram (Nova-3), ElevenLabs Scribe, and any OpenAI-compatible endpoint through the `HttpApiKey` / `Whisper` connections.
- Text to speech providers: OpenAI, Groq (Orpheus), ElevenLabs, Mistral Voxtral TTS, Deepgram Aura-2.
- Normalized `Transcript` output (`content`, `segments`, `words`, `speakers`, `language`, `duration`, `usage`, optional `raw`), so diarized output has the same shape regardless of the provider.
- Options (`diarize`, `timestamps`, `speakers`, `prompt`, `language`, `speed`, `instructions`, `output_format`, `sample_rate`) are checked against per-provider capability flags when the node is built, so an unsupported combination fails at workflow construction instead of mid-run.
- `SpeechToText` takes part in agent file injection: the `audio` field is filled from the agent's file store and hidden from the LLM, the same way the image nodes do it.
- New `Deepgram` connection (`DEEPGRAM_API_KEY`).
- Existing `WhisperSTT`, `ElevenLabsTTS` and `ElevenLabsSTS` nodes are untouched and stay registered.

## Design notes

- Adapters live under `dynamiq/components/audio/{stt,tts}` with a registry keyed by connection class. Adding a provider is one adapter file plus a registry entry; the node surface does not change.
- OpenAI-compatible providers go through the official `openai` SDK (already a dependency, supports `diarized_json`). Mistral, Deepgram and ElevenLabs are called with `requests` directly. LiteLLM is intentionally not used for audio because it drops diarization data.
- `provider_options` passes vendor-specific parameters through untouched (for example Deepgram's `diarize_model` or ElevenLabs voice settings).

## Related

- #888 (MiniMax TTS) also edits `dynamiq/nodes/audio/__init__.py`; whichever lands second needs a one-line rebase. MiniMax can later be exposed as a `TextToSpeech` adapter instead of a separate node.
- Companion changes: dynamiq-ai/mono#1363 (node registration and validation) and dynamiq-ai/ui#1574 (builder panels). Rollout order: release this package, bump the executor image, then deploy mono and ui.

## Testing

- `pytest tests/unit/components/audio tests/integration/nodes/audio` (54 tests, HTTP mocked per provider: diarization, word/segment timestamps, audio URLs, agent file injection, capability errors, YAML round trips)
- full `tests/unit` suite
- pre-commit hooks (black, isort, flake8, darker)

### DIFF
--- a/dynamiq/components/audio/registry.py
+++ b/dynamiq/components/audio/registry.py
@@ -1,0 +1,73 @@
+from typing import Any
+
+from dynamiq.components.audio.stt import (
+    BaseSTTAdapter,
+    DeepgramSTTAdapter,
+    ElevenLabsSTTAdapter,
+    GroqSTTAdapter,
+    MistralSTTAdapter,
+    OpenAICompatibleSTTAdapter,
+    OpenAISTTAdapter,
+)
+from dynamiq.components.audio.tts import (
+    BaseTTSAdapter,
+    DeepgramTTSAdapter,
+    ElevenLabsTTSAdapter,
+    GroqTTSAdapter,
+    MistralTTSAdapter,
+    OpenAICompatibleTTSAdapter,
+    OpenAITTSAdapter,
+)
+from dynamiq.connections import BaseConnection
+from dynamiq.connections import Deepgram as DeepgramConnection
+from dynamiq.connections import ElevenLabs as ElevenLabsConnection
+from dynamiq.connections import Groq as GroqConnection
+from dynamiq.connections import HttpApiKey as HttpApiKeyConnection
+from dynamiq.connections import Mistral as MistralConnection
+from dynamiq.connections import OpenAI as OpenAIConnection
+from dynamiq.connections import Whisper as WhisperConnection
+
+# The provider is the connection type: adding a provider means one adapter and one entry here.
+STT_ADAPTERS: dict[type[BaseConnection], type[BaseSTTAdapter]] = {
+    OpenAIConnection: OpenAISTTAdapter,
+    GroqConnection: GroqSTTAdapter,
+    HttpApiKeyConnection: OpenAICompatibleSTTAdapter,
+    WhisperConnection: OpenAICompatibleSTTAdapter,
+    MistralConnection: MistralSTTAdapter,
+    DeepgramConnection: DeepgramSTTAdapter,
+    ElevenLabsConnection: ElevenLabsSTTAdapter,
+}
+
+TTS_ADAPTERS: dict[type[BaseConnection], type[BaseTTSAdapter]] = {
+    OpenAIConnection: OpenAITTSAdapter,
+    GroqConnection: GroqTTSAdapter,
+    HttpApiKeyConnection: OpenAICompatibleTTSAdapter,
+    WhisperConnection: OpenAICompatibleTTSAdapter,
+    ElevenLabsConnection: ElevenLabsTTSAdapter,
+    MistralConnection: MistralTTSAdapter,
+    DeepgramConnection: DeepgramTTSAdapter,
+}
+
+
+def _resolve(registry: dict[type[BaseConnection], Any], connection: BaseConnection | None, client: Any, kind: str):
+    if connection is None:
+        # A bare OpenAI SDK client is the only client-only configuration the audio nodes accept.
+        if client is not None and hasattr(client, "audio"):
+            return registry[OpenAIConnection]
+        raise ValueError(f"A connection is required to pick the {kind} provider.")
+    # Walk the MRO so subclasses of a registered connection (MistralOCR, Dynamiq) map to the same adapter.
+    for base in type(connection).__mro__:
+        if base in registry:
+            return registry[base]
+    supported = ", ".join(sorted(item.__name__ for item in registry))
+    raise ValueError(
+        f"Connection {type(connection).__name__} is not supported for {kind}. Supported connections: {supported}."
+    )
+
+
+def resolve_stt_adapter(connection: BaseConnection | None, client: Any = None) -> type[BaseSTTAdapter]:
+    return _resolve(STT_ADAPTERS, connection, client, "speech-to-text")
+
+
+def resolve_tts_adapter(connection: BaseConnection | None, client: Any = None) -> type[BaseTTSAdapter]:
+    return _resolve(TTS_ADAPTERS, connection, client, "text-to-speech")
--- a/dynamiq/components/audio/stt/deepgram.py
+++ b/dynamiq/components/audio/stt/deepgram.py
@@ -1,0 +1,90 @@
+from typing import Any
+
+from dynamiq.components.audio.stt.base import BaseSTTAdapter, STTCapabilities, TranscriptionRequest
+from dynamiq.components.audio.stt.normalize import build_transcript, optional_str
+from dynamiq.components.audio.utils import format_param_value, raise_for_status, resolve_http_client, split_terms
+from dynamiq.types.audio import TimestampGranularity, Transcript, TranscriptSegment, TranscriptWord
+
+
+class DeepgramSTTAdapter(BaseSTTAdapter):
+    """Deepgram pre-recorded transcription (``/v1/listen``) with word-level diarization."""
+
+    provider = "Deepgram"
+    capabilities = STTCapabilities(diarization=True, word_timestamps=True, prompt=True, audio_url_input=True)
+    default_model = "nova-3"
+
+    def build_params(self, request: TranscriptionRequest) -> dict[str, Any]:
+        params: dict[str, Any] = {"model": request.model, "smart_format": True}
+        if request.language:
+            params["language"] = request.language
+        if request.diarize:
+            params["diarize"] = True
+        if request.diarize or request.timestamps != TimestampGranularity.NONE:
+            # Utterances are the only timed segments Deepgram returns; without them word timing
+            # would collapse the transcript into a single synthetic segment.
+            params["utterances"] = True
+        if request.prompt:
+            params["keyterm"] = split_terms(request.prompt)
+        params.update(request.provider_options)
+        # diarize_model is Deepgram's newer switch; sending the legacy flag alongside it would pin v1.
+        if "diarize_model" in params:
+            params.pop("diarize", None)
+        return {key: format_param_value(value) for key, value in params.items() if value is not None}
+
+    def transcribe(self, request: TranscriptionRequest) -> Transcript:
+        http = resolve_http_client(self.client)
+        url = f"{self.connection.url.rstrip('/')}/listen"
+        headers = {"Authorization": f"Token {self.connection.api_key}"}
+        params = self.build_params(request)
+        if request.audio_url:
+            headers["Content-Type"] = "application/json"
+            response = http.post(url, params=params, headers=headers, json={"url": request.audio_url})
+        else:
+            headers["Content-Type"] = getattr(request.audio, "content_type", None) or "application/octet-stream"
+            response = http.post(url, params=params, headers=headers, data=request.audio.getvalue())
+        raise_for_status(response, self.provider)
+        return self.normalize(response.json(), request)
+
+    @staticmethod
+    def normalize(data: dict[str, Any], request: TranscriptionRequest) -> Transcript:
+        results = data.get("results") or {}
+        channels = results.get("channels") or []
+        channel = channels[0] if channels else {}
+        alternatives = channel.get("alternatives") or []
+        alternative = alternatives[0] if alternatives else {}
+
+        words = [
+            TranscriptWord(
+                word=word.get("punctuated_word") or word.get("word") or "",
+                start=word.get("start"),
+                end=word.get("end"),
+                speaker=optional_str(word.get("speaker")),
+                confidence=word.get("confidence"),
+            )
+            for word in alternative.get("words") or []
+        ]
+        segments = [
+            TranscriptSegment(
+                id=str(utterance.get("id") or index),
+                text=utterance.get("transcript", ""),
+                start=utterance.get("start"),
+                end=utterance.get("end"),
+                speaker=optional_str(utterance.get("speaker")),
+                confidence=utterance.get("confidence"),
+            )
+            for index, utterance in enumerate(results.get("utterances") or [])
+        ]
+        metadata = data.get("metadata") or {}
+        duration = metadata.get("duration")
+        detected = channel.get("detected_language")
+        requested = request.language if request.language and request.language != "multi" else None
+        return build_transcript(
+            content=alternative.get("transcript") or "",
+            segments=segments,
+            words=words,
+            language=detected or requested,
+            languages=list(channel.get("languages") or []),
+            duration=duration,
+            usage={"seconds": duration} if duration is not None else {},
+            raw=data,
+        )
--- a/dynamiq/components/audio/stt/mistral.py
+++ b/dynamiq/components/audio/stt/mistral.py
@@ -1,0 +1,97 @@
+from typing import Any
+
+from dynamiq.components.audio.stt.base import BaseSTTAdapter, STTCapabilities, TranscriptionRequest
+from dynamiq.components.audio.stt.normalize import build_transcript, optional_str
+from dynamiq.components.audio.utils import multipart_fields, raise_for_status, resolve_http_client, split_terms
+from dynamiq.types.audio import SpeakerHints, TimestampGranularity, Transcript, TranscriptSegment, TranscriptWord
+
+MISTRAL_API_BASE_URL = "https://api.mistral.ai/v1"
+
+
+class MistralSTTAdapter(BaseSTTAdapter):
+    """Mistral Voxtral transcription (``/v1/audio/transcriptions``) with speaker diarization."""
+
+    provider = "Mistral"
+    capabilities = STTCapabilities(diarization=True, word_timestamps=True, prompt=True, audio_url_input=True)
+    default_model = "voxtral-mini-latest"
+
+    @classmethod
+    def check_config(
+        cls,
+        *,
+        diarize: bool,
+        timestamps: TimestampGranularity,
+        speakers: SpeakerHints | None,
+        prompt: str | None,
+        language: str | None = None,
+    ) -> None:
+        super().check_config(
+            diarize=diarize, timestamps=timestamps, speakers=speakers, prompt=prompt, language=language
+        )
+        if diarize and timestamps == TimestampGranularity.WORD:
+            raise ValueError("Mistral diarization returns segment timings only; use timestamps='segment' with diarize.")
+
+    def build_fields(self, request: TranscriptionRequest) -> dict[str, Any]:
+        fields: dict[str, Any] = {"model": request.model}
+        if request.language:
+            fields["language"] = request.language
+        if request.diarize:
+            fields["diarize"] = True
+            # Diarization is rejected outright unless segment granularity comes with it, whatever the
+            # node asked for: the timed segments are what carries the speaker labels.
+            fields["timestamp_granularities"] = ["segment"]
+        elif request.timestamps != TimestampGranularity.NONE:
+            fields["timestamp_granularities"] = [request.timestamps.value]
+        if request.prompt:
+            fields["context_bias"] = split_terms(request.prompt)
+        if request.audio_url:
+            fields["file_url"] = request.audio_url
+        fields.update(request.provider_options)
+        return fields
+
+    def transcribe(self, request: TranscriptionRequest) -> Transcript:
+        http = resolve_http_client(self.client)
+        files: list[tuple[str, Any]] = multipart_fields(self.build_fields(request))
+        if request.audio is not None:
+            audio = request.audio
+            files.append(("file", (audio.name, audio, getattr(audio, "content_type", None))))
+        response = http.post(
+            f"{MISTRAL_API_BASE_URL}/audio/transcriptions",
+            headers={"Authorization": f"Bearer {self.connection.api_key}"},
+            files=files,
+        )
+        raise_for_status(response, self.provider)
+        return self.normalize(response.json(), request)
+
+    @staticmethod
+    def normalize(data: dict[str, Any], request: TranscriptionRequest) -> Transcript:
+        segments = [
+            TranscriptSegment(
+                id=str(index),
+                text=segment.get("text", ""),
+                start=segment.get("start"),
+                end=segment.get("end"),
+                speaker=optional_str(segment.get("speaker_id")),
+                confidence=segment.get("score"),
+            )
+            for index, segment in enumerate(data.get("segments") or [])
+        ]
+        words = [
+            TranscriptWord(
+                word=word.get("word") or word.get("text") or "",
+                start=word.get("start"),
+                end=word.get("end"),
+                speaker=optional_str(word.get("speaker_id")),
+            )
+            for word in data.get("words") or []
+        ]
+        usage = data.get("usage") or {}
+        return build_transcript(
+            content=data.get("text") or "",
+            segments=segments,
+            words=words,
+            language=data.get("language") or request.language,
+            duration=usage.get("prompt_audio_seconds"),
+            usage=usage,
+            raw=data,
+        )
--- a/dynamiq/components/audio/stt/openai.py
+++ b/dynamiq/components/audio/stt/openai.py
@@ -1,0 +1,108 @@
+from typing import Any
+
+from dynamiq.components.audio.openai_client import resolve_openai_client
+from dynamiq.components.audio.stt.base import BaseSTTAdapter, STTCapabilities, TranscriptionRequest
+from dynamiq.components.audio.stt.normalize import build_transcript, optional_str
+from dynamiq.types.audio import TimestampGranularity, Transcript, TranscriptSegment, TranscriptWord
+from dynamiq.utils.logger import logger
+
+# The gpt-4o-transcribe family only returns plain JSON; verbose_json (segments, words) is a Whisper feature.
+JSON_ONLY_MODEL_MARKER = "transcribe"
+
+
+def normalize_openai_response(data: dict[str, Any], request: TranscriptionRequest) -> Transcript:
+    """Map ``json``, ``verbose_json`` and ``diarized_json`` responses to a ``Transcript``."""
+    segments = [
+        TranscriptSegment(
+            id=str(segment.get("id", index)),
+            text=segment.get("text", ""),
+            start=segment.get("start"),
+            end=segment.get("end"),
+            speaker=optional_str(segment.get("speaker")),
+        )
+        for index, segment in enumerate(data.get("segments") or [])
+    ]
+    words = [
+        TranscriptWord(word=word.get("word", ""), start=word.get("start"), end=word.get("end"))
+        for word in data.get("words") or []
+    ]
+    languages = [item["code"] for item in data.get("languages") or [] if isinstance(item, dict) and item.get("code")]
+    usage = data.get("usage") or {}
+    duration = data.get("duration")
+    if duration is None and usage.get("type") == "duration":
+        duration = usage.get("seconds")
+    return build_transcript(
+        content=data.get("text") or "",
+        segments=segments,
+        words=words,
+        language=data.get("language") or request.language,
+        languages=languages,
+        duration=duration,
+        usage=usage,
+        raw=data,
+    )
+
+
+class OpenAICompatibleSTTAdapter(BaseSTTAdapter):
+    """Any server that speaks the OpenAI ``/audio/transcriptions`` contract: Groq, vLLM, Speaches, NIM."""
+
+    provider = "OpenAI-compatible endpoint"
+    capabilities = STTCapabilities(word_timestamps=True, prompt=True)
+    default_model = "whisper-1"
+
+    def build_params(self, request: TranscriptionRequest) -> dict[str, Any]:
+        audio = request.audio
+        params: dict[str, Any] = {
+            "model": request.model,
+            "file": (audio.name, audio, getattr(audio, "content_type", None)),
+        }
+        if request.language:
+            params["language"] = request.language
+        if request.prompt:
+            params["prompt"] = request.prompt
+        if request.diarize:
+            params["response_format"] = "diarized_json"
+            params["chunking_strategy"] = "auto"
+        elif request.timestamps != TimestampGranularity.NONE and JSON_ONLY_MODEL_MARKER not in request.model:
+            params["response_format"] = "verbose_json"
+            params["timestamp_granularities"] = (
+                ["word", "segment"] if request.timestamps == TimestampGranularity.WORD else ["segment"]
+            )
+        else:
+            params["response_format"] = "json"
+        params.update(request.provider_options)
+        return params
+
+    def transcribe(self, request: TranscriptionRequest) -> Transcript:
+        client = resolve_openai_client(self.connection, self.client)
+        response = client.audio.transcriptions.create(**self.build_params(request))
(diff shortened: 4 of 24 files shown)


dynamiq/components/audio/stt/mistral.py
```diff
@@ -0,0 +1,82 @@
...
+    default_model = "voxtral-mini-latest"
+
+    def build_fields(self, request: TranscriptionRequest) -> dict[str, Any]:
+        fields: dict[str, Any] = {"model": request.model}
+        if request.language:
+            fields["language"] = request.language
+        if request.diarize:
+            fields["diarize"] = True
+        if request.timestamps == TimestampGranularity.WORD:
+            fields["timestamp_granularities"] = ["word"]
```
vitalii-dynamiq: ### Word timestamps combined with an explicit language send the field Mistral rejects

**Medium severity**

The `SEGMENT` branch below guards on `not request.language` because, as its own comment states, Mistral rejects `timestamp_granularities` combined with an explicit language. The `WORD` branch here applies no such guard, so `SpeechToText(connection=Mistral(...), timestamps="word", language="fr")` posts `language=fr` and `timestamp_granularities=word` together — exactly the combination the adjacent comment says is invalid — and `raise_for_status` turns the provider's rejection into an `HTTPError` at run time.

Both node options are individually valid and pass `check_config` (`word_timestamps=True` for this provider), so nothing catches this at construction. Neither Mistral test covers it: `test_mistral_word_timestamps_and_file_url` sets no language, and `test_mistral_diarized_transcription` sets `language="fr"` but with `diarize=True`, which skips the `elif` entirely.

---

dynamiq/components/audio/registry.py
```diff
@@ -0,0 +1,72 @@
...
+from dynamiq.connections import Groq as GroqConnection
+from dynamiq.connections import HttpApiKey as HttpApiKeyConnection
+from dynamiq.connections import Mistral as MistralConnection
+from dynamiq.connections import OpenAI as OpenAIConnection
+from dynamiq.connections import Whisper as WhisperConnection
+
+# The provider is the connection type: adding a provider means one adapter and one entry here.
+STT_ADAPTERS: dict[type[BaseConnection], type[BaseSTTAdapter]] = {
+    OpenAIConnection: OpenAISTTAdapter,
+    GroqConnection: OpenAICompatibleSTTAdapter,
```
vitalii-dynamiq: ### Groq speech-to-text defaults to a model Groq does not serve

**Medium severity**

Mapping `Groq` to `OpenAICompatibleSTTAdapter` also inherits its `default_model = "whisper-1"`, and `validate_provider_config` assigns that whenever `model` is left unset:

```python
if self.model is None:
    self.model = adapter_cls.default_model
```

`whisper-1` is an OpenAI-only model id; Groq serves `whisper-large-v3`, `whisper-large-v3-turbo` and `distil-whisper-large-v3-en`. So `SpeechToText(connection=Groq(api_key=...))` — the documented zero-config path for a provider this PR advertises — builds successfully and then fails on the first request with `model_not_found`.

`whisper-1` remains the right default for the `HttpApiKey` and `Whisper` entries, so this needs a Groq-specific adapter subclass overriding `default_model`. Worth noting that `test_default_models_follow_the_provider` asserts the default for every other provider but omits Groq, and `test_groq_uses_the_openai_contract` passes `model="whisper-large-v3-turbo"` explicitly, so the broken path is the one path left untested.

---

dynamiq/nodes/audio/speech_to_text.py
```diff
@@ -0,0 +1,174 @@
...
+from dynamiq.types.cancellation import check_cancellation
+
+DEFAULT_FILE_NAME = "audio.wav"
+DEFAULT_CONTENT_TYPE = "audio/wav"
+
+
+class SpeechToTextInputSchema(BaseModel):
+    audio: io.BytesIO | bytes | None = Field(
+        default=None, description="Audio file to transcribe, as bytes or a file object."
+    )
```
vitalii-dynamiq, line 34: ### `is_files_allowed = True` is inert because the audio field is not marked `map_from_storage`

**Medium severity**

The node sets `is_files_allowed = True` (line 108), but agent file injection only fills fields that opt in through `json_schema_extra`. In `Agent._inject_files_into_tool`:

```python
for field_name, field in tool.input_schema.model_fields.items():
    if not (field.json_schema_extra and field.json_schema_extra.get("map_from_storage", False)):
        continue
```

`audio` carries no `json_schema_extra`, so the loop skips it and `merged_input` is returned untouched. An agent using `SpeechToText` as a tool with an uploaded audio file therefore never receives it, and the node fails its own `validate_source` with "Either `audio` or `audio_url` must be provided."

The same omission leaks the field into the tool schema shown to the LLM: `_is_accessible_to_agent` defaults to `True` when `json_schema_extra` is absent, so the raw `io.BytesIO | bytes` parameter is advertised as something the model should fill in.

The two comparable nodes both declare both flags — `nodes/images/variation.py:25` and `nodes/images/edit.py:74`:

```python
json_schema_extra={"map_from_storage": True, "is_accessible_to_agent": False},
```

---

dynamiq/components/audio/stt/deepgram.py
```diff
@@ -0,0 +1,88 @@
...
+    capabilities = STTCapabilities(diarization=True, word_timestamps=True, prompt=True, audio_url_input=True)
+    default_model = "nova-3"
+
+    def build_params(self, request: TranscriptionRequest) -> dict[str, Any]:
+        params: dict[str, Any] = {"model": request.model, "smart_format": True}
+        if request.language:
+            params["language"] = request.language
+        if request.diarize:
+            params["diarize"] = True
+        if request.diarize or request.timestamps == TimestampGranularity.SEGMENT:
```
vitalii-dynamiq: ### Requesting word timestamps on Deepgram collapses segments into one

**Medium severity**

`utterances` is requested only for `diarize` or `timestamps == SEGMENT`, so `timestamps == WORD` omits it. Deepgram then returns no `results.utterances`, `normalize` builds an empty `segments` list, and `build_transcript` falls through to:

```python
if not segments and words:
    segments = segments_from_words(words)
```

Without diarization every word has `speaker=None`, so `segments_from_words` compares `groups[-1][0].speaker == word.speaker` as `None == None` for every word and accumulates all of them into a single group. The result is one synthetic segment spanning the entire recording.

The consequence is inverted granularity: `timestamps="segment"` yields properly split utterances, while `timestamps="word"` — the strictly finer setting — yields exactly one segment covering everything. The `OpenAICompatibleSTTAdapter` treats `WORD` as a superset for this reason, requesting `["word", "segment"]`, so the intended semantics are clear. Including `WORD` in the `utterances` condition keeps segments while still returning word timings.

---

vitalii-dynamiq (commented):
Reviewed b8e2597: 4 findings (4 medium).

---

vitalii-dynamiq:
Verified the nodes end to end against live provider APIs (OpenAI, Mistral, Deepgram, ElevenLabs): 14/14 checks pass. Speech was synthesised with each TTS provider, a two-speaker clip was built from two OpenAI voices, and every STT provider transcribed it back.

Two real defects surfaced that no mocked test could catch, both now fixed in this PR:

**Mistral TTS was unusable.** Voxtral has no implicit voice, so every request without one is rejected with `400 Either ref_audio or voice must be provided.` The adapter never sent a voice, so the provider failed on its default path. It now falls back to the global preset `en_paul_neutral`, unless a voice, a `voice_id`, or a `ref_audio` provider option is given. The default model also moves to `voxtral-mini-tts-latest`, matching how the STT side pins `voxtral-mini-latest`.

**Mistral diarization was rejected.** The live API requires segment granularity alongside `diarize`: `When diarize is set to True and streaming is disabled, the timestamp granularity must be set to ['segment'], got []`. The adapter deliberately omitted it, so every diarized request failed with 422. It now always sends `timestamp_granularities=["segment"]` with `diarize`, and word timestamps combined with diarization are rejected at construction, since Voxtral cannot return both.

The same run also corrected an assumption taken from Mistral's docs. The documented incompatibility between `timestamp_granularities` and `language` no longer holds: segment and word granularity both succeed with an explicit language, and with diarization. The construction-time guard that rejected that combination is gone.

Two other observations, neither a code change:

- Deepgram returns a single speaker on every clip tried here, including its own `interview_speech-analytics.wav` sample, across nova-3, nova-2 and `diarize_model=v2`. The adapter reports faithfully what the API returns; the speaker labels are attached to words and segments as expected. OpenAI, Mistral and ElevenLabs all resolved two speakers on the same audio.
- Groq is untested against a live key; it goes through the same OpenAI-compatible path as the OpenAI provider, which is covered.

dynamiq/components/audio/utils.py
```diff
@@ -0,0 +1,118 @@
...
+    detail = (response.text or "").strip()[:500] or response.reason
+    raise requests.HTTPError(
+        f"{provider} request failed with status {response.status_code}: {detail}", response=response
+    )
+
+
+def _looks_like_audio(file: Any) -> bool | None:
+    """Whether a file is a recording. ``None`` when it carries nothing to judge by."""
+    content_type = getattr(file, "content_type", None)
+    if not content_type:
```
vitalii-dynamiq: ### Agent-uploaded audio is rejected because the store stamps `application/octet-stream`

**High severity**

The filename is only consulted when `content_type` is falsy. Files that reach the store through the agent's upload path always carry a `content_type`, and for the ordinary case it is `application/octet-stream`:

`Agent._ensure_named_files` (`nodes/agents/base.py:1878`) sets `name` and `description` but never `content_type`, so a user-supplied `io.BytesIO` with `name = "meeting.mp3"` arrives without one. `_upload_files_to_file_store` then stores it as `content_type=getattr(file_obj, "content_type", "application/octet-stream")` (`base.py:2165`). Because `store()` only falls back to `mimetypes.guess_type` when `content_type is None`, the explicit `application/octet-stream` wins and the `.mp3` extension is never examined. `list_files_bytes` hands that same value back.

`_looks_like_audio` therefore returns `False` rather than `None`, the file is excluded from `unidentified`, and `select_audio_file` raises:

```
ValueError: Received 1 file(s) but none of them look like audio: meeting.mp3.
```

Confirmed by running the store path directly. This is the primary agent transcription flow — an agent handed a single recording as `BytesIO` can never transcribe it. Falling back to the filename whenever the content type is absent *or* `application/octet-stream` fixes it.

---

dynamiq/nodes/audio/speech_to_text.py
```diff
@@ -0,0 +1,199 @@
...
+    audio_url: str | None = Field(
+        default=None, description="Public URL of the audio, for providers that download it themselves."
+    )
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    @model_validator(mode="after")
+    def validate_source(self):
+        # Agent file injection hands over every stored file; the node transcribes one recording.
+        self.audio = select_audio_file(self.audio)
```
vitalii-dynamiq, line 52: ### A URL-only transcription fails when the agent's store holds no audio

**Medium severity**

`select_audio_file` runs before `audio_url` is considered, and it raises when every injected file is provably not audio. `execute` takes the opposite position — "An explicit URL wins over the file" (line 183) — so the file is irrelevant on this path, but validation has already failed.

`Agent._inject_files_into_tool` fills `audio` with the store's entire contents whenever the LLM leaves the field unset (`nodes/agents/base.py:1618`). So an agent whose store holds only, say, a PDF, calling `SpeechToText` with a Deepgram or Mistral connection and an `audio_url`, gets:

```
ValidationError: Value error, Received 1 file(s) but none of them look like audio: contract.pdf.
```

Verified against `SpeechToTextInputSchema(audio=[pdf], audio_url="https://example.com/call.mp3")`. Skipping file selection when `audio_url` is set matches what `execute` already does.

---

dynamiq/nodes/agents/components/schema_generator.py
```diff
@@ -143,10 +143,12 @@ def generate_input_formats(tools: list[Node], sanitize_tool_name: Callable[[str]
         for name, field in _reorder_fields(tool.resolved_input_schema.model_fields):
             if _is_accessible_to_agent(field):
                 args = get_args(field.annotation)
-                if get_origin(field.annotation) in (Union, types.UnionType):
-                    type_str = str(field.annotation)
-                elif field.json_schema_extra and field.json_schema_extra.get("map_from_storage", False):
+                # Checked before the union branch: a stored-file field accepts whatever the tool
+                # needs at execution (bytes, BytesIO, a mapping), but the LLM only ever names files.
+                if field.json_schema_extra and field.json_schema_extra.get("map_from_storage", False):
```
vitalii-dynamiq, line 155: ### Reordering the branches breaks Stagehand file uploads

**Medium severity**

Moving the `map_from_storage` check ahead of the union branch changes the advertised type of every union-typed stored-file field, not just the new audio one. `StagehandInputSchema.files` is `str | io.BytesIO | None` (`nodes/tools/stagehand.py:228`) and is accessible to the agent, so it previously rendered as the union and now renders as `tuple[str, ...]`.

The LLM follows that and emits a list. `_inject_files_into_tool` maps each name to a file (`base.py:1623-1624`), producing `files=[BytesIO]`, and `files_validator` passes lists through untouched — so the value hits the `str | io.BytesIO | None` annotation and fails:

```
ValidationError: 2 validation errors for StagehandInputSchema
files.str / files.is-instance[BytesIO]
```

Confirmed by constructing the schema both ways: a single `BytesIO` validates, a one-element list does not. The upload action becomes unusable whenever the model takes the schema at its word.

The other `map_from_storage` fields are unaffected — `cua_desktop`, `e2b_desktop`, `code_interpreter` and `python_code_executor` all accept list values.

---

dynamiq/nodes/audio/speech_to_text.py
```diff
@@ -0,0 +1,199 @@
...
+            language=self.language,
+            diarize=self.diarize,
+            timestamps=self.timestamps,
+            speakers=self.speakers,
+            prompt=self.prompt,
+            provider_options=dict(self.provider_options),
+        )
+        self._adapter.check_request(request)
+        transcript = self._adapter.transcribe(request)
+        return transcript.model_dump(exclude=None if self.include_raw_response else {"raw"})
```
vitalii-dynamiq: ### Diarized output never reaches the model when the node runs as an agent tool

**Medium severity**

The returned dict carries both `content` (plain text) and `transcript` (speaker-labelled), but `process_tool_output_for_agent` collapses any dict containing a `content` key to that key alone (`nodes/agents/utils.py:1094-1097`) and discards the rest of the dict. That is the function every agent observation goes through.

So an agent running `SpeechToText(diarize=True)` pays for diarization and then sees only the unlabelled transcript: `transcript`, `segments`, `speakers` and `words` are all dropped before the observation is built. Asked "who said what", the model has nothing to work from. The labelled text survives only in the streamed `tool_output_meta`, which the reasoning loop does not read.

Returning the speaker-labelled text under `content` when `diarize` is set keeps the plain-text contract for every other consumer while making the feature usable from an agent.

---

vitalii-dynamiq (commented):
Reviewed 7cf3d66: 1 high, 3 medium.

---

dynamiq/components/audio/utils.py
tyaroshko: Confirmed, and this one was the real breakage — thanks. Reproduced exactly as described: storing a `BytesIO` named `meeting.mp3` through the agent's own path gives it `content_type=application/octet-stream`, `_looks_like_audio` reads that as "provably not audio" rather than "no information", and the file is excluded before the name is ever examined.

Fixed by treating a generic content type the same as a missing one: `application/octet-stream` and `binary/octet-stream` now fall through to `mimetypes.guess_type(name)`. A file that is genuinely something else (`application/pdf`, `text/plain`) is still excluded, and a generic type with no usable name is still kept as a last resort.

Worth recording why the live agent test did not catch this: it was storing its fixture with an explicit `content_type="audio/wav"`, which is not what the upload path does. That helper now stamps `application/octet-stream` like `_upload_files_to_file_store` does, so the case cannot hide again.

dynamiq/nodes/audio/speech_to_text.py
tyaroshko: Confirmed. `SpeechToTextInputSchema(audio=[pdf], audio_url=...)` did raise, and the validator was contradicting `execute`, which prefers the URL and never looks at the files.

Fixed by returning early from `validate_source` when `audio_url` is set, so file selection is skipped entirely on that path — matching what `execute` already does. A run with no URL is unchanged.

dynamiq/nodes/agents/components/schema_generator.py
tyaroshko: Confirmed — the reorder was too broad. `StagehandInputSchema.files` is `str | io.BytesIO | None`, so it started advertising `tuple[str, ...]`, and a list from the model does not validate against that annotation.

The intent of the reorder stands (a stored-file field should be described as file names, not as `_io.BytesIO`), so the fix keeps it but makes the arity match the annotation: `tuple[str, ...]` only when the field can actually hold a collection, otherwise `str`. Every `map_from_storage` field now renders as:

```
stagehand              files (str)
python_code_executor   files (tuple[str, ...])
http_api_call          files (tuple[str, ...])
speech_to_text         audio (tuple[str, ...])
```

Stagehand is back to a single value — and better off than before the reorder, where it was showing the raw `str | _io.BytesIO | None` union.

dynamiq/nodes/audio/speech_to_text.py
tyaroshko: Confirmed — `process_tool_output_for_agent` collapses the dict to `content`, so an agent was paying for diarization and then seeing unlabelled text.

Taking the suggestion, with one adjustment: putting the labelled text in `content` unconditionally would change what every existing workflow reads from that field, and would leave no plain-text output at all. Instead it is gated on `is_optimized_for_agents`, which `Agent.init_components` already sets to `True` on every tool it owns — so the node returns the speaker-labelled transcript as `content` when it runs as an agent tool and diarization was requested, and the plain transcript everywhere else. `transcript`, `segments`, `speakers` and `words` are unchanged in both cases.

Verified live against `gpt-4o-transcribe-diarize`: the observation now starts with `Speaker A:` / `Speaker B:`, and the agent answers "who said what" correctly:

> Anna from the Zurich office says the invoice is overdue. The approval of the payment is mentioned by Speaker B...

dynamiq/components/audio/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    raise requests.HTTPError(
+        f"{provider} request failed with status {response.status_code}: {detail}", response=response
+    )
+
+
+def _looks_like_audio(file: Any) -> bool | None:
+    """Whether a file is a recording. ``None`` when it carries nothing to judge by."""
+    content_type = getattr(file, "content_type", None)
+    if not content_type or content_type in GENERIC_CONTENT_TYPES:
+        name = getattr(file, "name", None)
```
vitalii-dynamiq, line 91: ### Audio handed to an agent as raw `bytes` is rejected as "not audio"

**Medium severity**

The guessed content type is not put through the `GENERIC_CONTENT_TYPES` check that the attribute-supplied one gets two lines above, and `mimetypes.guess_type("file_0.bin")` returns `application/octet-stream`. So the generic type comes back as a positive identification, and `_looks_like_audio` returns `False` — "provably something else" — rather than the `None` that means "nothing to judge by".

`file_0.bin` is precisely the name the agent's own upload path produces. `Agent._ensure_named_files` (`nodes/agents/base.py:1884`) renames every raw-`bytes` entry of `AgentInputSchema.files` to `file_{i}.bin`, and `_upload_files_to_file_store` then stores it with `content_type="application/octet-stream"` (`base.py:2165`), which `list_files_bytes` hands back unchanged.

With the verdict `False` the file never reaches `unidentified`, and `select_audio_file` raises:

```
ValueError: Received 1 file(s) but none of them look like audio: file_0.bin.
```

which surfaces as a `SpeechToTextInputSchema` validation error. An agent given a recording as `bytes` — `Agent(...).run(input_data={"files": [Path("call.mp3").read_bytes()]})` — can therefore never transcribe it, because the file is discarded before the node runs.

The last-resort branch is tested only with an extension-less name: `test_a_generic_content_type_with_no_usable_name_is_a_last_resort` uses `"blob"`, which guesses `None` and so takes the `None` path. The `.bin` name that the upload path actually generates is not covered. Re-applying the generic-type check to the guessed value fixes it; a file that is genuinely something else (`contract.pdf`) is still excluded.

---

vitalii-dynamiq (commented):
Reviewed 4dbdb8b: 1 medium-severity finding.

---

dynamiq/components/audio/utils.py
tyaroshko: Confirmed — `mimetypes.guess_type("file_0.bin")` does return `application/octet-stream`, so the guess was being trusted where the stamped value was not. The generic check now applies to the guessed type as well.

Chasing it end to end showed the fix was necessary but not sufficient. With the file no longer discarded, the run got one step further and then failed at the provider:

```
400 - Unsupported file format bin
```

`prepare_audio_file` only supplies a default name when there is none, and `file_0.bin` is a name — just one that tells OpenAI nothing. So a name the agent invented for an unnamed upload now gets second-guessed: the file's first bytes are matched against the container signatures (RIFF/WAVE, ID3 and bare MP3 frames, fLaC, OggS, ftyp, EBML) and the extension is corrected to what the bytes actually are. `_looks_like_audio` uses the same sniff as its last word, so a recording whose name and content type both say "binary" is now identified positively rather than merely tolerated as unidentified. Reading the header restores the stream position, so nothing downstream sees a consumed file.

Covered by unit tests for each signature and for the `.bin` rename, and by a live test that runs the flow you described — `Agent(...).run(input_data={"files": [wav_bytes]})` — which fails on `main` with the 400 above and now transcribes and answers correctly.

Good catch on the test gap, too: `test_a_generic_content_type_with_no_usable_name_is_a_last_resort` used `"blob"`, which guesses `None` and took the wrong branch. There is now a case using the `.bin` name the upload path actually generates.

dynamiq/nodes/agents/utils.py
```diff
@@ -1053,14 +1086,21 @@ def process_tool_output_for_agent(content: Any, max_tokens: int = TOOL_MAX_TOKEN
         A processed string suitable for agent consumption.
     """
     if not isinstance(content, str):
-        if isinstance(content, dict):
+        if isinstance(content, (bytes, bytearray)):
```
vitalii-dynamiq, line 1104: ### Text and JSON bodies from `HttpApiCall` now reach the agent as `<N bytes of binary data>`

**High severity**

This branch treats every `bytes` payload as unreadable binary, but `HttpApiCall` routinely returns readable text as `bytes`. `response_type` defaults to `ResponseType.RAW`, and the auto-JSON override matches only an exact `content-type` of `application/json` (`nodes/tools/http_api_call.py:226`). Any endpoint answering with `text/html`, `application/xml`, or `application/json; charset=utf-8` therefore leaves `content` as `bytes`, which `describe_binary` replaces with a length summary.

Reproduced end to end with `requests_mock`: a GET returning `{"answer": 42, "detail": "what the agent needed"}` under `application/json; charset=utf-8` gives the agent the observation `<49 bytes of binary data>`. Before this change the payload fell through to `else: content = str(content)` and rendered as `b'{"answer": 42, ...}'` — readable. The body is now unrecoverable, so any agent workflow calling an HTTP API that does not send the exact `application/json` header stops working.

`nodes/agents/base.py:1856` applies the same assumption unconditionally and overwrites whatever this function returned, so both places need to distinguish a decodable payload from genuine binary. That override also removes the fallback: the pre-shortened summary never exceeds `dump_threshold_chars`, so the body can no longer be persisted to the sandbox and read back either.

---

dynamiq/components/audio/stt/openai.py
```diff
@@ -0,0 +1,108 @@
...
+        data = response.model_dump() if hasattr(response, "model_dump") else dict(response)
+        return normalize_openai_response(data, request)
+
+
+class OpenAISTTAdapter(OpenAICompatibleSTTAdapter):
+    """OpenAI transcription models, including ``gpt-4o-transcribe-diarize`` for speaker diarization."""
+
+    provider = "OpenAI"
+    capabilities = STTCapabilities(diarization=True, word_timestamps=True, prompt=True)
+    default_model = "gpt-4o-transcribe"
```
vitalii-dynamiq, line 88: ### OpenAI diarization always fails at run time because the default model cannot diarize

**Medium severity**

`OpenAISTTAdapter` declares `diarization=True`, but that capability belongs to `gpt-4o-transcribe-diarize`, not to the `default_model` set here. `validate_provider_config` fills `model` with `default_model` and then checks the requested options against the model-independent class `capabilities` (`nodes/audio/speech_to_text.py:143-151`), so `SpeechToText(connection=OpenAI(...), diarize=True)` passes construction. The model-specific guard lives in `build_params` and fires on the first execute:

```
ValueError: OpenAI diarization requires model 'gpt-4o-transcribe-diarize', got 'gpt-4o-transcribe'.
```

Confirmed by constructing and running that node. The zero-config diarization path for the provider this PR leads with is therefore a guaranteed run-time failure, and it contradicts the stated contract that an unsupported combination fails at workflow construction.

`word_timestamps=True` has the same shape: it is declared unconditionally, while `build_params` rejects it for every `*-transcribe` model including the default, so `SpeechToText(connection=OpenAI(...), timestamps="word")` also builds and then fails. Resolving the model before `check_config` — or selecting the diarization model when `diarize` is set — moves both back to build time.

---

vitalii-dynamiq (commented):
Reviewed 2ba2968: 1 high, 1 medium.

---

dynamiq/nodes/agents/utils.py
tyaroshko: Confirmed, and this was a regression I introduced — thanks for catching it. Reproduced through `_run_tool` with `requests_mock`: a body sent as `application/json; charset=utf-8` reached the agent as `<49 bytes of binary data>`, and the override in `base.py` meant the sandbox fallback could not recover it either.

The mistake was equating `bytes` with "unreadable". Both places now ask the payload instead: `decode_text_payload` returns the text when the bytes decode as UTF-8 and contain no NUL, and `None` otherwise. A JSON, HTML or XML body renders as its text and keeps going through the normal truncation and sandbox-persistence path; audio, images and archives (MP3 frames fail to decode, a WAV header carries NULs) still get the summary.

`_run_tool` applies the same test before overriding, so a text body is left exactly as `process_tool_output_with_sandbox_persistence` produced it.

Covered by unit tests for both directions and by an end-to-end test on the case you described — `HttpApiCall` at its default `ResponseType.RAW` with a charset-tagged JSON header, asserted through `Agent._run_tool`.

dynamiq/components/audio/stt/openai.py
tyaroshko: Confirmed. `SpeechToText(connection=OpenAI(...), diarize=True)` built fine and then failed on first execute, which contradicts the contract the rest of the node keeps.

Fixed by making the capability check model-aware rather than by switching the model: `check_config` now takes the `model` the node just resolved, and `OpenAISTTAdapter` overrides it to apply the same two rules `build_params` enforces. Both `diarize=True` and `timestamps="word"` on a `*-transcribe` model now raise at construction.

Selecting `gpt-4o-transcribe-diarize` automatically was the other option in your note, and I went the other way deliberately: it is a different model with different pricing and output, so quietly substituting it for the one the user asked for seemed worse than telling them. The error names the model to set.

`build_params` keeps its own guard — an adapter can be driven directly, without a node — so this is a second gate rather than a moved one. `test_unsupported_input_fails_at_execution` lost the diarization case, which is now asserted at construction instead.

dynamiq/components/audio/stt/mistral.py
tyaroshko: Closing the loop on this one, since the branch moved under it.

The asymmetry described here was real at `f9f54c75`: the `SEGMENT` branch guarded on `not request.language` and the `WORD` branch did not. That guard has since been removed entirely rather than extended, because the premise behind it turned out to be wrong — Mistral's docs state that `timestamp_granularities` is incompatible with `language`, and the live API disagrees.

Probed every combination against `/v1/audio/transcriptions` with a real key:

```
language only                              -> 200
language + segment timestamps              -> 200   segments=4
language + word timestamps                 -> 200   segments=23
diarize alone                              -> 422   "timestamp granularity must be set to ['segment']"
diarize + segment timestamps               -> 200   speakers=['speaker_1','speaker_2']
diarize + language + segment timestamps    -> 200   speakers=['speaker_1','speaker_2']
diarize + word timestamps                  -> 422
```

So the only real constraint is the one the adapter already enforces: diarization carries segment granularity. `build_fields` now sends `language` alongside whatever granularity was asked for, and there is a live test pinning `diarize` + `language` working, plus a control asserting that `diarize` without segment granularity really is refused.

dynamiq/components/audio/registry.py
tyaroshko: Addressed on the branch. There is now a `GroqSTTAdapter` subclass with `default_model = "whisper-large-v3-turbo"`, and the registry maps `GroqConnection` to it rather than to `OpenAICompatibleSTTAdapter`, so `whisper-1` stays the default only for the `HttpApiKey` and `Whisper` entries — which is where it belongs.

The test gap is closed too: `test_default_models_follow_the_provider` now asserts every provider including Groq (`whisper-large-v3-turbo`), so the zero-config path is covered rather than only the explicit-model one.

Still worth flagging: Groq is the one provider with no live coverage, since there is no key to hand. It goes through the same OpenAI-compatible request path that OpenAI itself is verified on, but the model ids have not been confirmed against the live catalogue.

dynamiq/nodes/audio/speech_to_text.py
tyaroshko: Addressed, though the branch ended up somewhere slightly different from the suggestion here.

`audio` now carries `json_schema_extra={"map_from_storage": True}`, so agent file injection fills it and `is_files_allowed = True` does something.

The second half — `is_accessible_to_agent: False`, as `images/edit.py` and `images/variation.py` declare it — was deliberately not copied. Those two nodes take whatever files they are given; transcription takes exactly one, and an agent's store routinely holds a contract next to the recording. With the field hidden the model has no way to say which file it means, and the sandbox branch of `_inject_files_into_tool` becomes unreachable, because it only collects paths the LLM wrote into a `map_from_storage` field.

So the field stays visible and the model names a file, which is the same shape `http_api_call.files` uses ("Provide strings for file IDs from files"). The concern about advertising a raw `io.BytesIO` parameter is handled in the schema generator instead: a `map_from_storage` field is described by what the model should send, not by its Python annotation, and it renders as `audio (tuple[str, ...])`. Selection also no longer depends on the model getting it right — the node picks the recording out of whatever it is handed by content type, then by file name, then by sniffing the container's magic bytes.

dynamiq/components/audio/stt/deepgram.py
tyaroshko: Addressed on the branch. The condition is now `if request.diarize or request.timestamps != TimestampGranularity.NONE`, so `utterances` is requested for `WORD` as well and Deepgram keeps returning real segments alongside the word timings — no more single synthetic segment spanning the recording.

`test_deepgram_word_timestamps_keep_utterances` covers it.

maksymbuleshnyi (approved):

PR #934, opened 2026-09-13
feat: add prompt caching support bedrock

mykhailobuleshnyi:


dynamiq/nodes/llms/bedrock.py
```diff
@@ -16,9 +43,38 @@ class Bedrock(BaseLLM):
...
+        control = self.cache_control.model_dump(
+            exclude_none=True,
+            exclude={"cache_injection_point_index", "cache_tools"},
+        )
+        points = params.setdefault("cache_control_injection_points", [])
+
+        # Bedrock allows 4 breakpoints, so don't spend one when there are no
+        # tools to cache.
+        if self.cache_control.cache_tools and params.get("tools"):
+            points.append({"location": "tool_config", "control": control})
```
vitalii-dynamiq: ### `tool_config` injection point leaks into the request body on the Bedrock Invoke route

**Medium severity**

LiteLLM's `AnthropicCacheControlHook` pops `cache_control_injection_points`, applies the `message` points to the messages, and puts the non-message points (here, the `tool_config` one) *back* into `non_default_params`. For `bedrock` — not an OpenAI-compatible provider — `add_provider_specific_params_to_optional_params` then copies that unknown key straight into `optional_params`.

Only the Converse transform consumes it: `AmazonConverseConfig._transform_request` does `additional_request_params.pop("cache_control_injection_points", None)`. The Invoke route has no such pop, so the key survives into `AnthropicConfig.transform_request`, which builds the body as `data = {"model": ..., "messages": ..., **optional_params}`, and `cache_control_injection_points` is sent to Bedrock as a top-level body field, which the Anthropic-on-Bedrock invoke API rejects with a 400.

The Invoke route is reachable from this node: `BedrockModelInfo.get_bedrock_route` falls back to `"invoke"` for any model whose base name is not in `litellm.bedrock_converse_models` (a newly released Claude model that LiteLLM's list does not yet carry — the case `tests/unit/nodes/llms/test_sampling_params_unsupported.py` already exercises with `bedrock/eu.anthropic.claude-some-future-model`), and also for an explicit `bedrock/invoke/...` model string. With `cache_tools=True` (the default) and any node that passes tools, every such call fails — a request that would have succeeded without `cache_control` set.

Gating the `tool_config` point on the model actually routing to Converse avoids it.

---

dynamiq/nodes/llms/bedrock.py
```diff
@@ -8,6 +12,29 @@
 )
 
 
+class BedrockCacheControl(BaseModel):
```
vitalii-dynamiq, line 26: ### `BedrockCacheControl` is not re-exported from `dynamiq.nodes.llms`

**Low severity**

`dynamiq/nodes/llms/__init__.py:6` imports only `Bedrock`, while line 2 of that file exports the sibling class as `from .anthropic import Anthropic, AnthropicCacheControl`. A user following the established pattern — `from dynamiq.nodes.llms import Bedrock, BedrockCacheControl`, which is how the repo's own tests import the Anthropic pair (`tests/integration/nodes/llms/test_anthropic.py:8`) — gets an `ImportError`. The new test suite does not catch this because it imports from `dynamiq.nodes.llms.bedrock` directly.

---

vitalii-dynamiq (commented):
Reviewed 2952fe9: 1 medium, 1 low.

---

dynamiq/nodes/llms/bedrock.py
```diff
@@ -8,6 +13,39 @@
...
 
 
+def _routes_to_converse(model: str) -> bool:
+    """Whether LiteLLM sends this model through the Converse API.
+
+    Only the Converse transform pops ``cache_control_injection_points``. On the
+    Invoke route the leftover ``tool_config`` point is spread into the request
+    body and Bedrock 400s, so the tool breakpoint is Converse-only.
+    """
+    return BedrockModelInfo.get_bedrock_route(model) == "converse"
```
vitalii-dynamiq, line 23: ### `_routes_to_converse` misses the `converse_like` route, silently dropping the tool breakpoint

**Low severity**

`BedrockModelInfo.get_bedrock_route` returns `"converse_like"` — not `"converse"` — for any model addressed as `bedrock/converse_like/<model>` (`litellm/llms/bedrock/common_utils.py:1130`, matched as a leading path segment).

That route uses the same transform this guard is gating on. `get_bedrock_chat_config` maps it to `AmazonConverseConfig` (`common_utils.py:1327-1328`: `elif bedrock_route == "converse" or bedrock_route == "converse_like": return litellm.AmazonConverseConfig()`), and `map_openai_params` branches the same way (`litellm/utils.py:4493`). So `AmazonConverseConfig._transform_request` — the code that pops `cache_control_injection_points` and turns a `tool_config` point into a `ToolBlock(cachePoint=...)` — runs on this route too. The `converse_like` package itself is only a pointer to that transform (`llms/bedrock/chat/converse_like/transformation.py`).

The strict `== "converse"` test therefore sends `converse_like` models down the `else` branch, which logs "routes to Bedrock Invoke, which has no toolConfig section" and skips the tool point. The Invoke-route 400 this guard exists to prevent cannot happen here, so the breakpoint is dropped with no upside: with `cache_tools=True` and a tool-carrying node, the tool schemas are re-billed as uncached input on every turn instead of being cached.

Widening the test to accept both routes preserves the Invoke protection and restores tool caching for these models.

---

tyaroshko (approved):

dynamiq/nodes/llms/bedrock.py
```diff
@@ -8,6 +13,39 @@
...
+class BedrockCacheControl(BaseModel):
+    """Bedrock (Converse API) prompt caching configuration.
+
+    A breakpoint caches everything before it. Converse renders the request as
+    ``tools -> system -> messages``, so one point in the message list also
+    covers the tool schemas and system prompt.
+
+    Attributes:
+        ttl: Cache lifetime. Applies to the tool breakpoint only; LiteLLM drops
+            it on the message path, which always uses Bedrock's default.
```
vitalii-dynamiq, line 35: ### `ttl` docstring misstates which breakpoint it affects

**Low severity**

The docstring states `ttl` "Applies to the tool breakpoint only; LiteLLM drops it on the message path, which always uses Bedrock's default." Both halves are wrong, and the same `control` dict — carrying `ttl`, since `model_dump` only excludes `cache_injection_point_index` and `cache_tools` — is attached to both points at lines 80 and 92-94.

Both paths converge on one builder. The tool point calls `_build_cache_point_block(point.get("control"), model)` (`llms/bedrock/chat/converse_transformation.py:1671`); the message point is placed on the message by `AnthropicCacheControlHook` preserving the client's TTL, then read back by `get_cache_point_block`, which calls the identical `_build_cache_point_block(cache_control, model)` (`converse_transformation.py:1206`). That builder's only gate is the model (`converse_transformation.py:1222-1225`): `ttl` is set whenever it is `"5m"`/`"1h"` and `is_claude_4_5_on_bedrock(model)` — nothing distinguishes the two call sites.

So on a Claude 4.5 Bedrock model `ttl` applies to *both* breakpoints, and on every other model it applies to *neither*. A user setting `ttl="1h"` on the documented assumption that it only affects the static tool schemas also pays the 1h cache-write premium on the rolling `cache_injection_point_index=-1` message breakpoint, which is rewritten every turn — the one place the longer TTL cannot pay for itself.

---

vitalii-dynamiq (commented):
Reviewed 6485085580fef27191a90fa02ee92fce238563b0 — 2 findings (2 low).

---

PR #936, opened 2026-09-15
fix(examples): DYNAMIQ_TRACE_BASE_URL defaults to the UI host in two of eleven examples

xizhuomengcontin:
## What

Two examples default `DYNAMIQ_TRACE_BASE_URL` to the web UI host instead of the collector host.

## The split

`DYNAMIQ_TRACE_BASE_URL` appears with a literal default in eleven example files:

| default | files |
| --- | --- |
| `https://collector.sandbox.getdynamiq.ai` | 9 |
| `https://ui.sandbox.getdynamiq.ai` | 2 — `agent_feedback_analyst.py:127`, `agent_web_researcher.py:109` |

Same variable, same consumer. In all eleven the value goes straight to `DynamiqTracingCallbackHandler(base_url=...)`, which at `dynamiq/callbacks/tracing.py:603-606` hands it to `DynamiqTracingClient`.

## Which of the two is right

The library answers it. `dynamiq/clients/dynamiq.py:20`:

```python
DYNAMIQ_BASE_URL = "https://collector.getdynamiq.ai"
```

The tracing client's own default is a **collector** host. `ui.` is the web application — where a human looks at traces, not where a client posts them. So the two outliers point the tracing client at the wrong service, and anyone running those examples without setting the variable gets traces sent somewhere that is not the ingest endpoint.

This PR changes those two literals to `https://collector.sandbox.getdynamiq.ai`, matching the other nine.

## The function name, which I checked before assuming

Both outliers sit in a function called `run_workflow_with_ui_tracing`, so the `ui.` host looks deliberate at first glance. It is not load-bearing: the function builds the same `DynamiqTracingCallbackHandler` as every other example, with no UI-specific path. The name distinguishes *UI-uploaded files* handling in these two examples, not a different tracing destination — `run_workflow_with_ui_tracing` also takes `files_to_upload`, which the collector-defaulted examples do too. If the `ui.` host genuinely is a valid ingest endpoint in your sandbox, say so and I will close this; from the code I can only see one collector constant and nine examples agreeing with it.

## Scope

Two literals in two example files. No library code, no signature, no behaviour for anyone who sets `DYNAMIQ_TRACE_BASE_URL` explicitly — which is what the examples invite you to do.


tyaroshko:
Hi @xizhuomengcontin — good catch. To answer your open question: ui. is not an ingest endpoint, collector. is the only one.

Closing in favour of https://github.com/dynamiq-ai/dynamiq/pull/938 , which replaces every sandbox host with its production counterpart across 13 files and covers these same two lines — they end up at https://collector.getdynamiq.ai.

PR #937, opened 2026-09-15
fix: upd thought reasoning

olbychos:


maksymbuleshnyi (dismissed):

dynamiq/nodes/agents/components/schema_generator.py
```diff
@@ -208,7 +219,7 @@ def generate_structured_output_schemas(
                 "properties": {
                     "thought": {
                         "type": "string",
-                        "description": "Your reasoning about the next step.",
+                        "description": THOUGHT_DESCRIPTION_TOOL_CALL,
```
vitalii-dynamiq: ### Structured-output `thought` description is tool-call-specific but also governs the `finish` action

**Low severity**

`plan_next_action` is a single schema used for every structured-output turn, and its `action` field explicitly enumerates `finish` alongside the tool names (line 226). The `thought` description now reads "why this specific tool is the right one right now (not a restatement of the original task), and what you expect it to return" — instructions that do not apply on the turn where the agent selects `finish` and produces its answer, since there is no tool being chosen and no tool return to anticipate. The previous text ("Your reasoning about the next step.") was generic enough to cover both cases; the narrowed wording is only correct for the tool-selection subset.

The function-calling path does not have this problem because `provide_final_answer` carries its own `THOUGHT_DESCRIPTION_FINAL_ANSWER`. The structured-output path has no equivalent split, so the final-answer turn is described by tool-call wording.

---

vitalii-dynamiq (commented):
Reviewed ea12a4b: 1 low-severity finding.

---

maksymbuleshnyi (dismissed):

vitalii-dynamiq (commented):
Reviewed e08cb22432206ed1d73c566a0b7d4b73159129d4 — 0 findings (0 critical, 0 high, 0 medium, 0 low).

---

mykhailobuleshnyi (approved):

tyaroshko (approved):

PR #939, opened 2026-09-16
fix(operators): evaluate path operators and negated groups in Choice

vitalii-dynamiq:
## Problem

`Choice.evaluate` had no branch for the `-path` condition operators (`numeric-equals-path`, `string-greater-than-path`, `boolean-equals-path`, …), even though they are part of `ConditionOperator` and the platform UI emits them for the "Number variable", "String variable" and "Boolean variable" value types. Any Choice option comparing a variable against another variable failed the whole node at runtime:

```
ValueError: Operator ConditionOperator.NUMERIC_EQUALS_PATH not supported.
```

While in the same function: `is_not=True` on an `and`/`or` group evaluated to `False` unconditionally (`any(...) and not is_not`) instead of negating the group.

## Fix

- Map every `-path` operator to its literal counterpart, resolve the condition's `value` as a JSONPath against the same input the `variable` is resolved from, and compare with the literal operator.
- A path operator refuses a value that is not a rooted JSONPath, and one that does not select exactly one value, naming the operator and the path. Both refusals are loud on purpose: a branch that cannot read one of its operands fails rather than routing on what it could not read.
- Negate groups with the same `== (not is_not)` form the leaf operators already use.

No changes to the wire format, the enum, or the UI.

## Review findings addressed

- A value path that matched nothing was resolved by `jsonpath.filter` to `None` and compared as a value, so `numeric-equals-path` over `$.a` and `$.b` with the input `{"c": 4}` found `None == None` and took its branch, an ordering operator failed the node with a raw `TypeError` about `int` and `NoneType`, and a path matching several values (`$.items[*].n`) compared against the list. The value is now resolved with `jsonpath.filter_all`, which tells a path that matched nothing from one that matched a null, and anything other than exactly one match fails the condition naming the operator and the path. The gain is a path typo that used to misroute silently: a Choice comparing `$.left.output.total` with `$.right.output.limit` over a Python node, whose result is nested under `content`, now says `Value path '$.right.output.limit' ... selected 0 values, expected exactly one` instead of crashing on `10 > None`.
- The JSONPath guard accepted any non-empty string, so a bare word such as `USD` was read as a path to the input field of that name and a leftover literal such as `5` reached `jsonpath.filter` for its generic `Invalid expression_filter` message. The guard is now `jsonpath.is_expression`, which is rooted-and-parseable rather than merely parseable, so both are refused with the message that names the operator.
- The tests covered only the paths that exist and select one value. They now also cover a missing value path, both paths missing, an ordering operator over a missing path, a wildcard path selecting several values, a bare word and a quoted literal as the value, a single match whose value is a list (which is what a multiple match has to be told apart from), and a workflow-level run where an ordering operator over a missing path fails the node with the path in the message.

Main was merged in after [#941](https://github.com/dynamiq-ai/dynamiq/pull/941) landed. It added the `hit_policy` tests to `tests/integration/nodes/operators/test_choice.py`, where this PR had added the path-operator tests after the same decorator, so the merge kept both blocks, each under its own `@pytest.mark.parametrize`; `dynamiq/nodes/operators/operators.py`, which both changed, merged on its own, since the `hit_policy` work sits in `Choice.run` and this fix in `Choice.evaluate`.

## Verification

- New tests cover all 11 path operators (both outcomes, plus `is_not`), the enum-to-mapping completeness, the refused values and the refused match counts, path operands inside a scoped group, negated `and`/`or` groups, a workflow run using the exact option payload the UI serializes, and a workflow run that fails on a missing value path.
- `pytest tests --ignore=tests/integration_with_creds` on the merged tree: 4807 passed, 90 skipped.
- Also exercised, against this head: a YAML round trip of a workflow whose Choice holds a path operator inside a negated `or` group, and a multi-node flow where a Choice compares two upstream nodes' outputs and gates an Output node, which routes and skips as expected in both directions.
- `pre-commit` clean on the changed files (bandit, flake8, darker, pyupgrade included).


vitalii-dynamiq (commented):
Reviewed 3a3309d1756f0425bb6663d2b0551674dcb00ec2 — 0 findings (0 critical, 0 high, 0 medium, 0 low).

---

dynamiq/nodes/operators/operators.py
```diff
@@ -110,22 +126,26 @@ def evaluate(cond: ChoiceCondition, input_data: Any) -> bool:
...
-                and not cond.is_not
+            return all(Choice.evaluate(operand, value) for operand in cond.operands) == (not cond.is_not)
+
+        # Resolve the value of a path operator against the same input as the variable, then compare
+        # with the literal operator.
+        if cond.operator in PATH_OPERATORS:
+            if not isinstance(cond.value, str) or not cond.value:
+                raise ValueError(f"Operator {cond.operator} requires a JSONPath as value, got {cond.value!r}.")
+            cond = cond.model_copy(
+                update={"operator": PATH_OPERATORS[cond.operator], "value": jsonpath.filter(input_data, cond.value)}
```
acoola, line 144: **Medium: a value path that matches nothing is compared as `None`, so an equals path operator can pick its branch by mistake.**

`jsonpath.filter` returns `None` when a path has no match (`dynamiq/utils/jsonpath.py:194-204`), and this line passes that straight into the comparison below. When both paths are missing, `None == None` is true and the option wins. I ran this at head: a `Choice` with `numeric-equals-path` comparing `$.a` to `$.b`, then a default option, given the input `{"c": 4}`. The node succeeded and chose "equal" instead of "default".

The ordering operators do not misroute, they crash: `numeric-greater-than-path` on `{"a": 4}` fails the node with `TypeError: '>' not supported between instances of 'int' and 'NoneType'`, the same kind of raw error this PR removes elsewhere. A path that matches several values, like `$.items[*].n`, gives a list and behaves the same way.

Suggested fix: resolve the value with `jsonpath.filter_all`, which returns an empty list for "no match" (`jsonpath.py:207-212`), and raise a `ValueError` that names the operator and the path unless there is exactly one match. That is the same rule this PR already uses for a value that is not a JSONPath string:

```python
matches = jsonpath.filter_all(input_data, cond.value)
if len(matches) != 1:
    raise ValueError(
        f"Value path {cond.value!r} of {cond.operator} matched {len(matches)} values, expected one."
    )
cond = cond.model_copy(update={"operator": PATH_OPERATORS[cond.operator], "value": matches[0]})
```

If you prefer "missing path means not met", to match how a missing `variable` behaves today, then say so in the code and the docstring instead of relying on `None == None`.

tests/integration/nodes/operators/test_choice.py
```diff
@@ -780,3 +780,136 @@ def test_workflow_with_string_operator_edge_cases(
...
+        (operators.ConditionOperator.STRING_GREATER_THAN_PATH, {"a": "a", "b": "a"}, False),
+        (operators.ConditionOperator.STRING_GREATER_THAN_OR_EQUALS_PATH, {"a": "a", "b": "a"}, True),
+        (operators.ConditionOperator.STRING_GREATER_THAN_OR_EQUALS_PATH, {"a": "a", "b": "b"}, False),
+        (operators.ConditionOperator.STRING_LESS_THAN_PATH, {"a": "a", "b": "b"}, True),
+        (operators.ConditionOperator.STRING_LESS_THAN_PATH, {"a": "a", "b": "a"}, False),
+        (operators.ConditionOperator.STRING_LESS_THAN_OR_EQUALS_PATH, {"a": "a", "b": "a"}, True),
+        (operators.ConditionOperator.STRING_LESS_THAN_OR_EQUALS_PATH, {"a": "b", "b": "a"}, False),
+    ],
+)
+def test_evaluate_path_operators(operator, input_data, expected):
```
acoola, line 812: **Suggestion: define tests for the unhappy paths too.** These cases only cover both paths existing and giving one value. It would be good to add: a missing value path, both paths missing, a wildcard path such as `$.items[*].n`, a bare word as the value, and one workflow-level case that checks what the node reports when an ordering operator hits `None`. These are the cases users hit when an upstream node changes its output shape, and they pin down whichever behaviour you choose for the missing-path case.

dynamiq/nodes/operators/operators.py
```diff
@@ -110,22 +126,26 @@ def evaluate(cond: ChoiceCondition, input_data: Any) -> bool:
...
         elif cond.operator == ConditionOperator.AND:
-            return (
-                all(Choice.evaluate(cond, value) for cond in cond.operands)
-                and not cond.is_not
+            return all(Choice.evaluate(operand, value) for operand in cond.operands) == (not cond.is_not)
+
+        # Resolve the value of a path operator against the same input as the variable, then compare
+        # with the literal operator.
+        if cond.operator in PATH_OPERATORS:
+            if not isinstance(cond.value, str) or not cond.value:
```
acoola, line 141: **Minor:** this accepts any non-empty string. A bare word like `USD` is treated as a path to the field `USD`, and a leftover literal like `"5"` gets the generic `Invalid expression_filter` error from `jsonpath.filter` instead of this message. `jsonpath.is_expression` (`dynamiq/utils/jsonpath.py:53-59`) exists for exactly this check: at head, `is_expression("USD")` is false while `is_jsonpath("USD")` is true.

dynamiq/nodes/operators/operators.py
vitalii-dynamiq: Confirmed at head, in all three shapes: with `{"c": 4}` the equals-path option succeeded and the default was skipped, `numeric-greater-than-path` over `{"a": 4}` failed the node with `'>' not supported between instances of 'int' and 'NoneType'`, and `$.items[*].n` compared the variable against the list of matches.

Fixed in 57a4f95, taking your first option: the value is resolved with `jsonpath.filter_all`, which tells a path that matched nothing from one that matched a null, and anything other than exactly one match raises with the operator and the path named, `Value path '$.b' of operator ConditionOperator.NUMERIC_EQUALS_PATH selected 0 values, expected exactly one.` Your `{"c": 4}` case now fails the node with that message instead of picking `equal`, and the ordering operators raise it too rather than the `TypeError`.

Raising rather than reading a missing path as "not met" is deliberate, and the docstring says so: an option that cannot read one of its operands is a mistake in the flow, and a routing node that quietly takes the default hides it. It is also what this PR already does for a value that is not a path. The gain shows on an ordinary typo: a Choice comparing `$.left.output.total` with `$.right.output.limit` over a Python node, whose result is nested under `content`, now names the path it could not read instead of crashing on `10 > None`. A value path that resolves to a genuine `null` still compares, since that is one match, and one selecting a field that holds a list still compares against that list.

Covered by unit cases for no match, both paths missing, an ordering operator over a missing path and a wildcard selecting two, by a workflow-level run that asserts the failure and the path in the message, and by a case pinning the single-match-that-is-a-list behaviour.

---

tests/integration/nodes/operators/test_choice.py
vitalii-dynamiq: Added in 57a4f95, all five. A missing value path and both paths missing, in `test_evaluate_path_operator_needs_exactly_one_value`, which also covers an ordering operator over a missing path and a wildcard `$.items[*].n` selecting two. A bare word and a quoted literal as the value, in `test_evaluate_path_operator_requires_jsonpath_value`, now parametrized. And a workflow-level run where `numeric-greater-than-path` over a missing `$.limit` fails the node, asserting the node status and that the message names the path, in `test_workflow_with_path_operator_over_a_missing_value_path`.

One more pins what a wildcard has to be told apart from: a path selecting a single field whose value is a list compares against that list, which is why the count comes from `filter_all` rather than from the shape of what `filter` returns.

Together they pin the behaviour chosen for the missing-path case, which is to fail the condition rather than read it as not met.

---

dynamiq/nodes/operators/operators.py
vitalii-dynamiq: Right on both counts. Fixed in 57a4f95: the guard is `jsonpath.is_expression`, so `USD` is refused with the message that names the operator instead of reading the input field of that name, and `5` no longer reaches `jsonpath.filter` for its generic `Invalid expression_filter`. A value that is not a string is still refused, since `is_expression` returns False for anything that is not a rooted string, so the explicit `isinstance` check is gone. Covered by parametrized cases for `4`, an empty string, `"5"` and `USD`.

---

vitalii-dynamiq (commented):
Reviewed 57a4f955318e2c8fa7c7f466b8b637173f571ca8: no findings (0 critical, 0 high, 0 medium, 0 low).

---

vitalii-dynamiq (commented):
Reviewed fa4f79fe724e242e2d0e6fd3d79eb9da0a9fcbbd: 0 findings (0 critical, 0 high, 0 medium, 0 low).

---

vitalii-dynamiq (commented):
Reviewed f8df8d24879b40df3358978c52ba810c8ac925b5 — no findings (0 critical, 0 high, 0 medium, 0 low).

---

tyaroshko (approved):

PR #940, opened 2026-09-16
feat(agents): enable prompt caching by default for Anthropic and Bedrock

mykhailobuleshnyi:


dynamiq/nodes/llms/base.py
```diff
@@ -997,7 +1010,12 @@ def _build_completion_params(
         if not tools and common_params.get("tool_choice") is not None:
             common_params.pop("tool_choice")
 
-        return self.update_completion_params(common_params)
+        params = self.update_completion_params(common_params)
+        # `update_completion_params` already applied the node's own config; a per-call one
+        # only fills in when the node has none, so the two can never double up.
+        if cache_control is not None and getattr(self, "cache_control", None) is None:
```
vitalii-dynamiq: ### Per-call `cache_control` breaks cross-provider fallback between Anthropic and Bedrock

**High severity**

The per-call config is applied based only on `self.cache_control is None`, with no check that the config type matches the provider. `_prepare_fallback_run` (base.py:1301) copies every run kwarg except `run_depends` to the fallback LLM, so `cache_control` travels with it.

Failure chain: an `Agent` whose `llm` is an `Anthropic` node with `fallback.llm` set to a `Bedrock` node. `_resolve_cache_control` builds an `AnthropicCacheControl` and `_run_llm` passes it on every call. The primary fails, the trigger matches, and the fallback runs with `cache_control=AnthropicCacheControl(...)`. Bedrock's node has `cache_control is None`, so this branch calls `Bedrock._apply_cache_control`, which reads `cache_control.cache_tools` (bedrock.py:82) — a field `AnthropicCacheControl` does not define — raising `AttributeError`. The mirror case (Bedrock primary, Anthropic fallback) hits `cache_control.cache_system` at anthropic.py:187.

The exception is swallowed into a FAILURE `RunnableResult`, so the fallback never produces output and the caller sees the primary's original error. Cross-provider fallback between these two nodes is silently dead for every agent, with no configuration change on the user's part. The `Any` annotation on the parameter keeps the type checker from flagging it.

---

dynamiq/nodes/agents/agent.py
```diff
@@ -188,6 +189,36 @@ def mark_tool_failure(content: str, success: Any) -> str:
...
+    Anthropic-family models (including on Bedrock) read back no cached prefix unless the
+    request carries an explicit breakpoint, unlike OpenAI/Gemini which cache automatically.
+    The field is discovered by annotation, so a node that gains one is picked up here.
+
+    Returns None when the node has no ``cache_control`` field, when the model does not
+    support caching, or when the caller set it themselves -- an explicit
+    ``cache_control=None`` is how you opt out.
+    """
+    field = type(llm).model_fields.get("cache_control")
+    if field is None or "cache_control" in llm.model_fields_set:
```
vitalii-dynamiq: ### YAML round-trip silently disables the new caching default

**Medium severity**

Using `model_fields_set` as the opt-out signal makes the default survive only in-process construction. It does not survive serialization.

`cache_control` is not in `Node.to_dict_exclude_params` (node.py:817) or `BaseLLM.to_dict_exclude_params` (base.py:369), so `to_dict` emits `cache_control: None` for every LLM node. `WorkflowYAMLDumper.get_updated_node_data` only drops nulls under `skip_nullable`, which is passed `True` for connections only (yaml.py:47), so the dumped YAML carries `cache_control: null`. On load, `get_node_without_depends` builds the node with `node_cls(**node_init_data)` (loaders/yaml.py:845), passing `cache_control=None` explicitly — which puts `"cache_control"` into `model_fields_set`.

`default_cache_control` then reads that as a deliberate opt-out and returns `None`, so an agent loaded from a dumped workflow makes uncached calls while the identical agent built in code caches. Nothing logs the difference, and no code change by the user is involved: dump and reload is enough.

---

vitalii-dynamiq (commented):
Reviewed f07fccd: 1 high, 1 medium.

---

dynamiq/nodes/llms/base.py
vitalii-dynamiq: ### Per-call `cache_control` reaches a fallback LLM without re-checking `supports_prompt_caching`

**High severity**

The `supports_prompt_caching` gate runs once, in `default_cache_control` (agent.py:213), against the *primary* node's model. This line applies the resulting config on whatever node ends up receiving the call, checking only that the node has no config of its own.

`_prepare_fallback_run` (base.py:1301) copies every run kwarg except `run_depends` into the fallback run, so `cache_control` travels to `self.fallback.llm`. A fallback node whose own `cache_control` is `None` therefore takes this branch and builds `cache_control_injection_points` for a model the gate never saw.

Failure chain: `Agent(llm=Bedrock(model="us.anthropic.claude-sonnet-4-5-...", fallback=Fallback(llm=Bedrock(model="us.meta.llama3-3-70b-instruct-v1:0"))))`. `_resolve_cache_control` passes the gate on the Claude model and stores a `BedrockCacheControl`; `_run_llm` passes it on every call (agent.py:1800). The primary fails, the fallback runs with that `cache_control`, `_apply_cache_control` attaches a cachePoint, and Bedrock rejects the request for a model that does not support it — the exact behavior the comment at agent.py:209 documents as the reason the gate exists. The fallback meant to rescue the failed primary fails too, so the agent step produces nothing.

This is not the config-shape mismatch fixed earlier in this PR: the two config classes are now interchangeable, but a structurally valid config still gets attached to a model that cannot accept one. The gate belongs in the node that sends the request.

---

vitalii-dynamiq (commented):
Reviewed fcf5da6: 1 finding (1 high).

---

maksymbuleshnyi (dismissed):

vitalii-dynamiq (commented):
Reviewed 4a4d659b4a38cae340c4dfbcc58b6cc4f4e2f786 — 0 findings.

---

maksymbuleshnyi (approved):

PR #942, opened 2026-09-17
fix(tools): create new todos in merge mode instead of failing

tyaroshko:
## What changed

`todo-write`'s merge mode now creates the list when the store is empty, instead of rejecting every item:

- **`merge=true`, empty store:** the items are stored as the new list. This is what a model's first call of a run hits, since `merge` defaults to true.
- **`merge=true`, existing list:** only the status of existing ids is updated. An unknown id fails the whole call with the recoverable `Todo ids not found: [...]. Existing ids: [...]`, and nothing is written.
- **`merge=false`:** always replaces the list.

Two guards sit around the create path, so a failed read can't be mistaken for an empty store:

- `_load_todos` now separates "no list yet" (`[]`) from "couldn't read or parse it" (`None`). A failing `exists()` or `retrieve()`, truncated JSON, or a `todos` key that isn't a list all return `None`, and merge mode fails without writing. Otherwise a status update would replace the real plan with the placeholder `"content": "ignored"` items models send.
- The tool records whether it wrote a non-empty list in the current run (`_list_created_this_run`, reset in `clear()`, which the agent calls at the end of every run). After that, an empty read is treated as a suspect read rather than permission to recreate the list. Any save that leaves the store empty clears the flag, so it can't latch.

Every failure names `merge=false` as the way out, which always replaces the stored list, so the model has a one-call recovery instead of retrying until its step limit.

Todo status is also now serialised with `model_dump(mode="json")`, so the listing the model reads shows `in_progress`, not `TodoStatus.IN_PROGRESS`.

Unchanged: replace mode, the CREATE/UPDATE contract in the tool description (wording only), and the sandbox backends' `exists()`, which returns "missing" on any error. That last one is why the run-scoped flag exists rather than trusting `exists()`.

## Why

`merge` defaults to true, so models routinely create their first plan with it, either explicitly or by omitting the flag. The tool then rejected every item with `Existing ids: []`, costing a loop until the model retried with `merge=false`. Found while running catalyst's issue-discovery agent end to end; traces showed it only ever happened on the first call of a run.

## Notes for review

- Merge mode is stricter than before: it refuses whenever the store's state is ambiguous, rather than guessing. That costs a round trip when a sandbox read is merely flaky, and avoids overwriting a real plan when it isn't.
- The run flag lives on the tool instance. A sub-agent borrowing a shared sandbox gets its own `TodoWriteTool`, so nothing needs to be shared between them.

## Tests

`tests/unit/nodes/tools/test_todo_tools.py`, 15 tests: create on an empty store (flag set and omitted), status updates keeping the original content, an unknown id rejected against an existing list, state surviving across calls, replace mode, an unreadable store, a corrupt store, a missing-file result after a list was created, an empty replace save followed by a merge, both guard messages naming `merge=false`, recovery through `merge=false` after the corrupt-store and empty-store guards, a store cleared externally mid-run, and statuses rendering as plain values.


dynamiq/nodes/tools/todo_tools.py
```diff
@@ -169,22 +170,21 @@ def execute(
...
-                )
-
+            # Upsert: merge=true is the default, so models routinely create the first list
+            # with it. Every item carries id, content and status, which is enough to add it.
             for todo in new_todos:
-                existing_by_id[todo["id"]]["status"] = todo["status"]
+                if todo["id"] in existing_by_id:
+                    existing_by_id[todo["id"]]["status"] = todo["status"]
+                else:
+                    existing_by_id[todo["id"]] = todo
```
vitalii-dynamiq, line 269: ### Unconditional upsert injects placeholder-content todos on an id mismatch

**Medium severity**

The upsert is applied to every merge call, not just the empty-store case the PR describes ("Traces show this only on the first call of a run"). That interacts badly with the update convention documented a few lines above: the `UPDATE (merge=true)` example tells the model content is ignored and to send `{"id": "1", "content": "ignored", "status": "completed"}`.

So when a status update carries an id the store does not have — a renumbered plan, a hallucinated or mistyped id, exactly the case the removed guard existed for — the `else` branch inserts that item verbatim. The stored plan gains a todo whose content is the literal placeholder ("ignored"), `final_todos` grows, and the listing and `📊 Stats` returned to the model report that phantom item as part of the plan. Previously the model got a recoverable `Todo ids not found: [...]. Existing ids: [...]` naming the valid ids, which it could act on; now the corruption is silent, and merge-mode calls can only ever add, so the orphan survives every subsequent update until something issues a full `merge=false` replacement.

Gating the upsert on an empty store (or on the id being absent from a store that is itself empty) fixes the reported first-call failure while keeping the guard for the mid-run case that the PR's own traces say was never broken.

---

vitalii-dynamiq (commented):
Reviewed 3a18f64c25824071157529bf1257ebfdde1a7e22 — 1 finding (0 critical, 0 high, 1 medium, 0 low).

---

dynamiq/nodes/tools/todo_tools.py
tyaroshko: Fixed in 72e3e373. Merge mode now creates the list only when the store is empty, which is the first-call case the traces showed. On a store that already has todos, the original guard is back: any id the store doesn't have fails the whole call with the recoverable `Todo ids not found: [...]. Existing ids: [...]` error, so a renumbered, mistyped or invented id can no longer add a todo whose content is "ignored". The tool description, field descriptions and `TODO_TOOLS_INSTRUCTIONS` now describe this rule. The test for the old behaviour is replaced by one that checks the error and that the store is unchanged.

dynamiq/nodes/tools/todo_tools.py
```diff
@@ -169,22 +176,34 @@ def execute(
...
+                unknown_ids = [t["id"] for t in new_todos if t["id"] not in existing_by_id]
+                if unknown_ids:
+                    raise ToolExecutionException(
+                        f"Todo ids not found: {unknown_ids}. Existing ids: {list(existing_by_id.keys())}",
+                        recoverable=True,
+                    )
+
+                for todo in new_todos:
+                    existing_by_id[todo["id"]]["status"] = todo["status"]
+            else:
```
vitalii-dynamiq, line 264: ### An unreadable todo store is treated as an empty one, so a transient read failure overwrites the whole plan

**Medium severity**

This branch keys off `existing_by_id` being empty, but `_load_todos()` returns `[]` for two very different situations: the list genuinely does not exist yet, and the list exists but could not be read. Its body is wrapped in `except Exception` (lines 137-139), so a failing `retrieve()`, a truncated/corrupt JSON payload, or a non-list `todos` key all come back as `[]`. The `exists()` check upstream fails open the same way — `E2BSandbox.exists()`, `DaytonaSandbox.exists()`, `BedrockAgentCoreSandbox.exists()` and `BedrockAgentCoreRuntimeSandbox.exists()` each catch every exception and `return False` with a "treating as missing" debug log. These are network-backed sandboxes, so a transient RPC error on the `test -f` probe during a run is a realistic event, not a hypothetical one.

When that happens on a routine status update, the call takes this `else` path. The update items are the placeholder ones the docstring above instructs the model to send (`{"id": "3", "content": "ignored", "status": "completed"}`), so `final_todos` becomes just those items, and `_save_todos` writes them with `overwrite=True` over the real file. A five-item plan is replaced by a single todo whose content is the literal string `"ignored"`. The run does not recover: `_refresh_agent_state` reloads the corrupted list into agent state, and every later `merge=true` call now finds a non-empty store and updates against the corrupted ids, so nothing surfaces the loss until the run ends.

Before this change the same situation was safe — an empty `existing_by_id` made every id unknown and raised the recoverable `Todo ids not found: [...]` error before any write, leaving the stored file intact for the model to retry against. Distinguishing "absent" from "unreadable" in `_load_todos` and taking the create path only when the store is definitively empty preserves the first-call fix while keeping that protection.

---

vitalii-dynamiq (commented):
Reviewed 72e3e373e329b4b7c6810ca7fa23d474772de463 — 1 finding (1 medium).

---

dynamiq/nodes/tools/todo_tools.py
tyaroshko: Fixed in 9c76a457. `_load_todos` now returns `None`, not `[]`, when the store's state can't be established: `exists()` or `retrieve()` raised, or the content is corrupt or not a list. In that case merge mode raises a recoverable error and writes nothing. The sandbox backends' `exists()` swallows errors, and their `retrieve()` doesn't reliably separate "not found" from other failures, so I also track whether this tool already wrote a list during the current run. That flag is reset in `clear()`, which the agent calls at the end of every run. Once a list exists, a later "empty" read is treated as a transient failure and the call fails instead of recreating the list over it. Replace mode is unchanged. Three new tests cover a failing read, corrupt content, and a missing-file check after the list was created in the run; all three fail on the previous head.

dynamiq/nodes/tools/todo_tools.py
```diff
@@ -169,29 +204,68 @@ def execute(
...
+                for todo in new_todos:
+                    existing_by_id[todo["id"]] = todo
 
             final_todos = list(existing_by_id.values())
         else:
             # Replace all
             final_todos = new_todos
 
         self._save_todos(final_todos)
+        self._list_created_this_run = True
```
vitalii-dynamiq: ### Latching `_list_created_this_run` on an empty save makes merge mode permanently unusable for the rest of the run

**Medium severity**

The flag is set after every successful save, including one where `final_todos` is empty. `run({"todos": [], "merge": False})` stores `{"todos": []}` and latches the flag, but leaves no list behind.

The next call in merge mode — the default, so this is the normal path — reloads the store. `_load_todos()` returns `[]` (the file exists and parses, it is just empty), so `existing_by_id` is empty and the `elif` at line 242 fires: the call is rejected with "The todo store looks empty, but this run already has a todo list ... Retry the call." Since the store really is empty, every retry re-reads the same empty list and fails identically. `recoverable=True` feeds that text back to the LLM, which retries as instructed until the agent hits its iteration limit. The only escape is `merge=false`, which the message never mentions.

Confirmed by running the tool directly: after `{"todos": [], "merge": False}`, three consecutive `merge=true` calls with a valid plan all return `RunnableStatus.FAILURE` with that message.

The same dead end is reached from the other direction whenever the store is genuinely emptied mid-run after a list was created — `_clear_todos_file()` (`base.py:710`) deletes `.agent/todos.json` in `execute()`'s `finally`, and that path is shared when a sub-agent borrows the same sandbox, which is exactly the case the comment at `base.py:717` describes. The parent's tool instance keeps `_list_created_this_run = True`, so its subsequent merge calls fail forever against a store that is legitimately gone.

Latching only when `final_todos` is non-empty fixes the first trigger; the guard also needs a path back to a usable state (or an error naming `merge=false`) so a genuinely empty store is not an unrecoverable loop.

---

vitalii-dynamiq (commented):
Reviewed 9c76a457: 1 medium-severity finding.

---

dynamiq/nodes/tools/todo_tools.py
tyaroshko: Fixed in 38f8f27a. The run flag is now set only when a save leaves a non-empty list, so an explicit empty `merge=false` save no longer blocks the next `merge=true` call. When the store looks empty after a list was created in this run, the guard still fails without writing. Its error now tells the model that if the list really is gone, it should recreate the whole plan with `merge=false`. `merge=false` never checks the flag, so there is always a one-call way out, including after a sub-agent sharing the sandbox clears the file. New tests cover the empty save followed by a merge, the error message, recovery through `merge=false`, and a store cleared externally mid-run.

dynamiq/nodes/tools/todo_tools.py
```diff
@@ -169,29 +206,78 @@ def execute(
...
+            if existing is None:
+                # The store exists but couldn't be read/parsed (retrieve() raised, or the
+                # content is truncated/corrupt). Treating this as "empty" would make
+                # final_todos just the placeholder items the model sends on a status update
+                # ({"id": "3", "content": "ignored", ...}), and _save_todos would overwrite the
+                # real plan with those. Fail without writing instead — the caller can retry.
                 raise ToolExecutionException(
-                    f"Todo ids not found: {unknown_ids}. Existing ids: {list(existing_by_id.keys())}",
+                    "Could not read the current todo list (read failed or its content is "
+                    "corrupt). Not overwriting it — retry the call.",
```
vitalii-dynamiq: ### Corrupt-store error tells the model to retry a call that can never succeed

**Medium severity**

`_load_todos()` returns `None` for two different conditions, and only one of them is retryable. A transient `exists()`/`retrieve()` failure clears on the next attempt, so "retry the call" is right there. Permanently bad content is not: a truncated or partially-written `todos.json` fails `json.loads`, and a payload whose `todos` key is not a list fails the `isinstance` check at line 157 — both deterministically, on every subsequent read of the same unchanged bytes.

In that state every `merge=true` call raises this same recoverable error, and the model does exactly what it is told. `recoverable=True` feeds the text back to the LLM, and `TODO_TOOLS_INSTRUCTIONS` says "Use merge=true for ALL subsequent updates", so the retry loop runs until the agent hits its iteration limit. Nothing in the message points at `merge=false`, which is the one call that bypasses `_load_todos` entirely (the `else` branch at line 267 sets `final_todos = new_todos` and overwrites the bad file), so the model has no reason to try it.

The sibling guard 30 lines below handles the identical situation correctly — it tells the model to "recreate the full plan with merge=false instead of retrying merge=true" — and `test_empty_store_guard_error_names_merge_false_as_the_escape_hatch` pins that as the standard for these messages. This branch is the one place that still leaves the model without the exit. Naming `merge=false` here closes it; `test_merge_on_corrupt_store_fails_without_overwriting` currently asserts only that the store is left alone, not that the error is actionable.

Before this change the corrupt store surfaced as `Todo ids not found: [...]. Existing ids: []`, which the PR description notes the model already recovered from by switching to `merge=false`.

---

vitalii-dynamiq (commented):
Reviewed 38f8f27a17e63858831d5ddeefa2b560ef4d680a — 1 finding (1 medium).

---

dynamiq/nodes/tools/todo_tools.py
tyaroshko: Fixed in 33eca310. When the store can't be read or its content is corrupt, the error now says both things: retry if it was a transient failure, and if it keeps failing, recreate the full plan with `merge=false`, which replaces the stored list. `merge=false` never reads the store, so it always gets past corrupt content. The unreadable-store and corrupt-store tests now check that the error names `merge=false`. A new test covers the full recovery: a corrupt store, then `merge=false` replaces it, then `merge=true` updates work again.

vitalii-dynamiq (commented):
Reviewed 33eca310 — 0 findings (0 critical, 0 high, 0 medium, 0 low).

---

maksymbuleshnyi (approved):

mykhailobuleshnyi (approved):

PR #943, opened 2026-09-17
fix(tools): return JSON/YAML from file-read as text

tyaroshko:
## What changed

`file-read` only recognised `.txt` and `.md` as text. Every other text format (`.json`, `.yaml`, `.log`, `.toml`, `.xml`, ...) hit "Unknown file extension", fell through to the raw path, and reached the agent as a Python bytes repr: `"b'{\n  \"window_start\"..."`.

A new `FileType.PLAIN_TEXT_DATA` in `dynamiq/utils/file_types.py` covers `json`, `jsonl`, `ndjson`, `yaml`, `yml`, `toml`, `xml`, `log`, `ini`, `cfg`, `conf`, `text` and `markdown`. It is deliberately left out of `DEFAULT_FILE_TYPE_TO_CONVERTER_CLASS_MAP`, so `FileReadTool.execute` decodes these files straight from their bytes on every call instead of running them through `TextFileConverter`:

- No extracted-text cache is written or read for them. A read after a write or an append returns current content, and `FileSearchTool` searches the live file instead of a stale `<path>.extracted.txt`.
- No content is stripped, so `start_line` / `end_line` and `total_lines` match the file on disk.
- Decoding uses the same encoding detection `.txt` already had, now shared helpers (`detect_encoding`, `decode_text_bytes`) in `dynamiq/components/converters/text.py`: byte-order mark first (UTF-32, UTF-16, UTF-8-sig), then strict UTF-8, then `charset_normalizer`, then UTF-8 with replacement as a last resort. A cp1252 `.log` keeps `café`, and a UTF-16 file written by PowerShell's `Out-File` reads correctly.
- Files with an extension outside this list are unchanged: still returned as bytes unless the content is valid UTF-8.

`TextFileConverter` now calls the shared detection helper, with identical behaviour.

## Why

An agent that reads a JSON manifest or a log back from its sandbox got escaped byte noise instead of text. Nothing failed, so the agent worked from the wrong content. Found while running catalyst's issue-discovery agent end to end.

## Notes for review

- `.txt` and `.md` are untouched: they still go through `TextFileConverter` and its extraction cache, with the same stale-read and content-stripping behaviour they have on `main`. Fixing those is a separate change.
- When a file is not valid UTF-8, `charset_normalizer` scans it, the same as `.txt` does today. Only noticeable on large non-UTF-8 files.

## Tests

11 new tests in `tests/unit/nodes/tools/test_file_tools.py` (65 in the file): each new extension read as text, a `.log` with undecodable bytes, valid UTF-8 under an unknown extension, binary content still returned as bytes, a fresh re-read after a write and after an append with no cache file, line ranges on a file with leading blank lines, file search after a write, a cp1252 `.log`, a UTF-16-with-BOM `.json`, and a UTF-8-with-BOM `.yaml`.


dynamiq/utils/file_types.py
```diff
@@ -77,6 +77,21 @@ class FileType(str, enum.Enum):
...
     FileType.EBOOK: {"epub"},
     FileType.HTML: {"html"},
-    FileType.TEXT: {"txt"},
-    FileType.MARKDOWN: {"md"},
+    # Plain-text data formats: read as text rather than handed back as raw bytes.
+    FileType.TEXT: {
+        "txt",
+        "text",
+        "log",
+        "json",
```
vitalii-dynamiq, line 90: ### Mapping these extensions to TEXT makes re-reads return stale content after a write

**High severity**

Routing `.json`/`.yaml`/`.log`/`.toml`/... through the converter path also enrolls them in the extracted-text cache. `FileReadTool.execute` persists `<path>.extracted.txt` via `_persist_extracted_text` (file_tools.py:830) whenever `allow_cache` is set — which it is by default, since `allow_cache = input_data.instructions is None and input_data.document_mode == "file"`. On the next read, `_load_cached_text` (file_tools.py:1141) returns that cache unconditionally: it only checks `exists()`, with no mtime or hash comparison, and `FileWriteTool._execute_write` never removes it. Before this diff these extensions fell through to the raw path and were never cached, so every read was fresh.

Reproduced on this head with `InMemoryFileStore`:

```
fs.store("state.json", b'{"v": 1}')
read1 -> '{"v": 1}'            # state.json.extracted.txt now exists
write  -> disk holds b'{"v": 2}'
read2 -> '{"v": 1}'            # stale
```

Append is affected the same way, which hits log tailing hard — a `.log` written `start\n` then appended `more\n` re-reads as `start` forever. The same stale copy is what `FileSearchTool._load_search_text` prefers (file_tools.py:1801), so searches return hits against content that is no longer on disk. Swapping the extension to an unmapped one (`.jsonx`) on the same head yields no cache file and a correct `{"v": 2}` on the second read, confirming the regression comes from this mapping.


---

dynamiq/utils/file_types.py
```diff
@@ -77,6 +77,21 @@ class FileType(str, enum.Enum):
     FileType.DATABASE: {"sqlite"},
     FileType.EBOOK: {"epub"},
     FileType.HTML: {"html"},
-    FileType.TEXT: {"txt"},
-    FileType.MARKDOWN: {"md"},
+    # Plain-text data formats: read as text rather than handed back as raw bytes.
+    FileType.TEXT: {
+        "txt",
+        "text",
+        "log",
```
vitalii-dynamiq, line 89: ### TextFileConverter strips content, so start_line/end_line return the wrong lines

**Medium severity**

`TextFileConverter._create_documents` does `content = content.strip()` (dynamiq/components/converters/text.py:82). Because these extensions now reach that converter, `_slice_lines` runs over stripped text while the caller's line numbers refer to the file on disk. Leading blank lines are silently deleted, shifting every line number and under-reporting `total_lines`. The raw path this diff replaces preserved the bytes exactly.

Reproduced on this head, file content `"\n\nline3\nline4\nline5\n"`, requesting lines 3-3:

```
run.log   -> "--- Lines 3-3 of 3 ---\nline5"    total_lines=3
run.txtx  -> "--- Lines 3-3 of 5 ---\nline3"    total_lines=5   (unmapped extension, correct)
```

An agent paging through a log or a pretty-printed JSON file that starts with a blank line gets different content than it asked for, with a header that asserts the wrong range and no indication anything shifted. The added tests don't catch this: the `.json`/`.yaml` cases use a payload with no surrounding whitespace, and the trailing-newline assertion lives in the unknown-extension test, which never touches the converter.


---

vitalii-dynamiq (commented):
Reviewed 0d6e7ca — 2 findings: 1 high, 1 medium.

---

dynamiq/utils/file_types.py
tyaroshko: Fixed in c03e1582. These formats no longer go through the converter or the extracted-text cache. A new `FileType.PLAIN_TEXT_DATA` covers them and has no converter, so `FileReadTool` decodes their bytes from the file store on every read and never writes or reads `<path>.extracted.txt`. `FileSearchTool` only uses that cache when the file exists, so it reads fresh content too. New tests for write then re-read and append then re-read on `.json` and `.log`, and for search after a write, fail on the PR head and pass now. `.txt` and `.md` have the same stale-cache behaviour, but that predates this PR and isn't changed here.

dynamiq/utils/file_types.py
tyaroshko: Fixed in c03e1582, with the same change as the other thread. These extensions skip `TextFileConverter` and its `.strip()`, and are decoded straight from the bytes, so `_slice_lines` runs over exactly what is on disk. Your repro is now a test on `.json` and `.log`: leading blank lines, lines 3-3, expecting `total_lines == 5` and the right line. It fails on the PR head (`total_lines == 3`) and passes now. `.txt` and `.md` still strip their content, which also predates this PR.

dynamiq/utils/file_types.py
```diff
@@ -77,6 +78,24 @@ class FileType(str, enum.Enum):
     FileType.DATABASE: {"sqlite"},
     FileType.EBOOK: {"epub"},
     FileType.HTML: {"html"},
-    FileType.TEXT: {"txt"},
-    FileType.MARKDOWN: {"md"},
+    FileType.TEXT: {"txt", "text"},
+    FileType.MARKDOWN: {"md", "markdown"},
```
vitalii-dynamiq: ### `.text` and `.markdown` are newly routed into the converter cache, so they get stale re-reads and shifted line numbers

**Medium severity**

`"text"` and `"markdown"` are added by this PR, but unlike the other formats they land on `FileType.TEXT`/`FileType.MARKDOWN` rather than `FileType.PLAIN_TEXT_DATA`. Both of those types are in `DEFAULT_FILE_TYPE_TO_CONVERTER_CLASS_MAP` (file_tools.py:157-158), so files with these two extensions now go through `TextFileConverter` and `_persist_extracted_text`. On `main` they were unmapped and took the raw path, so this PR moves them into exactly the two failure modes the `PLAIN_TEXT_DATA` comment above describes.

Reproduced on this head with `InMemoryFileStore` (`.md` behaves the same but predates this PR; `.json` is correct here):

```
notes/a.markdown  read1 '{"v": 1}' -> cache file written -> write '{"v": 2}' -> read2 '{"v": 1}'   stale
notes/b.text      read1 '{"v": 1}' -> cache file written -> write '{"v": 2}' -> read2 '{"v": 1}'   stale
notes/c.json      read1 '{"v": 1}' -> no cache          -> write '{"v": 2}' -> read2 '{"v": 2}'   fresh
```

and for `_slice_lines` over content `"\n\nline3\nline4\nline5\n"` requesting lines 3-3:

```
notes/a.markdown  -> "--- Lines 3-3 of 3 ---\nline5"   total_lines=3   (converter .strip() ate the blanks)
notes/b.text      -> "--- Lines 3-3 of 3 ---\nline5"   total_lines=3
notes/c.json      -> "--- Lines 3-3 of 5 ---\nline3"   total_lines=5   (correct)
```

`FileSearchTool._load_search_text` (file_tools.py:1816) prefers the same `.extracted.txt` copy, so searches over a `.markdown` file also match content that is no longer on disk. `test_file_read_tool_returns_text_formats_as_text` covers `notes/readme.markdown` but passes only because its fixture has no surrounding whitespace and is read once.

Moving `"text"` and `"markdown"` into `FileType.PLAIN_TEXT_DATA` fixes both.

---

vitalii-dynamiq (commented):
Reviewed c03e15824d8acaf168da7a13265e2f3ea39972a8: 1 medium-severity finding.

---

dynamiq/utils/file_types.py
tyaroshko: Fixed in 08331a7c. `text` and `markdown`, the two extensions this PR added, now map to `FileType.PLAIN_TEXT_DATA`, so they are decoded raw like `.json` and `.log` and never touch `TextFileConverter` or the extracted-text cache. `TEXT` and `MARKDOWN` are back to `txt` and `md`, exactly as on `main`, so every extension is either unchanged from main or read raw. Your `.markdown` and `.text` cases are added to the fresh re-read and line-range tests; they fail on the previous head and pass now.

dynamiq/nodes/tools/file_tools.py
```diff
@@ -892,12 +902,19 @@ def execute(
...
-                except UnicodeDecodeError:
+            # Bytes reach the agent as their Python repr (b'...' with escaped newlines),
+            # so anything that is valid UTF-8 is handed back as text. A recognized plain-text
+            # data format (e.g. .log, .json) is decoded leniently instead of falling back to
+            # bytes on a stray invalid byte, since the caller already knows it is text.
+            try:
+                text_fallback = content.decode("utf-8")
+            except UnicodeDecodeError:
+                if detected_type in RAW_TEXT_FILE_TYPES:
+                    text_fallback = content.decode("utf-8", errors="replace")
```
vitalii-dynamiq: ### Non-UTF-8 plain-text data is silently corrupted, unlike the identical bytes in a `.txt`

**Medium severity**

The raw path decodes `PLAIN_TEXT_DATA` with a hardcoded `utf-8` plus `errors="replace"`, with no charset detection. `TextFileConverter`, which every other text type still goes through, calls `_detect_encoding` (`dynamiq/components/converters/text.py:89`) and runs `charset_normalizer.from_bytes` before decoding. So the same bytes now decode differently depending only on the extension.

`detect_file_type` (file_tools.py:112) is purely extension-based, so a cp1252-encoded log — ordinary for anything written on Windows — takes this branch:

```
b'start\ncaf\xe9 au lait\nend\n'

run.log  -> 'start\ncaf� au lait\nend\n'   (this line: strict utf-8 raises on 0xe9, then errors="replace")
run.txt  -> 'start\ncaf\xe9 au lait\nend\n'     (charset_normalizer detects the codepage, decodes intact)
```

The agent gets `caf<?>` with nothing in the payload signalling that characters were dropped, and the original byte is unrecoverable from the returned string. Before this PR these extensions fell through to the binary rendering, where the bytes were escaped but preserved.

UTF-16 is worse: a manifest or log written by PowerShell's default `Out-File` starts with the BOM `\xff\xfe`, which fails strict UTF-8, so the whole file comes back as a replacement character followed by NUL-interleaved text.

Routing this decode through `TextFileConverter._detect_encoding` — which needs no converter pass and no extraction cache — gives these formats the same encoding handling `.txt` already has.

---

vitalii-dynamiq (commented):
Reviewed 08331a7c6daa51a374e10726a79f5461e7722d29: 1 medium.

---

dynamiq/nodes/tools/file_tools.py
tyaroshko: Fixed in c388a81c. Plain-text data formats now go through `decode_text_bytes`, a new helper in `components/converters/text.py`. It checks for a BOM (UTF-32, UTF-16, UTF-8-sig) first, then tries strict UTF-8, then the same `charset_normalizer` detection `.txt` uses, and only falls back to UTF-8 with replacement if all of those fail. The detection logic is shared, so `TextFileConverter` delegates to it and `.txt` behaves as before. Unknown extensions still return bytes unless the content is valid UTF-8. New tests cover a cp1252 `.log`, a UTF-16-with-BOM `.json` and a UTF-8-with-BOM `.yaml`; all three fail on the previous head.

vitalii-dynamiq (commented):
Reviewed c388a81c242195172cfb9b91cd2a175e1573dd81 — 0 findings (0 critical, 0 high, 0 medium, 0 low).

---

dynamiq/nodes/tools/file_tools.py
```diff
@@ -875,12 +886,24 @@ def execute(
                     f"Tool {self.name} - {self.id}: file processing failed: {str(e)}, falling back to raw content"
                 )
 
-            if input_data.start_line is not None or input_data.end_line is not None:
+            # Bytes reach the agent as their Python repr (b'...' with escaped newlines),
```
mykhailobuleshnyi: Let's leave more concise comments. Rest look good.

mykhailobuleshnyi (dismissed):

maksymbuleshnyi (dismissed):

vitalii-dynamiq (commented):
Reviewed 17364f3213f31410bb1c1ab64edf4c86e90b7ec7: 0 findings (0 critical, 0 high, 0 medium, 0 low).

---

dynamiq/nodes/tools/file_tools.py
tyaroshko: Fixed

vitalii-dynamiq (commented):
Reviewed 44202cfb730b68951695c701e54a1ec1a88428fd: 0 findings (0 critical, 0 high, 0 medium, 0 low).

---

TrachukT (approved):

PR #950, opened 2026-09-21
feat: judgement node

mykhailobuleshnyi:
## What this adds

A `Judgement` node: it asks typed questions about a piece of state and returns calibrated answers a workflow can route on. It turns "an LLM reads this and decides" from prose an operator has to parse back into structured, thresholded values.

The node is the contract; **the judge behind it is pluggable**. Exactly one of two is configured:

- `connection` — a `TypeSafe` connection, sending every question to a System One model in a single call, with probabilities calibrated for that purpose.
- `judge` — an LLM node (answering against a JSON schema built from the questions), or an agent node, which may use its tools before answering.

All judges read the same questions and produce the same output shape, so a flow can start on an LLM and move to System One, or back, without touching anything downstream.

## Questions and output

Three question types: `noul` (yes/no, returned as a probability), `choice` (one option), and `score` (an ordered scale).

Every answer carries a probability distribution and a `confidence` computed the same way for every judge. The node output splits into:

- `decisions` — one settled value per question, ready for a `Choice` or `DecisionTable`: a boolean for yes/no (decided by `noul_threshold`), the option or level name otherwise.
- `answers` — the full detail behind each: distribution, confidence, score index, optional rationale.
- `needs_review` / `low_confidence` — which answers fell below `min_confidence`, and whether any did.
- `content` — a rendered prose version, which is what an agent sees when it calls the node as a tool.
- `evidence` — an agent judge's tool calls.

`min_confidence` is the piece worth noting: **certainty is a routable output, not just a diagnostic.** An uncertain verdict can reach a person instead of falling through to a default branch. The example workflow demonstrates exactly this — the same ticket routes to `billing` when the judge is confident and to `human-review` when the team choice splits three ways.

## Using it as an agent tool

State arrives as `state`, or — when the node declares `input_fields` — as those fields by name, which is how a workflow wires in a record without templating it. Questions are configured on the node and can be replaced or extended by name at call time, which is how an agent asks its own; `allow_agent_questions=False` keeps the rubric fixed.

## Also in this PR

- `TypeSafe` connection (`TYPESAFE_URL` / `TYPESAFE_API_KEY`).
- `ActionType.JUDGEMENT`.
- `confidence_mode`: `verbalized` (default) or `sampling`, which averages `samples` runs and requires an LLM or agent judge.
- flowcheck validation for the node, so a malformed Judgement is caught before save rather than at load.
- An example DAG (`examples/components/core/dag/judgement_workflow.yaml`) wiring a Judgement into a `DecisionTable` for support triage.

## Tests

48 tests across unit and integration, all passing. Covers each judge kind, all three question types, threshold and confidence behaviour, the agent-as-caller path, YAML round-tripping, and the flowcheck rules.

## Review feedback

The latest commit addresses two checker bugs raised in review:

- `confidence_mode: sampling` with `samples` omitted passed flowcheck but failed at load, because the check skipped an absent `samples` instead of treating it as the node's default of `1` — the value sampling rejects.
- A Judgement depending on an agent was told to move itself into that agent's `tools` array, because `JUDGEMENT_TYPE` matched the `dynamiq.nodes.tools.` prefix test. Following that advice would have broken the flow, since as an agent tool the node no longer publishes an output for downstream selectors to read.

The same warning exists in a second copy at `warn_tool_chained_after_agent` in `dynamiq/cli/commands/workflow.py` and still has the second bug; it was left alone deliberately in this PR.


dynamiq/cli/flowcheck.py
```diff
@@ -663,6 +674,75 @@ def check_expression(node, label) -> list:
...
+    samples = node.get("samples")
+    if samples is not None and not (_is_number(samples) and samples == int(samples) and 1 <= samples <= MAX_SAMPLES):
+        errors.append(f"judgement {label!r}: samples {samples!r} is not a whole number between 1 and {MAX_SAMPLES}.")
+    if mode == "sampling":
+        if has_connection:
+            errors.append(
+                f"judgement {label!r}: sampling needs an LLM or agent judge; "
+                "a System One connection returns calibrated probabilities in one call."
+            )
+        elif samples is not None and _is_number(samples) and samples < 2:
```
vitalii-dynamiq: ### `confidence_mode: sampling` with `samples` omitted passes flowcheck but fails at load

**Medium severity**

The `samples is not None` guard skips this check whenever `samples` is absent from the flow. But `Judgement.samples` defaults to `1` (`judgement.py:271`), and `validate_judge` raises `"sampling needs at least 2 samples"` when `confidence_mode == SAMPLING and self.samples < 2` (`judgement.py:306`).

So a node written as `{"type": "dynamiq.nodes.tools.Judgement", "judge": {...llm...}, "confidence_mode": "sampling"}` — sampling requested, `samples` simply not written, which is the natural way to author it — reports clean from flowcheck and then raises a pydantic validation error when the flow is loaded. That is exactly the failure this check was added to catch before save.

Treating a missing `samples` as its default fixes it:

```python
        elif (samples if samples is not None else 1) < 2 or not _is_number(samples if samples is not None else 1):
```

or more readably, resolve the effective value once above and check that.


---

dynamiq/cli/flowcheck.py
```diff
@@ -463,7 +472,7 @@ def validate(flow, known_types: set | None = None):
                     "agent's \"tools\" array instead."
                 )
 
-        if node_type.startswith("dynamiq.nodes.operators."):
+        if node_type.startswith("dynamiq.nodes.operators.") or node_type == JUDGEMENT_TYPE:
```
vitalii-dynamiq, line 468: ### A Judgement that judges an agent's output gets told to move itself into that agent's tools array

**Low severity**

This line carves `JUDGEMENT_TYPE` out of the operator branch, but the block twelve lines above (line 463) still matches it on the raw prefix:

```python
if node_type.startswith("dynamiq.nodes.tools."):
    after = [... d.get("node") in agent_ids ...]
    if after:
        warnings.append(f"node {label!r} is a standalone step that runs AFTER agent {after[0]!r} ...")
```

`JUDGEMENT_TYPE` is `"dynamiq.nodes.tools.Judgement"`, so any Judgement with `depends: [{node: <agent>}]` — an agent answers, the Judgement scores that answer — trips the warning and is told to move into the agent's `tools` array instead.

Following that advice breaks the flow: as an agent tool the Judgement no longer publishes a node output, so the downstream selectors the node exists to feed (`$.triage.output.decisions.*`, `$.triage.output.needs_review`, as in `examples/components/core/dag/judgement_workflow.yaml`) resolve against nothing.

Excluding `JUDGEMENT_TYPE` from the prefix test at line 463, the same way this line does, keeps the warning correct for real standalone tools.


---

vitalii-dynamiq (commented):
Reviewed c59924558d4fc3a00585284fa3d88a766472fc94: 2 findings (1 medium, 1 low).

---

dynamiq/nodes/tools/judgement.py
```diff
@@ -0,0 +1,900 @@
...
+        evidence = [item for run in runs for item in run["evidence"]]
+        return self._output(verdicts, model=self._judge_model(), evidence=evidence if self.backend == "agent" else None)
+
+    def _judge_verdict(self, question: JudgementQuestion, answers: list[Any]) -> _Verdict:
+        if any(not isinstance(answer, dict) for answer in answers):
+            raise ValueError("no answer")
+        names = ["yes", "no"] if question.type == QuestionType.NOUL else question.option_names
+        if self._verbalized():
+            answer = answers[0]
+            chosen = self._chosen(question, answer.get("answer"))
```
vitalii-dynamiq: ### An agent judge's answer is rejected in verbalized mode whenever `answer` is missing or reworded, even though its probabilities are usable

**Medium severity**

`_chosen` is evaluated eagerly here, before either branch decides whether it is needed, and it raises on anything it cannot map (`f"{answer!r} is not a yes or a no"`, `f"{answer!r} is not one of ..."`). In the verbalized branch it is frequently not needed:

- NOUL with `probability` present (line 744): `chosen` is used only in the `else` half of the conditional expression.
- CHOICE/SCORE (line 747): `chosen` is just the `anchor` argument, and `_distribution` reads it only when the probabilities sum to zero — it is typed `str | None` and raises `"no usable probabilities and no answer"` only in that case.

So the surrounding code is written to survive a missing answer when a distribution is present, but this line fails first.

That state is reachable on the agent backend. `_judge_call` attaches `response_format` only when `isinstance(self.judge, BaseLLM)`; an agent judge (`backend == "agent"`, exercised at `tests/unit/nodes/tools/test_judgement.py:493`) receives the schema only as prose inside the brief, and `_judge_answers` validates nothing beyond "parses as a JSON object". `confidence_mode` defaults to `VERBALIZED`, so this is the default path for an agent judge.

Chain: agent judge returns `{"urgent": {"probability": 0.9, "rationale": "..."}}` — or `{"answer": "Yes.", "probability": 0.9}`, where the trailing period defeats the exact/casefold match — → `_chosen` raises `ValueError` → caught in `_judge_output` → `ToolExecutionException: the judge's answer to 'urgent' is unreadable`. The node fails on a reply that carries a perfectly usable distribution.

Deferring the call so it runs only when the distribution cannot stand on its own fixes it; the sampling branch below genuinely needs `chosen` for every answer and is unaffected.

---

vitalii-dynamiq (commented):
Reviewed 9d749b1: 1 medium finding.

---

dynamiq/cli/flowcheck.py
```diff
@@ -30,12 +30,21 @@
 RULES_TYPE = "dynamiq.nodes.operators.Rules"
 EXPRESSION_TYPE = "dynamiq.nodes.operators.Expression"
 SUB_WORKFLOW_TYPE = "dynamiq.nodes.operators.SubWorkflow"
+# A tool by group, but a decision node by use: it feeds the operators above, and is checked with them.
```
vitalii-dynamiq: not sure if this comment is needed

dynamiq/cli/flowcheck.py
```diff
@@ -30,12 +30,21 @@
...
+JUDGEMENT_TYPE = "dynamiq.nodes.tools.Judgement"
 
 CHOICE_HIT_POLICIES = ("first", "all")
 TABLE_HIT_POLICIES = ("first", "unique", "collect")
 TABLE_AGGREGATIONS = ("list", "sum", "min", "max", "count")
 RULE_SEVERITIES = ("fail", "warn", "info")
 RULE_MISSING_POLICIES = ("not_evaluated", "fail")
+QUESTION_TYPES = ("noul", "choice", "score")
+CONFIDENCE_MODES = ("verbalized", "sampling")
+# The judge behind a Judgement node is an LLM or an agent; anything else cannot answer a question.
```
vitalii-dynamiq: not sure if this comment is needed

dynamiq/nodes/tools/judgement.py
```diff
@@ -0,0 +1,900 @@
...
+        return 1.0
+    return max(0.0, min(1.0, (count * max(probabilities) - 1) / (count - 1)))
+
+
+def _distribution(names: list[str], raw: Any, anchor: str | None) -> dict[str, float]:
+    """A probability per name, normalized to sum to 1; a missing or unusable distribution becomes the anchor."""
+    values: dict[str, float] = {}
+    for name in names:
+        try:
+            value = float((raw or {}).get(name, 0)) if isinstance(raw, dict) else 0.0
```
vitalii-dynamiq: ### A case-mismatched probability map collapses to a one-hot distribution reported with confidence 1.0

**Medium severity**

This lookup matches option names exactly, while `_chosen` (line 767) deliberately does not — its docstring is *"tolerating the case and whitespace an LLM may change"*. The two run on the same judge reply, so the answer maps and the probabilities do not.

For a choice question with `option_names == ["urgent", "normal"]` and a judge reply of `{"answer": "Urgent", "probabilities": {"Urgent": 0.55, "Normal": 0.45}}`:

- `_chosen` casefolds and returns `"urgent"`.
- Here both keys miss, every value becomes `0.0`, `total <= 0`, and the anchor branch returns `{"urgent": 1.0, "normal": 0.0}`.
- `_output` (line 803) finds `verdict.confidence is None` and computes `confidence_of([1.0, 0.0]) == 1.0` instead of the `0.1` the reply actually expresses.

A partial mismatch is worse than the documented "unusable distribution becomes the anchor" fallback: with `{"Urgent": 0.55, "normal": 0.45}` the total is `0.45`, and renormalizing yields `{"urgent": 0.0, "normal": 1.0}` — a verdict that contradicts the judge's own `answer`, again at confidence 1.0.

Either way the failure is silent and in the unsafe direction: `min_confidence` never trips, `low_confidence` stays empty, `needs_review` stays `False`, and the ambiguous record routes down the confident branch — the exact case the node exists to catch.

The state is reachable on both judge backends. `_judge_call` (line 672) attaches `response_format` only when `isinstance(self.judge, BaseLLM)`, so an agent judge receives the schema as prose in the brief with nothing validating the keys it returns; and the LLM `json_schema` is sent without `"strict": True`, so key adherence is best-effort rather than enforced.

Matching keys the way `_chosen` matches the answer — strip and casefold before the lookup — closes the gap.

---

vitalii-dynamiq (commented):
Reviewed e111810: 1 medium finding.

---

vitalii-dynamiq (commented):
Reviewed 6cc543e583d29ee1b1d6ae88fddc8cf50bea2f0c: no findings (0 critical, 0 high, 0 medium, 0 low).

---

dynamiq/nodes/tools/judgement.py
```diff
@@ -0,0 +1,916 @@
...
+    """
+
+    group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
+    name: str | None = "judgement"
+    description: str = DESCRIPTION_JUDGEMENT
+    action_type: ActionType = ActionType.JUDGEMENT
+    is_parallel_execution_allowed: bool = True
+
+    connection: TypeSafe | None = None
+    judge: Node | None = Field(default=None, description="An LLM or an agent node that answers the questions.")
```
delamainer: `TypeSafe` should be one judge implementation, and `connection`, `model`, and other TypeSafe-specific params should come off `Judgement`.

delamainer (changes requested):

dynamiq/nodes/tools/judgement.py
vitalii-dynamiq: technically LLM and Agent can also be used as judgement model

dynamiq/cli/flowcheck.py
```diff
@@ -663,6 +674,79 @@ def check_expression(node, label) -> list:
...
+    has_connection = node.get("connection") is not None
+    judge = node.get("judge")
+    if has_connection == (judge is not None):
+        errors.append(
+            f"judgement {label!r} needs exactly one judge: a TypeSafe `connection`, or a `judge` node "
+            "(an LLM or an agent)."
+        )
+    if judge is not None:
+        judge_type = str(judge.get("type") or "") if isinstance(judge, dict) else ""
+        if not judge_type.startswith(JUDGE_TYPE_PREFIXES):
```
vitalii-dynamiq: ### A `judge` with only a `type` passes flowcheck and then fails to load

**Medium severity**

`check_judgement` validates the `judge` sub-object's `type` prefix and nothing else. The judge is a real node object that the loader instantiates, and `BaseLLM` declares `model: str` with no default (`dynamiq/nodes/llms/base.py`); `OpenAI` overrides only `connection` (`connection: OpenAIConnection | None = None`), not `model`.

So `judge: {"type": "dynamiq.nodes.llms.OpenAI"}` — a judge written with a type and no `model` — reports clean from `dynamiq workflow check` and then raises a pydantic validation error for the missing field when the flow is loaded. The PR's own test locks this in: `tests/unit/cli/test_flowcheck.py:331` asserts `errors_of(flow_with(judgement(connection=None, judge={"type": "dynamiq.nodes.llms.OpenAI"}))) == []`.

The same gap covers `judge.connection`. The agent path a few hundred lines up does check both halves — `agent {label!r} has no 'llm' object.` and the `llm.connection` UUID check at lines 327-331 — because the platform rejects an agent whose `llm` is incomplete. A judgement's `judge` reaches the loader the same way with neither check.

Validating the judge's required fields the way an agent's `llm` is validated closes it.

---

vitalii-dynamiq (commented):
Reviewed 6cc543e583d29ee1b1d6ae88fddc8cf50bea2f0c — 1 medium finding.

---

dynamiq/cli/flowcheck.py
vitalii-dynamiq: Confirmed and fixed in 902a7640.

Reproduced both halves: `check_judgement` returned `[]` for `judge: {"type": "dynamiq.nodes.llms.OpenAI"}`, and building that same judge raises `ValidationError: model Field required`. The API agrees — `BaseLLM` in mono validates `Model` as `Required` and a nested `ConnectionRef` as `Required, is.UUID`, so this is a flow the platform rejects on save.

The `connection` + `model` pair the top-level LLM branch already checked is now a helper (`llm_requirements`), called from that branch unchanged and from the judge — directly for an LLM judge, and one level down at `judge.llm` for an agent judge, which also gets the missing-`llm` error the agent path has. The two tests that passed a bare judge now build a complete one, and a new test covers each case; it fails without the fix.

---

dynamiq/nodes/tools/judgement.py
```diff
@@ -0,0 +1,916 @@
...
+    sets `needs_review`, so uncertain records reach a person instead of a default branch. An LLM or agent
+    judge can also explain itself: `include_rationale` adds one per answer, and an agent's tool calls are
+    returned as `evidence`.
+    """
+
+    group: Literal[NodeGroup.TOOLS] = NodeGroup.TOOLS
+    name: str | None = "judgement"
+    description: str = DESCRIPTION_JUDGEMENT
+    action_type: ActionType = ActionType.JUDGEMENT
+    is_parallel_execution_allowed: bool = True
```
vitalii-dynamiq, line 268: ### Concurrent tool calls share one stateful agent judge

**Medium severity**

`is_parallel_execution_allowed` is unconditionally `True`, but the judge is only isolated per run when sampling is on: `_judge_for_run` (line 673) returns `self.judge` itself whenever `_sample_count() <= 1`, which is the default (`samples: int = ... default=1`). Its own comment states the reason cloning is needed — "an agent keeps per-run state".

That state is real. `Agent.execute` begins with `self.reset_run_state()` (`dynamiq/nodes/agents/base.py:750`), which calls `self.state.reset()` and clears `_streaming_tool_run_ids` and `_pending_fc_tool_call_ids` (`dynamiq/nodes/agents/agent.py:306-318`); the run then rebuilds `self._prompt.messages` (`agent.py:1235-1237`) and appends to it throughout the loop, and writes `self._last_llm_output` and `self._run_depends`.

Chain: a `Judgement` with an agent `judge` is registered as a tool on an agent with `parallel_tool_calls_enabled`; the model emits two calls to it in one batch (judging two records). `_is_tool_parallel_eligible` (`agent.py:2370`) returns `True` for both, so `_execute_tools` runs them concurrently in a `ContextAwareThreadPoolExecutor` (`agent.py:2528`) against the same `Judgement` instance. Both threads reach `_judge_for_run()`, both get the same `Agent` object, and both enter `Agent.execute`. The second call's `reset_run_state()` wipes the first call's in-flight loop state and its `_prompt.messages` is reassigned out from under the first — the two judgements interleave into one conversation and return verdicts built from the wrong record's reasoning, silently, on a node whose whole purpose is to route on that verdict.

`SubAgentTool` guards exactly this case rather than defaulting the flag on: `self.is_parallel_execution_allowed = self.is_factory_mode` (`dynamiq/nodes/tools/agent_tool.py:155`), with the description hint "Shared agent: all calls reuse the same instance — calls are executed sequentially." Either clone the judge for every run, or set the flag from whether the judge is an isolated instance.

---

vitalii-dynamiq (commented):
Reviewed 902a764: 1 finding (1 medium).

---

dynamiq/nodes/tools/judgement.py
vitalii-dynamiq: Confirmed and fixed in 69a13d08. I walked every link: the flag is unconditionally `True` where the base default is `False`; `_judge_for_run` returns `self.judge` whenever `_sample_count() <= 1`, which `samples=1` makes the default; `Agent.execute` opens with `reset_run_state()`, which resets `state` and clears the streaming and pending-call lists; and `_is_tool_parallel_eligible` returns the tool's own flag, so `_execute_tools` puts both calls in the concurrent group against one instance. `SubAgentTool:155` guards exactly this.

Took the first of your two options — `_judge_for_run` now always clones — rather than deriving the flag. Cloning makes `is_parallel_execution_allowed = True` honest for all three judges instead of quietly serialising the one case that is unsafe, it is the path sampling already exercises, and `clone()` honours `_clone_shared` so the connection is shared by reference rather than rebuilt per call. Worth noting the LLM judge was not immune either: `BaseLLM` keeps `_is_fallback_run` and overrides `reset_run_state`, so the same race existed there with a narrower window.

Covered by a test that runs the node twice and asserts each run saw a distinct judge that is not the configured one; reverting the one-line change fails it. Full unit suite green at 3297.

---

dynamiq/cli/flowcheck.py
```diff
@@ -512,6 +510,8 @@ def check_operator(node, label) -> list:
         return check_expression(node, label)
     if node_type == SUB_WORKFLOW_TYPE:
         return check_sub_workflow(node, label)
+    if node_type == JUDGEMENT_TYPE:
+        return check_judgement(node, label)
```
vitalii-dynamiq, line 518: ### A Judgement inside an agent's `tools` array is never validated by flowcheck

**Medium severity**

This is the only call site of `check_judgement`, and it is reached only from `check_operator`, which the per-node loop calls for entries of the top-level `flow["nodes"]` list (line 463: `if node_type.startswith("dynamiq.nodes.operators.") or node_type == JUDGEMENT_TYPE`). Nested nodes are not recursed into: the only place a `tools[]` entry is inspected is the loop at line 423, which dispatches `check_pipedream` for `dynamiq.nodes.tools.Pipedream` and otherwise checks nothing beyond the entry's `connection` UUID.

So a Judgement authored as an agent tool — the usage the node's `DESCRIPTION_JUDGEMENT`, `allow_agent_questions` and `is_parallel_execution_allowed` exist for — bypasses every rule this PR added. A tool written as:

```yaml
tools:
  - type: dynamiq.nodes.tools.Judgement
    name: triage
    questions:
      - { name: urgent, type: noul, instructions: "..." }
      - { name: urgent, type: choice, instructions: "..." }
```

reports clean from `dynamiq workflow check` and then raises a pydantic `ValidationError` from `validate_judge` at load, on two counts: neither `connection` nor `judge` is set ("needs exactly one judge"), and `urgent` is declared twice ("question names used more than once"). The same holds for a missing `instructions`, an out-of-range `noul_threshold`, and `confidence_mode: sampling` with a `connection` — every error `check_judgement` was written to catch before save.

Dispatching `check_judgement` from the `tools[]` loop the way `check_pipedream` is dispatched there covers both placements.

---

dynamiq/nodes/tools/judgement.py
```diff
@@ -0,0 +1,919 @@
...
+        if len(text) / CHARS_PER_TOKEN > SYSTEM_ONE_STATE_MAX_TOKENS:
+            raise ToolExecutionException(
+                f"Judgement '{self.name}': the state is about {len(text) // CHARS_PER_TOKEN:,} tokens; System One "
+                f"reads at most {SYSTEM_ONE_STATE_MAX_TOKENS:,} with a question. Judge a part of it, or summarize it "
+                "first.",
+                recoverable=True,
+            )
+        return {
+            "model": self.model,
+            "state": state,
```
vitalii-dynamiq: ### A non-JSON-serializable value in the state passes the size check and then crashes the request

**Medium severity**

The size pre-check on line 417 serializes with `default=str`, so a `datetime`, `Decimal`, `UUID` or any other object in the state measures fine. The payload built here then carries `state` as the raw object, and both senders hand it to a library encoder that has no `default` hook:

- `_send_system_one` (line 454): `requests.post(url, json=payload, ...)`
- `_send_system_one_async` (line 475): `await client.post(url, json=payload, ...)`

Both raise `TypeError: Object of type datetime is not JSON serializable` before any request is made. `TypeError` is neither `requests.RequestException` nor `httpx.HTTPError`, so it escapes the retry loop and `_transport_error`, and surfaces as a raw encoder traceback instead of the node's own recoverable `ToolExecutionException`.

The state field is typed `str | dict[str, Any] | list[Any] | None` (line 153), and `_state_of` assembles the `input_fields` variant straight from `input_data.model_extra` with no coercion, so pydantic validates a dict whose values are arbitrary Python objects. A record wired in from an upstream node — the `ticket` input field in `examples/components/core/dag/judgement_workflow.yaml` is exactly this shape — carrying a `created_at` timestamp or a `Decimal` amount reaches here unchanged.

Serializing once with `default=str` and sending that, or passing the same `default` to the transport, makes the payload match what the size check already measured.

---

vitalii-dynamiq (commented):
Reviewed 69a13d08ce16e52ae479389874a14182471545c1: 2 findings (2 medium).

---

dynamiq/cli/flowcheck.py
vitalii-dynamiq: Confirmed and fixed in b40c9046. Reproduced it: a Judgement in an agent's `tools[]` with no judge, `urgent` declared twice and `noul_threshold: 1.5` returned zero judgement errors from `validate`, then raised `needs exactly one judge` on load.

The tools loop now dispatches `check_judgement` the way it dispatches `check_pipedream`. Errors carry the placement — `judgement 'triage on node writer': ...` — since a tool name alone would not say which agent it is on. Test asserts all three errors appear for the tool placement and that a well-formed one still passes, so the check is not just rejecting every tool entry.

---

dynamiq/nodes/tools/judgement.py
vitalii-dynamiq: Confirmed and fixed in b40c9046. Reproduced the whole chain: a `datetime` in the state passes the size check, reaches the payload raw, `json.dumps` raises `TypeError: Object of type datetime is not JSON serializable`, and `isinstance(e, requests.RequestException)` is `False` — so it escapes the retry loop and `_transport_error` exactly as described.

Took the first option: the payload now carries `json.loads(text)`, the same `default=str` rendering the size check already measured, so what is measured is what is sent. Passing a `default` to the transport instead would have meant switching both senders from `json=` to `data=` plus a content-type header, and would still leave the measured and sent payloads as two different serializations.

Test sends a `datetime` and a `Decimal` through the node and asserts the wire payload is `{"opened": "2026-09-21 14:30:00", "total": "42.50"}` and that `json.dumps` of it succeeds; reverting the change fails it.

---

dynamiq/nodes/tools/judgement.py
```diff
@@ -0,0 +1,922 @@
...
+            return input_data.state
+        if not self.input_fields:
+            raise ToolExecutionException(
+                f"Judgement '{self.name}' has nothing to judge: pass `state`, or declare input fields on the node.",
+                recoverable=True,
+            )
+        provided = input_data.model_extra or {}
+        state: dict[str, Any] = {}
+        for field in self.input_fields:
+            if field.name not in provided:
```
vitalii-dynamiq: ### An unresolved `input_fields` selector judges an all-null state, and the warning written to catch it never fires

**Medium severity**

`field.name not in provided` cannot be false-negative-free on the workflow path, because an input transformer never omits a key. `mapper` (`dynamiq/utils/jsonpath.py:158-166`) writes `new_json[key] = None` when a rooted path matches nothing:

```python
found = parse(path).find(json)
if not found:
    if use_expression_as_value and not path.startswith(expression_prefixes):
        new_json[key] = path
    else:
        new_json[key] = None
```

`JudgementInputSchema` is `extra="allow"`, so that `None` lands in `model_extra` as a present key. The `in provided` test therefore passes, the warning is skipped, and `state[field.name] = provided.get(field.name)` stores `None`.

Chain, on the shape the example DAG uses (`examples/components/core/dag/judgement_workflow.yaml`: `input_fields: [{name: ticket}]`, `selector: {ticket: $.start.output.ticket}`): an upstream rename or typo makes `$.start.output.ticket` resolve to nothing → `{"ticket": None}` reaches the node → `_state_of` returns `{"ticket": None}` with no warning → the judge is asked to answer `is_urgent`, `team` and `anger` about `{"ticket": null}` and returns a distribution over the options anyway → `decisions` and `needs_review` feed the `DecisionTable` and the record routes on a verdict about nothing.

The failure is silent in both directions that matter: no log line, and no error — while the sibling case one branch up (no `state` and no `input_fields`, line 399) is a hard `ToolExecutionException`. A judge answering about a null field is not obliged to report low confidence, so `min_confidence` is not a backstop here.

Treating a `None` value as not provided — `if provided.get(field.name) is None` — restores the warning, and erroring when every declared field is unresolved matches how the no-state case is already handled.

---

dynamiq/cli/flowcheck.py
```diff
@@ -421,6 +429,9 @@ def validate(flow, known_types: set | None = None):
                 tool_errors, tool_advisory = check_pipedream(tool, where)
                 errors.extend(tool_errors)
                 warnings.extend(tool_advisory)
+            # A Judgement is built to be an agent's tool, so it reaches the loader from here too.
+            elif tool.get("type") == JUDGEMENT_TYPE:
```
vitalii-dynamiq, line 434: ### A Judgement tool is the one tool type whose `connection` is never checked

**Low severity**

This new branch joins the same `elif` chain as the two `connection` checks below it (lines 435-441), so a `tools[]` entry of type `Judgement` never reaches them. `check_judgement` does not cover the gap: it reads `connection` only for presence — `has_connection = node.get("connection") is not None` (line 671) — to decide the "exactly one judge" rule, and never validates the value.

So a Judgement authored as an agent tool with `connection: typesafe`, the name form the example YAML uses, skips both `requirement_problems` and the `UUID_RE` test, and `dynamiq workflow check` reports clean on a connection reference the platform rejects. Any other tool with the same mistake gets `connection 'typesafe' is not a UUID. Run 'dynamiq connection list'.`

Running the connection checks unconditionally for every tool entry and keeping only the type dispatch in the chain closes it, and also covers the Judgement placed directly in the DAG, where no node-level connection check applies either.

---

vitalii-dynamiq (commented):
Reviewed b40c904: 2 findings (1 medium, 1 low).

---

dynamiq/nodes/tools/judgement.py
vitalii-dynamiq: Confirmed and fixed in e12e9c56. Reproduced the mapper half directly: `mapper({'start': {'output': {}}}, {'ticket': '$.start.output.ticket'})` returns `{'ticket': None}` — key present, value null — so `field.name not in provided` could never fire and the warning was dead code.

A null field now counts as unresolved. All of them is the same nothing-to-judge `ToolExecutionException` as no state at all, naming the fields; some of them logs which resolved to nothing rather than passing them off as values, since the rest of the record is still worth judging.

The test asserting the old behaviour (`system_one.run(input_data={})` → `state == {"ticket": None}`) locked exactly this in, so it is gone; a new test covers both the all-unresolved failure and the partial case, and fails without the change.

---

dynamiq/cli/flowcheck.py
vitalii-dynamiq: The elif-chain reading is right, but this is now moot: as of b89e9da1 the Judgement node has no `connection` at all. TypeSafe moved onto `dynamiq.nodes.detectors.SystemOne`, which sits in `judge` beside the LLM and agent judges, so there is no value for the skipped checks to validate — and the judge's connection *is* checked, by `connection_requirement` in `check_judgement`.

Leaving the chain as it is. What your suggestion would still change is Pipedream, which skips the same two checks and predates this PR; folding that in would widen the diff past what this PR is for.

---

dynamiq/nodes/tools/judgement.py
vitalii-dynamiq: @delamainer check again

vitalii-dynamiq: Done as asked — `connection`, `model` and `timeout` are off `Judgement`, and `judge` is one required slot:

```python
judge: Node    # a SystemOne, an LLM or an agent
```

`dynamiq.nodes.detectors.SystemOne` is a `ConnectionNode` in the `DETECTORS` group carrying the TypeSafe connection, the model and the timeout — the same field triple `PromptInjectionDetector` already has, so it is an existing pattern rather than a new one. Each judge now carries its own settings, and switching judges is switching one node.

The exactly-one-judge rule is gone from all four places it was written: `validate_judge`, `check_judgement`, mono's `exactlyOneJudge`, and the UI's schema. `judgement.py` is 223 lines lighter; the UI's `formatJudgementMetadata`, which existed only to split one judge back into two wire fields, is mostly deleted.

It also runs standalone, returning the service's answers unread. A `Judgement` calls it directly rather than through `run()`, so a rejected API key stays the unrecoverable failure it is instead of being flattened into a failed child result.

Matching changes are green in [mono#1390](https://github.com/dynamiq-ai/mono/pull/1390), [ui#1595](https://github.com/dynamiq-ai/ui/pull/1595) and [docs#17](https://github.com/dynamiq-ai/docs/pull/17). All checks pass here too.

One open question worth your view: `DETECTORS` fits the shape exactly but surfaces under "Validators" in the docs taxonomy. If "a judgement provider is a detector" reads wrong, a `JUDGES` group is a one-line enum addition plus a category mapping.

---

dynamiq/nodes/tools/judgement.py
```diff
@@ -0,0 +1,775 @@
...
+    def _judge_verdict(self, question: JudgementQuestion, answers: list[Any]) -> _Verdict:
+        if any(not isinstance(answer, dict) for answer in answers):
+            raise ValueError("no answer")
+        names = ["yes", "no"] if question.type == QuestionType.NOUL else question.option_names
+        if self._verbalized():
+            answer = answers[0]
+            # The answer is what a distribution falls back on, so read it only when one cannot stand on its own.
+            if question.type == QuestionType.NOUL:
+                probability = (
+                    _probability(answer["probability"])
```
vitalii-dynamiq: ### A percent-scale `probability` aborts the whole judgement, while the same mistake on a choice question is renormalised

**Medium severity**

`_probability` raises on anything outside 0..1 (line 203), and here it runs whenever the key is merely *present*. The `else` half of this conditional is the documented fallback — the comment one line up says "the answer is what a distribution falls back on, so read it only when one cannot stand on its own" — but a present-and-unusable `probability` never reaches it.

The choice branch four lines down handles the identical mistake correctly: `_weights` + `_distribution` renormalise, so `{"probabilities": {"urgent": 55, "normal": 45}}` becomes `{"urgent": 0.55, "normal": 0.45}`. A yes/no question answered `{"answer": true, "probability": 90}` instead raises `ValueError: probability 90.0 is outside 0..1`, which `_judge_output` (line 594) turns into `ToolExecutionException: the judge's answer to 'urgent' is unreadable` — failing every question in the run, including the ones that parsed cleanly, on a reply whose `answer` field states the verdict unambiguously.

Nothing on the judge side prevents the percent scale. `_judge_call` (line 540) attaches `response_format` only when `isinstance(self.judge, BaseLLM)`; an agent judge receives the schema as prose inside the brief with nothing validating what it returns, and `_judge_answers` checks only that the reply parses as a JSON object. `_VERBALIZED_GUIDANCE` (line 61) describes spreading weight across answers and never states the 0..1 range. `confidence_mode` defaults to `VERBALIZED`, so this is the default path for an agent judge.

Falling back to `answer` when the probability is unusable — the behaviour this expression already has when the key is absent — makes the two question types agree.

---

dynamiq/cli/flowcheck.py
```diff
@@ -663,6 +667,92 @@ def check_expression(node, label) -> list:
...
+            errors.append(f"{where} has a name that is not an identifier, so no node could read its answer.")
+        names.append(name)
+        kind = question.get("type") or "noul"
+        if kind not in QUESTION_TYPES:
+            errors.append(f"{where}: type {kind!r} is not one of {', '.join(QUESTION_TYPES)}.")
+        if not str(question.get("instructions") or "").strip():
+            errors.append(f"{where} has no `instructions`.")
+        if kind in ("choice", "score"):
+            limit, what = (MAX_CHOICE_OPTIONS, "option") if kind == "choice" else (MAX_SCORE_LEVELS, "level")
+            options = [o for o in (question.get("options") or []) if isinstance(o, dict)]
```
vitalii-dynamiq: ### A non-dict entry in `options` is dropped before the count check, so a mixed list passes flowcheck and fails at load

**Low severity**

The `isinstance(o, dict)` filter removes malformed entries and nothing afterwards reports them — the count check on the next line sees only what survived the filter.

So a question written as:

```yaml
options:
  - {name: billing}
  - {name: sales}
  - refunds
```

leaves `options` with two dicts, satisfies `2 <= len(options)`, and `dynamiq workflow check` reports clean. At load, `JudgementQuestion.options` is `list[JudgementOption]` (line 107) with no string coercion, so the bare `refunds` raises a pydantic `ValidationError` — the before-save-not-at-load gap this checker exists to close.

The all-string form is caught, but with a count that describes neither the flow nor the fix: `options: [billing, sales, refunds]` filters to zero and reports `needs between 2 and 8 options, got 0`.

Counting the raw entries and reporting the non-dict ones separately covers both.

---

vitalii-dynamiq (commented):
Reviewed e12e9c568aaf755d82d7c012c9ead9125ea31e25 — 2 findings: 1 medium, 1 low.

---

dynamiq/nodes/tools/judgement.py
vitalii-dynamiq: Confirmed and fixed in f93000bf. Ran both halves side by side on the same node:

```
choice  {"answer": "urgent", "probabilities": {"urgent": 55, "normal": 45}}  ->  {urgent: 0.55, normal: 0.45}
noul    {"answer": true, "probability": 90}                                  ->  ValueError: probability 90.0 is outside 0..1
```

— the asymmetry exactly as described, and the ValueError fails every question in the run, not just that one.

An unusable probability now falls back on the answer, which is what the surrounding code already did when the key was absent. Added `_usable_probability`, which returns `None` for a value outside 0..1 or one that is not a number; a literal `null` falls back too, where before it raised a `TypeError` from `float(None)`.

I did not renormalise the 90 to 0.9: a lone probability has no distribution to normalise against, so 90 could as easily be noise as a percent, and the `answer` field states the verdict unambiguously. Test covers both question types in one reply and fails without the change.

---

dynamiq/cli/flowcheck.py
vitalii-dynamiq: Confirmed and fixed in f93000bf. Reproduced: `options: [{name: billing}, {name: sales}, "refunds"]` returned no errors from `check_judgement`, and `JudgementQuestion(**q)` then raised `ValidationError` on `options.2`.

The count is now over what was written rather than what survived the filter, and the non-object entries are reported separately — `has an option that is not an object (1 of 3); each one needs a \`name\`.` The all-string case you mention now reports both the real count and the malformed entries instead of a misleading `got 0`. A non-list `options` is guarded, so a stray string does not get counted by character.

Also fixed the article while I was in there: the three option messages said "a option". Test asserts the new error and that the old `got 2` count is gone; it fails without the change.

---

dynamiq/nodes/tools/judgement.py
```diff
@@ -0,0 +1,788 @@
...
+        input_data, child_kwargs = self._judge_call(state, questions, **kwargs)
+        recorder = _EvidenceRecorder()
+        check_cancellation(config)
+        result = await self._judge_for_run().run_async(
+            input_data=input_data, config=recorder.config_for(config), **child_kwargs
+        )
+        return self._judge_answers(result, recorder)
+
+    def _judge_answers(self, result: RunnableResult, recorder: "_EvidenceRecorder") -> dict:
+        if result.status != RunnableStatus.SUCCESS:
```
vitalii-dynamiq, line 589: ### A canceled judge is reported as a node failure, and the flow never learns the run was canceled

**Medium severity**

This test folds `RunnableStatus.CANCELED` in with the judge's failures and turns it into a `ToolExecutionException`. Cancellation is a distinct terminal status in this codebase, and every comparable composing node checks it first before the generic failure branch: `agents/base.py:1355`, `operators/operators.py:309` and `operators/sub_workflow.py:122` all do `if result.status == RunnableStatus.CANCELED: raise CanceledException()`.

The state is reachable on the ordinary run path. `_EvidenceRecorder.config_for` (line 775) is a `model_copy(update={"callbacks": ...})`, so the judge's config keeps the caller's `CancellationConfig` and the same `CancellationToken`. A caller that signals `config.cancellation.token.cancel()` while the judge is executing makes the judge's own `check_cancellation` raise, `Node._handle_canceled` returns a result with `status=CANCELED` (`node.py:1311`), and that result arrives here.

Chain: the judge returns CANCELED → this raises `ToolExecutionException` → `execute_with_retry`'s `except CanceledException: raise` (`node.py:1635`) does not match it, nor does `run_sync`'s `except CanceledException` (`node.py:1383`) → `_handle_failure` runs, so the Judgement result is FAILURE and `on_node_execute_error` fires where `on_node_canceled` should have. The flow then propagates cancellation only from a node result whose status is CANCELED (`flows/flow.py:583-586`), so nothing propagates; the Judgement lands in `_get_failed_nodes_with_raise_behavior()` and the run reports as failed with `the judge failed: Node '<judge>' was canceled during execution` instead of returning CANCELED.

Checking `result.status == RunnableStatus.CANCELED` and raising `CanceledException()` before the existing branch matches the three sibling nodes and restores the status.

---

vitalii-dynamiq (commented):
Reviewed f93000bf: 1 finding (0 critical, 0 high, 1 medium, 0 low).

---

dynamiq/nodes/tools/judgement.py
vitalii-dynamiq: Confirmed and fixed in 32b964e6. Checked both halves:

- All four sibling sites do exactly what you describe — `agents/base.py:1355` and `:1857`, `operators/operators.py:309`, `operators/sub_workflow.py:122` — each checking `CANCELED` before the generic failure branch.
- `_EvidenceRecorder.config_for` is `model_copy(update={"callbacks": ...})`, so only the callbacks are replaced and the caller's `CancellationConfig` and token reach the judge unchanged. The state is the ordinary run path once a caller cancels mid-judgement.

`_judge_answers` now checks `RunnableStatus.CANCELED` first and raises `CanceledException()`. Test asserts the Judgement result is CANCELED rather than FAILURE; without the change it returns `failure`.

---

vitalii-dynamiq (commented):
Reviewed 32b964e — no findings (0 critical, 0 high, 0 medium, 0 low).

---

delamainer (approved):

